PR #6, opened 2021-07-01
Adds table headers for DEX pairs table

mayurbhandary:
Adds column headers to DEX Pairs table

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before:
image

After:
image


src/containers/Token/DEXPairs/index.js
```diff
@@ -138,6 +138,13 @@ const DEXPairs = props => {
       <div className="title"> {t('top_trading_pairs')}</div>
       <div className="pairs-table">
         <table>
+          <thead>
+            <tr>
+              <th className="pair-header">Pair</th>
+              <th>Issuer</th>
+              <th className="stats-header">Stats</th>
```
mvadari, line 145: Is there a better header we can use than "Stats"? Seems a bit vague to me

mayurbhandary: I'm open to suggestions. The component shows the lowest and highest offers currently available. In the future we might add the average exchange rate as a tooltip.

mvadari: What about something like "Offer Range"?

mayurbhandary: Sure that works for me.

mayurbhandary:
@mvadari, I've included the fix to the error we were getting in this PR. I moved the `/fonts` folder into `src`.

src/containers/Token/DEXPairs/index.js
mayurbhandary: Updated in latest

mvadari (approved):

PR #12, opened 2021-07-15
linting changes

ledhed2222:

Please read commit-by-commit:

1. First commit moves our lint config directly into `package.json` and removes separate linter config files.
2. Second commit requires that we use trailing commas where possible in JS
3. Third commit autofixes based on new config.
4. Format `package.json` via `npm`, which resulted in a new package.lock

### Type of Change

- [ ] Documentation Updates
- [ ] Release


intelliot:
What's the benefit of using trailing commas where possible in JS?

I don't care either way, but if it makes no difference then I try to keep using whatever a majority of the project is currently using, to make diffs smaller and make things like `git blame` a little more efficient to use in the future.

ledhed2222:
That’s exactly the rationale behind trailing commas :) If you add to a
collection at the end of it in the future, you don’t have to also editing
the line above to add a comma. I’m very very bullish on enforcing them, for
clean history moving forward

intelliot (approved):
LGTM

mvadari (approved):
why move the settings to package.json?

ledhed2222:
@mvadari it's what newer repos do. i wanted to make changes to linting anyway so thought we should move towards this pattern while I was at it.

PR #13, opened 2021-07-15
Fix warnings

mukulljangid:
Fix Warnings

### Context of Change
Fixed component lifecycle related warnings and other props related ones.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
There were warnings related to componentWillReceiveProps and componentWillMount, which are now fixed.

## Test Plan
All existing tests passed.


src/containers/Accounts/AccountHeader/index.js
```diff
@@ -23,17 +23,26 @@ class AccountHeader extends Component {
...
     this.state = {
-      showBalanceSelector: false
+      showBalanceSelector: false,
+      prevProps: this.props
     };
     props.actions.loadAccountState(props.accountId);
   }
 
-  componentWillReceiveProps(nextProps) {
+  static getDerivedStateFromProps(nextProps, prevState) {
```
mayurbhandary, line 32: I got rid of this function and the  Balance Selector seems to work now. Looks like it happened because prevState is always not equal to nextProps which always sets showBalanceSelector to false.

mayurbhandary (approved):
Great Job!

PR #16, opened 2021-07-20
Fix warnings

mukulljangid:
Hash Duplicates and Search fixed.

### Context of Change
Token and Account Transaction tables had issue concatenating new records which lead to duplicates.
Search bar was broken because of small destructuring issue. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Introduced a function, concatTransactions, to concat new and old transactions so as to only have non-duplicate entries.


src/containers/Accounts/AccountTransactionsTable/index.js
```diff
@@ -59,11 +59,17 @@ export const AccountTxTable = props => {
...
       }));
     }
   }, [props]);
 
+  const concatTx = (a, b) => {
+    if (a.length === 0) return b;
+    if (a[0].hash === b[0].hash) return a;
+    return b;
+  };
+
```
mvadari, line 72: This method doesn't look like it's actually concatenating anything, what is it supposed to be doing?

src/containers/Header/Search.jsx
```diff
@@ -73,15 +73,14 @@ const normalize = (id, type) => {
...
   }
 
-  static getDerivedStateFromProps(nextProps, prevState) {
-    if (nextProps.redirect !== prevState.prevProps.redirect) {
-      return { redirect: '', prevProps: nextProps };
-    }
-    return null;
+  /* eslint-disable */
+  UNSAFE_componentWillReceiveProps(nextProps) {
+    /* eslint-enable */
```
mvadari, line 82: There's an eslint command that will only disable for the next line, it's something like `disable-next-line`. It would be better to use that.

src/containers/Ledger/index.js
```diff
@@ -64,7 +64,9 @@ class Ledger extends Component {
     }
   }
 
-  componentDidUpdate(nextProps) {
+  /* eslint-disable */
+  UNSAFE_componentWillReceiveProps(nextProps) {
+    /* eslint-enable */
```
mvadari, line 69: ditto

src/containers/Token/TokenTransactionsTable/index.js
```diff
@@ -61,11 +61,17 @@ export const TokenTxTable = props => {
...
       }));
     }
   }, [props]);
 
+  const concatTx = (a, b) => {
+    if (a.length === 0) return b;
+    if (a[0].hash === b[0].hash) return a;
+    return b;
+  };
+
```
mvadari, line 74: ditto above

mvadari (changes requested):

src/containers/Header/Search.jsx
mvadari: The advantage of this is that it'll only disable the line for that specific rule, instead of for everything, which is what you're doing here

src/containers/Accounts/AccountTransactionsTable/index.js
mukulljangid: It's a mistake. I changed it to a function that concatenates.

mvadari:
Please re-run `npm install` and push, you deleted a bunch of the `package-lock` file and it doesn't look like anything here should have done that

src/containers/Accounts/AccountTransactionsTable/index.js
```diff
@@ -67,7 +67,13 @@ export const AccountTxTable = props => {
   const concatTx = (a, b) => {
     if (a.length === 0) return b;
     if (a[0].hash === b[0].hash) return a;
-    return b;
+    let i = 0;
+    for (i = 0; i < b.length; i += 1) {
```
mvadari, line 71: What is `i` representing here?

src/containers/Header/Search.jsx
```diff
@@ -77,9 +77,8 @@ class Search extends Component {
     this.onKeyDown = this.onKeyDown.bind(this);
   }
 
-  /* eslint-disable */
+  // eslint-disable-next-line
```
mvadari, line 80:
```suggestion
  // eslint-disable-next-line camelcase
```
This disables the next line just for that eslint rule, instead of all of them. If you look at the fix options for this error in VS Code, it'll suggest that as the first one. (if you don't know how to do that I can show you)

src/containers/Ledger/index.js
```diff
@@ -64,9 +64,8 @@ class Ledger extends Component {
     }
   }
 
-  /* eslint-disable */
+  // eslint-disable-next-line
```
mvadari, line 67:
```suggestion
  // eslint-disable-next-line camelcase
```

src/containers/Token/TokenTransactionsTable/index.js
```diff
@@ -69,7 +69,13 @@ export const TokenTxTable = props => {
...
     if (a[0].hash === b[0].hash) return a;
-    return b;
+    let i = 0;
+    for (i = 0; i < b.length; i += 1) {
+      if (b[i].hash === a[0].hash) {
+        break;
+      }
+    }
+    return b.splice(0, i).concat(a);
   };
```
mvadari: If you're using this code in 2 places, you should write it once and put it in a utils folder

mvadari (changes requested):

src/containers/shared/utils.js
```diff
@@ -139,6 +139,18 @@ export const localizeDate = (date, lang = 'en-US', options = {}) => {
...
+export const concatTx = (a, b) => {
+  if (a.length === 0) return b;
+  if (a[0].hash === b[0].hash) return a;
+  let iterator = 0;
+  for (iterator = 0; iterator < b.length; iterator += 1) {
+    if (b[iterator].hash === a[0].hash) {
+      break;
+    }
+  }
+  return b.splice(0, iterator).concat(a);
```
mvadari, line 151: Sorry, I meant what purpose does `i`/`iterator` serve in this line? Where are you trying to end it?

src/containers/shared/utils.js
```diff
@@ -139,6 +139,18 @@ export const localizeDate = (date, lang = 'en-US', options = {}) => {
   return new Intl.DateTimeFormat(lang, options).format(date);
 };
 
+export const concatTx = (a, b) => {
+  if (a.length === 0) return b;
```
mvadari, line 143: Add some comments explaining what this function does/what it's used for, and rename `a` and `b` to be more descriptive

src/containers/shared/utils.js
```diff
@@ -139,9 +139,19 @@ export const localizeDate = (date, lang = 'en-US', options = {}) => {
...
 };
 
+/**
+ *
+ * @param {list} a The transactions in state.
+ * @param {list} b The new transactions from props.
+ * @returns {list} Concatenated list.
+ */
+// extract new items from top of array b using iterator
+// and merge it into the state array a.
```
mvadari, line 149: This should be above the params, e.g. in like 143

mukulljangid: Done!

mvadari (approved):
looks good. Nice job!

PR #17, opened 2021-07-22
Fixes precision for small number

mukulljangid:
Can go up to 20 digits after decimal point.

By default, fractional part will now have 6 digits. And for number with a lot of zeros, the fractional part can increase up to 20.

### Context of Change
The previous implementation restricted the numbers to only 4 digits in the fractional part which didn't let us see a good part of it, sometimes only zeros. If there are more than 20 zeros in the most significant fractional part, the program shall bring it down to 0 with 6 digits in the fractional part.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Before:
image
After:
image

### DIFF
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -151,8 +151,6 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') => 
   return formatted.split('1')[0].trim();
 };
 
-// displaying 4 digits will give user a sense of how small the number is.
-// less number of digits might not have siginificant information since tokens have really small value.
 export const formatLargeNumber = (d = 0, digits = 4) => {
   let variableDigits = digits;
   let numberOfZeros = 0;
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -151,8 +151,10 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') => 
   return formatted.split('1')[0].trim();
 };
 
+// displaying 4 digits will give user a sense of how small the number is.
+// less number of digits might not have siginificant information since tokens have really small value.
 export const formatLargeNumber = (d = 0, digits = 4) => {
-  let variableDigits = 4;
+  let variableDigits = digits;
   let numberOfZeros = 0;
   let numberCopy = d;
   if (d >= QUADRILLION) {
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -152,7 +152,7 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') => 
 };
 
 export const formatLargeNumber = (d = 0, digits = 6) => {
-  let digits_local = 6;
+  let digitsLocal = 6;
   let zeros = 0;
   let copy = d;
   if (d >= QUADRILLION) {
@@ -190,16 +190,16 @@ export const formatLargeNumber = (d = 0, digits = 6) => {
     };
   }
 
-  while (copy < 1 && digits_local < 20) {
+  while (copy < 1 && digitsLocal < 20) {
     copy *= 10;
     zeros += 1;
-    digits_local = zeros > digits_local - 1 ? digits_local + 1 : digits_local;
+    digitsLocal = zeros > digitsLocal - 1 ? digitsLocal + 1 : digitsLocal;
   }
 
-  digits = Math.trunc(d.toFixed(20)) === 0 ? digits : digits_local;
+  digitsLocal = Math.trunc(d.toFixed(20)) === 0 ? digits : digitsLocal;
 
   return {
-    num: d.toFixed(digits),
+    num: d.toFixed(digitsLocal),
     unit: '',
   };
 };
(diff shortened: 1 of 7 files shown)


src/containers/shared/utils.js
```diff
@@ -151,7 +151,10 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
   return formatted.split('1')[0].trim();
 };
 
-export const formatLargeNumber = (d = 0, digits = 4) => {
+export const formatLargeNumber = (d = 0, digits = 6) => {
```
mvadari, line 154: This method is used in a lot of different places across the explorer, are you sure those places aren't affected by your changes?

src/containers/shared/utils.js
```diff
@@ -151,7 +151,10 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
...
+  let digitsLocal = 6;
+  let zeros = 0;
+  let copy = d;
```
mvadari, line 157: This variable should be named more descriptively

src/containers/shared/utils.js
```diff
@@ -151,7 +151,10 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
   return formatted.split('1')[0].trim();
 };
 
-export const formatLargeNumber = (d = 0, digits = 4) => {
+export const formatLargeNumber = (d = 0, digits = 6) => {
+  let digitsLocal = 6;
```
mvadari, line 155: This variable should be named more descriptively

mvadari (changes requested):

src/containers/shared/utils.js
mukulljangid: Done!

src/containers/shared/utils.js
mukulljangid: Done!

src/containers/shared/utils.js
mukulljangid: It's only for the Tokens page and it works perfectly.

mvadari: It's also used in the token page header (the per-token page)

src/containers/shared/utils.js
```diff
@@ -151,7 +151,10 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
   return formatted.split('1')[0].trim();
 };
 
-export const formatLargeNumber = (d = 0, digits = 4) => {
+export const formatLargeNumber = (d = 0, digits = 6) => {
+  let variableDigits = 6;
+  let numberOfZeros = 0;
+  let numberCopy = d;
```
mvadari, line 157: Looks like you don't actually have to copy `d` - it's not modified anywhere else, so you can just use `d`

src/containers/shared/utils.js
mvadari, line 154: Why did the default number of digits change? You can just pass in a different value for the variable and then you have to worry less about differences

mukulljangid: Either way is fine!

src/containers/shared/utils.js
```diff
@@ -152,6 +152,9 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
 };
 
 export const formatLargeNumber = (d = 0, digits = 4) => {
+  let variableDigits = 4;
```
mvadari, line 155: Why does this start at 4?

src/containers/shared/utils.js
```diff
@@ -187,8 +190,21 @@ export const formatLargeNumber = (d = 0, digits = 4) => {
     };
   }
 
+  while (numberCopy < 1 && variableDigits < 20) {
+    numberCopy *= 10;
+    numberOfZeros += 1;
+    variableDigits = numberOfZeros > variableDigits - 1 ? variableDigits + 1 : variableDigits;
+  }
```
mvadari, line 197: Leave a comment here that explains what this is doing

mukulljangid: Counting the number of zeros after the decimal point and displaying two non-zero digits (eg. 0.00000023/45/50) at the end. Also, we can display a maximum of 20 digits.

src/containers/shared/utils.js
mukulljangid: The can be very very small. So displaying a minimum of 4 digits might give us a sense. If we try to reduce that, most small numbers won't have significant information to display.

src/containers/shared/utils.js
mvadari: No, leave a comment in the code

src/containers/shared/utils.js
mvadari: Why does it not start at `digits`?

src/containers/shared/utils.js
```diff
@@ -151,7 +151,12 @@ export const getLocalizedCurrencySymbol = (lang = 'en-US', currency = 'USD') =>
   return formatted.split('1')[0].trim();
 };
 
+// displaying 4 digits will give user a sense of how small the number is.
+// less number of digits might not have siginificant information since tokens have really small value.
```
mvadari, line 155: This comment isn't needed anymore, now that we changed `variableDigits` from a magic number.

mvadari (approved):

PR #18, opened 2021-07-28
Add Typescript Foundation

mvadari:

This PR sets up the repo to be able to handle TypeScript, and converts a few files to TypeScript as a base/example. It also changes the linting settings to handle TypeScript files, and fixes all errors that resulted from that change.

### Context of Change

making the code nicer/neater/easier to use/better

### Type of Change

## Test Plan

CI passes


mvadari:
I don't know what's going on with the `package-lock` file but it's not going away with `npm install` so I guess it's fine.

package.json
```diff
@@ -153,7 +160,8 @@
       "react-app",
       "plugin:prettier/recommended",
       "plugin:jsx-a11y/recommended",
-      "prettier/react"
+      "prettier/react",
```
natenichols, line 163: Does `npm start` interperet the typescript code, or do you have to build and start?

natenichols: I.E. does `react-scripts start` work with typescript?

natenichols: Similarly, do you have to specify anything special w/ running `npm build`.

natenichols:
Just a couple of initial questions that I don't know the answer to. I've never really done react + Typescript stuff

src/containers/shared/components/TxDetails.tsx
```diff
@@ -1,13 +1,57 @@
...
+  amount: any;
+  destination: string;
+  partial: boolean;
+}
+
+interface Props {
+  instructions: Instructions;
+  type: string | null;
+  language: string;
+  t: (s: string) => string;
```
JST5000, line 49: What does t represent here? (I see it in almost every file, but I can't figure out what it's supposed to do)

src/containers/shared/transactionUtils.ts
```diff
@@ -167,14 +203,14 @@ const zeroPad = (num, size, back = false) => {
   return s;
 };
 
-export const normalizeAmount = (amount, language = 'en-US') => {
-  const currency = amount.currency || 'XRP';
-  const value = amount.value || amount / XRP_BASE;
+export const normalizeAmount = (amount: IssuedCurrencyAmount | number, language = 'en-US') => {
+  const currency = typeof amount === 'object' ? amount.currency : 'XRP';
+  const value = typeof amount === 'object' ? amount.value : amount / XRP_BASE;
```
JST5000, line 208: It might be nice to use parentheses here to make it clear that the code will execute the right order between "(typeof amount === 'object') ? ..." or "(typeof amount === ('object' ? amount.currency : 'XRP'))"

src/containers/shared/components/TxDetails.tsx
mayurbhandary: It's from the internationalization framework that we use: https://www.i18next.com/overview/api#t

src/containers/Accounts/AccountTransactionsTable/index.js
```diff
@@ -26,53 +26,39 @@ const DATE_OPTIONS = {
 };
 
 export const AccountTxTable = props => {
-  const [state, setState] = useState({
-    transactions: [],
-    marker: undefined,
-  });
-  const { accountId, actions, data } = props;
+  const [transactions, setTransactions] = useState([]);
+  const [marker, setMarker] = useState(undefined);
```
ledhed2222, line 30: `useState(null)` please - i think explicit use of `undefined` is a smell

server/routes/v1/nodes.js
```diff
@@ -29,6 +23,12 @@ const semverCompare = (a = '', b = '') => {
   return 0;
 };
 
+const ledgerCompare = (a = 0, b = 0) => {
+  const aLedger = a.validated_ledger ? a.validated_ledger.ledger_index : 0;
+  const bLedger = b.validated_ledger ? b.validated_ledger.ledger_index : 0;
+  return bLedger === aLedger ? semverCompare(b.version, a.version) : bLedger - aLedger;
+};
+
```
ledhed2222, line 31: why did this have to move?

mvadari: because `ledgerCompare` uses `semverCompare`, and the new linter rules mean that the method must be defined before it can be used

src/containers/Tokens/TokensTable/tests/hooks.test.js
```diff
@@ -20,14 +19,6 @@ const mountTable = (isError = false, allTokens = []) => {
 };
 
 describe('Testing hooks', () => {
-  beforeEach(() => {
```
natenichols, line 23: Was this no longer being used?

mvadari: Nope, legacy stuff that I noticed while trying to fix tests for one of the other components

src/containers/shared/test/txDetails.test.js
```diff
@@ -22,7 +22,7 @@ import TxDetails from '../components/TxDetails';
 import summarize from '../../../../server/lib/txSummary';
 
 describe('TxDetails', () => {
-  const createWrapper = (tx, width = 1200) =>
+  const createWrapper = tx =>
```
natenichols, line 25: Was width no longer being used.

mvadari: The variable wasn't being used in the method, so I took it out.

src/containers/shared/transactionUtils.ts
natenichols: I like to structure my ternaries as 
```ts
const value = typeof amount === 'object'
    ? amount.value
    : amount / XRP_BASE;
```
Similar reason as using parenthesis, it improves readability.

Up to you what you wanna do here, as long as the linter is cool w/ it.

natenichols (approved):
LGTM 🎉

src/containers/shared/transactionUtils.ts
mvadari: `prettier` gets mad at me if I do either of those things, so I'm just going to leave it alone

src/containers/shared/utils.js
```diff
@@ -130,6 +122,15 @@ export const localizeNumber = (num, lang = 'en-US', options = {}) => {
   return new Intl.NumberFormat(lang, config).format(number);
 };
 
+export function formatPrice(number, lang = 'en-US', currency = 'USD', decimals = 4) {
+  return number
+    ? localizeNumber(number.toPrecision(decimals), lang, {
+        style: 'currency',
+        currency,
+      })
+    : undefined;
```
ledhed2222, line 131: FYI - we should never return `undefined` but instead return `null`. Outside of the scope of this PR I think.

tsconfig.eslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  // extend your base config so you don't have to redefine your compilerOptions
+  "extends": "./tsconfig.json",
+  "include": ["src/**/*.ts", "src/**/*.tsx", "test/**/*.ts", "./.eslintrc.js", "./webpack.config.js"]
```
ledhed2222, line 4: i'm not understanding the purpose of this file. why don't we want to include tests in the base tsconfig? and why DO we want to include estlintrc.js (which doesn't exist) or webpack.confg (which also doesn't exist)?

types.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-i18next';
```
ledhed2222, line 1: i'm not familiar enough with typescript to know about these declaration files, but I do know that the way create-react-app is supposed to handle typescript integration is to put a similar file into the `src` directory. See https://github.com/ledhed2222/gregweisbrod.com/blob/master/src/react-app-env.d.ts (from my personal site lol)

tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+    "skipLibCheck": true,
+    "allowSyntheticDefaultImports": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true
+  },
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "types.d.ts"
```
ledhed2222, line 33: i believe this can be simpler if you move the declaration file (see below comment). check out my simpler `include` section: https://github.com/ledhed2222/gregweisbrod.com/blob/master/tsconfig.json 

i'm guessing that that requires the use of `includeJs`.

.vscode/settings.json
```diff
@@ -40,4 +40,5 @@
     "css.validate": false, // Make sure to install the recommended extensions for StyleLint
     "jestRunIt.jestPath": "./backend/node_modules/.bin/jest",
     "jestRunIt.jestConfigPath": "./backend/jest.config.js",
+    "typescript.tsdk": "/Users/mvadari/Documents/explorer-open/node_modules/typescript/lib",
```
ledhed2222, line 43: this references your specific file system. this can't be right :)

package.json
```diff
@@ -3,7 +3,14 @@
...
   "version": "1.5.0",
   "dependencies": {
+    "@babel/eslint-parser": "^7.14.7",
     "@google-cloud/bigquery": "^5.5.0",
+    "@types/jest": "^26.0.24",
+    "@types/node": "^16.4.3",
+    "@types/react": "^17.0.15",
+    "@types/react-dom": "^17.0.9",
+    "@types/react-i18next": "^8.1.0",
+    "@types/react-router-dom": "^5.1.8",
```
ledhed2222, line 13: none of these are needed at runtime so they could all be "devDependencies". It doesn't really matter though, since we aren't publishing this as a package on npm; the distinction between "dependencies" and "devDependencies" is totally unnecessary in this project.

tsconfig.eslint.json
mvadari: whoops, that's a copy-paste mistake, the test folder also doesn't exist

package.json
mvadari: yeah that's why I didn't bother to move them, but I can if you want

mvadari: Just did it anyways

tsconfig.eslint.json
mvadari: Just deleted this file, linter seems to be fine with it

src/containers/shared/utils.js
```diff
@@ -157,7 +159,7 @@ export const concatTx = (a, b) => {
       break;
     }
   }
-  return b.splice(0, iterator).concat(a);
+  return a.concat(b.slice(0, iterator));
```
mukulljangid, line 162: Any reason why the order was changed?

mvadari: When using it, it's the same order as `concat`, so it makes more intuitive sense

mukulljangid (approved):

ledhed2222 (approved):

PR #22, opened 2021-08-10
fix: remove updatedTime footer & make frequent refreshes

mukulljangid:
Remove last updated time from tokens page. Make refreshes every 5 minutes to keep the data as updated as possible.

### Context of Change
Change the refresh time to 5 minutes to let users see updated information. Delete the display footer to get rid of the time inconsistencies between the 3 pods.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Before:
image

After:
image

## Test Plan
Ran `npm test` and verified that nothing is breaking.


mukulljangid:
Close this because of incorrect base.

PR #24, opened 2021-09-02
Adds logging when healthcheck fails

natenichols:
When healthcheck fails, the reason for failure will be logged.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


mvadari (approved):

intelliot (approved):
It's unfortunate that whatever is calling the healthcheck isn't able to just look at the code/message in the response

PR #34, opened 2021-10-29
fix: more GCP fixes

mvadari:

This PR:
* For dev and staging environments, it only runs the caching for 1 hour before stopping (and starting again when someone calls the endpoint again).
* Changes the running of the BQ instance to every half hour.
* Removes the `token/top` endpoint for devnet and testnet.

### Context of Change

BQ queries happening too often

### Type of Change

### TypeScript/Hooks Update

Not worth doing it here for a quick bugfix

## Test Plan

CI passes. The code works as designed locally. It also works as designed on devnet.

### DIFF
--- a/server/routes/v1/tokenDiscovery.js
+++ b/server/routes/v1/tokenDiscovery.js
@@ -11,6 +11,8 @@ const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK
   ? process.env.REACT_APP_MAINNET_LINK.includes('xrpl.org')
   : false;
 // how long the auto-caching should run in dev and staging environments
+// We want to turn it off after some time so it doesn't run when we don't need it, which costs us
+// money per BigQuery query
 const TIME_TO_TEST = 1000 * 60 * 60 * 1; // 1 hour
 
 const TIME_INTERVAL = 1000 * 60 * 30; // 30 minutes
@@ -123,10 +125,19 @@ async function cacheTokensList() {
   }
 }
 
+// starts the caching process for bigquery
 function startCaching() {
+  // only run if on mainnet (the tokens page doesn't exist on devnet/testnet)
   if (process.env.REACT_APP_ENVIRONMENT === 'mainnet') {
+    // initialize the cache
     cacheTokensList();
+    // cache every TIME_INTERVAL ms (only starts after one interval)
     const intervalId = setInterval(() => cacheTokensList(), TIME_INTERVAL);
+    // Stop the auto-running of the caching in the previous line after TIME_TO_TEST ms
+    // Only do this if not in the prod env, so we don't have excessive BigQuery queries when we're
+    // not actually using what's in the dev and staging environments
+    // We don't want regular caching to stop in prod, though, because then a missed cache would
+    // result in a several-second delay while the query is re-run, and this is a poor UX
     if (!IS_PROD_ENV) {
       setTimeout(() => {
         log.info('stopping caching tokens');
@@ -144,17 +155,26 @@ function sleep(ms) {
 
 module.exports = async (req, res) => {
   // * * * * * * * * IMPORTANT * * * * * * * * * * * *
-  // Running the BQ query costs money. Don't let it run if you don't need to.
+  // Running the BigQuery query costs money. Don't let it run if you don't need to.
   log.info(`getting token discovery`);
   try {
-    // if it's been a while since caching happened in the non-prod envs,
-    // then restart the caching
+    // if it's been a while since caching happened in the non-prod envs, then restart the caching
     // (needed because `startCaching` turns off caching after TIME_TO_TEST for non-prod envs)
     if (
+      // true if in dev or staging
       !IS_PROD_ENV &&
+      // true if the cache has already been initialized (i.e. if `startCaching` has already run at
+      // least once)
+      // (prevents race conditions where an API call is made on container startup before the
+      // initial `startCaching` has returned)
       cachedTokensList.time != null &&
+      // true if some time has passed since the cache was last updated (i.e. `startCaching` was
+      // turned off)
+      // this uses `TIME_INTERVAL * 2` to prevent race conditions around `TIME_INTERVAL`, in case
+      // the cache just took some time to load but `startCaching` has already been triggered
       Date.now() - cachedTokensList.time > TIME_INTERVAL * 2
     ) {
+      // reset the cache so the API call waits below until it's been refilled
       cachedTokensList.tokens = [];
       cachedTokensList.time = null;
       startCaching();
--- a/server/routes/v1/tokenDiscovery.js
+++ b/server/routes/v1/tokenDiscovery.js
@@ -3,6 +3,10 @@ const rippled = require('../../lib/rippled');
 
 const log = require('../../lib/logger')({ name: 'token discovery' });
 
+const IS_PROD_ENV = process.env.RELEASE_ENV.includes('prod-');
+// how long the auto-caching should run in dev and staging environments
+const TIME_TO_TEST = 1000 * 60 * 60 * 1; // 1 hour
+
 const TIME_INTERVAL = 1000 * 60 * 5; // 5 minutes
 
 const NUM_TOKENS_FETCH_ALL = 10;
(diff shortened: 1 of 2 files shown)


server/routes/v1/tokenDiscovery.js
```diff
@@ -114,21 +123,45 @@ async function cacheTokensList() {
...
+  // Running the BQ query costs money. Don't let it run if you don't need to.
   log.info(`getting token discovery`);
   try {
+    // if it's been a while since caching happened in the non-prod envs,
+    // then restart the caching
+    // (needed because `startCaching` turns off caching after TIME_TO_TEST for non-prod envs)
+    if (
+      !IS_PROD_ENV &&
+      cachedTokensList.time != null &&
+      Date.now() - cachedTokensList.time > TIME_INTERVAL * 2
```
JST5000, line 156: Why is this twice the time interval instead of 1x?

mvadari: It's a sanity check, in case there's a race condition around `TIME_INTERVAL`

server/routes/v1/tokenDiscovery.js
```diff
@@ -114,21 +123,45 @@ async function cacheTokensList() {
...
-  setInterval(() => cacheTokensList(), TIME_INTERVAL);
+function startCaching() {
+  if (process.env.REACT_APP_ENVIRONMENT === 'mainnet') {
+    cacheTokensList();
+    const intervalId = setInterval(() => cacheTokensList(), TIME_INTERVAL);
+    if (!IS_PROD_ENV) {
+      setTimeout(() => {
+        log.info('stopping caching tokens');
+        clearInterval(intervalId);
+      }, TIME_TO_TEST);
```
JST5000, line 134: Why do we want the test environment to wait for an hour here? (Is it simulating how long it would take to actually get the tokens, or something else?)

server/routes/v1/tokenDiscovery.js
```diff
@@ -3,12 +3,23 @@ const rippled = require('../../lib/rippled');
...
+// whether this is running in the prod environment (or in dev/staging)
+// for the purpose of running locally, this equals false if the env var doesn't exist
+// DO NOT SET TO TRUE UNLESS YOU'RE SURE ABOUT BIGQUERY USAGE
+// aka don't let the site run after you're done using it, because it'll cost $$
+const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK
+  ? process.env.REACT_APP_MAINNET_LINK.includes('xrpl.org')
+  : false;
+// how long the auto-caching should run in dev and staging environments
+// We want to turn it off after some time so it doesn't run when we don't need it, which costs us
+// money per BigQuery query
```
JST5000, line 15: Comment style nit: I think it's nice when comments follow normal sentence capitalization (So I think 'How' should be capitalized, but not 'money')

server/routes/v1/tokenDiscovery.js
```diff
@@ -114,21 +125,63 @@ async function cacheTokensList() {
...
+      // true if the cache has already been initialized (i.e. if `startCaching` has already run at
+      // least once)
+      // (prevents race conditions where an API call is made on container startup before the
+      // initial `startCaching` has returned)
+      cachedTokensList.time != null &&
+      // true if some time has passed since the cache was last updated (i.e. `startCaching` was
+      // turned off)
+      // this uses `TIME_INTERVAL * 2` to prevent race conditions around `TIME_INTERVAL`, in case
+      // the cache just took some time to load but `startCaching` has already been triggered
+      Date.now() - cachedTokensList.time > TIME_INTERVAL * 2
```
JST5000, line 175: In-lining comments in the if-statement makes it kinda hard to read. It might be better to pull out the relevant terms into variables with comments, then just combine them in the if-statement.

server/routes/v1/tokenDiscovery.js
```diff
@@ -114,21 +125,63 @@ async function cacheTokensList() {
...
+    ) {
+      // reset the cache so the API call waits below until it's been refilled
+      cachedTokensList.tokens = [];
+      cachedTokensList.time = null;
+      startCaching();
```
JST5000, line 180: While the cache is empty, is the site still able to show the old data? (Would this leave a window of slowness while the cache gets re-populated?)

JST5000: (If so it might be better to create a new copy, and only replace the old cache once the new one is fully created)

mvadari: Yes, but this is in the dev and staging environment so we don't care about UX

natenichols (approved):
LGTM 👍

server/routes/v1/tokenDiscovery.js
```diff
@@ -3,12 +3,23 @@ const rippled = require('../../lib/rippled');
...
+// for the purpose of running locally, this equals false if the env var doesn't exist
+// DO NOT SET TO TRUE UNLESS YOU'RE SURE ABOUT BIGQUERY USAGE
+// aka don't let the site run after you're done using it, because it'll cost $$
+const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK
+  ? process.env.REACT_APP_MAINNET_LINK.includes('xrpl.org')
+  : false;
+// how long the auto-caching should run in dev and staging environments
+// We want to turn it off after some time so it doesn't run when we don't need it, which costs us
+// money per BigQuery query
+const TIME_TO_TEST = 1000 * 60 * 60 * 1; // 1 hour
```
natenichols, line 16: We should consider making this like 8 hours or something, since it will still show the tables, it just wont update as often

florent-uzio (approved):
LGTM

server/routes/v1/tokenDiscovery.js
mvadari: This is how long the updating happens - so in this case it updates every half hour, but only for an hour (then it refreshes when someone calls the endpoint again).

mvadari: IMO we're not going to need to work with that page too much now that it's been built, and it's fine if the endpoint needs to be re-refreshed a couple hours after it's been deployed in dev or staging.

server/routes/v1/tokenDiscovery.js
```diff
@@ -3,12 +3,23 @@ const rippled = require('../../lib/rippled');
...
+// aka don't let the site run after you're done using it, because it'll cost $$
+const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK
+  ? process.env.REACT_APP_MAINNET_LINK.includes('xrpl.org')
+  : false;
+// How long the auto-caching should run in dev and staging environments
+// We want to turn it off after some time so it doesn't run when we don't need it, which costs us
+// money per BigQuery query
+const TIME_TO_TEST = 1000 * 60 * 60 * 1; // 1 hour
+
+const TIME_INTERVAL = 1000 * 60 * 30; // 30 minutes
```
sappenin, line 18:
```suggestion
const TIME_INTERVAL = 1000 * 60 * 180; // 180 minutes
```

I propose we make this call every 3 hours instead. After the changes pushed on 10/29/21, we've reduced our BQ costs from $2,822 per day to now around $500/day (i.e., by reducing the # of servers each calling BQ every 5 mins). However, this is still too costly in dollar terms, and 180 mins still feels reasonable from a UX perspective. Thoughts?

mvadari: We'll move it down another 6x for 30 minutes (so $80ish/day). Is that still too much in your opinion?

server/routes/v1/tokenDiscovery.js
```diff
@@ -3,12 +3,23 @@ const rippled = require('../../lib/rippled');
...
 const log = require('../../lib/logger')({ name: 'token discovery' });
 
-const TIME_INTERVAL = 1000 * 60 * 5; // 5 minutes
+// Whether this is running in the prod environment (or in dev/staging)
+// For the purpose of running locally, this equals false if the env var doesn't exist
+// DO NOT SET TO TRUE UNLESS YOU'RE SURE ABOUT BIGQUERY USAGE
+// aka don't let the site run after you're done using it, because it'll cost $$
+const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK
+  ? process.env.REACT_APP_MAINNET_LINK.includes('xrpl.org')
+  : false;
```
ledhed2222, line 12: this could be simpler as:

```javascript
const IS_PROD_ENV = process.env.REACT_APP_MAINNET_LINK?.includes('xrpl.org')
```

server/routes/v1/tokenDiscovery.js
```diff
@@ -114,21 +125,63 @@ async function cacheTokensList() {
...
 }
 
-if (!timerStarted && process.env.REACT_APP_ENVIRONMENT === 'mainnet') {
-  timerStarted = true;
-  cacheTokensList();
-  setInterval(() => cacheTokensList(), TIME_INTERVAL);
+// Starts the caching process for bigquery
+function startCaching() {
+  // Only run if on mainnet (the tokens page doesn't exist on devnet/testnet)
+  if (process.env.REACT_APP_ENVIRONMENT === 'mainnet') {
```
ledhed2222, line 131: since the body of the function is within the `if` you can remove an indentation by reversing this:

```javascript
function startCaching() {
  if (!process.env.REACT_APP_ENVIRONMENT === 'mainnet') {
    return
  }
  // rest of function
}
```

server/routes/v1/tokenDiscovery.js
ledhed2222: i'm with jackson

PR #35, opened 2021-11-01
bug: Better error handling for Tokens page

mukulljangid:
Add error messages for tokens page.

### Context of Change
Explorer wasn't getting any response from `rippled` because of an internal and response had `status: 200`, but data had the error.
Handling errors better.

### Type of Change

## Before / After
Before:
image
After:
image

## Test Plan
Passes all CI tests

### DIFF
--- a/public/locales/ko-KP/translations.json
+++ b/public/locales/ko-KP/translations.json
@@ -196,7 +196,5 @@
   "invalid_ripple_address": "잘못된 Ripple 주소입니다.",
   "tokens": "토큰",
   "total_issuers": "총 발행자",
-  "total_tokens": "총 토큰",
-  "error": "오류",
-  "internal error": "인터넷 서버 오류"
+  "total_tokens": "총 토큰"
 }
--- a/public/locales/ko-KP/translations.json
+++ b/public/locales/ko-KP/translations.json
@@ -196,5 +196,7 @@
   "invalid_ripple_address": "잘못된 Ripple 주소입니다.",
   "tokens": "토큰",
   "total_issuers": "총 발행자",
-  "total_tokens": "총 토큰"
+  "total_tokens": "총 토큰",
+  "error": "오류",
+  "internal error": "인터넷 서버 오류"
 }
--- a/src/containers/Tokens/index.js
+++ b/src/containers/Tokens/index.js
@@ -16,7 +16,6 @@ const Tokens = props => {
   const [allTokens, setAllTokens] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
   const [isError, setIsError] = useState(false);
-  const errorInfoRef = useRef({ title: '', hints: '' });
 
   const { t, error } = props;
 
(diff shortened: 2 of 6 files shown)


src/containers/Tokens/index.js
```diff
@@ -25,6 +27,17 @@ const Tokens = props => {
       .get(TOP_TOKENS_URL)
       .then(res => {
         const { tokens } = res.data;
+
+        if (res.data.result === 'error') {
```
mukulljangid, line 31: I was thinking that we could keep this as !== success, for all random cases? @natenichols @mvadari

public/locales/ko-KP/translations.json
```diff
@@ -196,5 +196,7 @@
   "invalid_ripple_address": "잘못된 Ripple 주소입니다.",
   "tokens": "토큰",
   "total_issuers": "총 발행자",
-  "total_tokens": "총 토큰"
+  "total_tokens": "총 토큰",
+  "error": "오류",
+  "internal error": "인터넷 서버 오류"
```
natenichols, line 201: Where did we get the translations?

src/containers/Tokens/index.js
natenichols: `res?.data?.result !== 'success'` should handle cases where either res or data are `undefined`

mvadari: Doesn't really matter - the response is either `success` or `error` for this endpoint

src/containers/Tokens/index.js
```diff
@@ -40,8 +53,13 @@ const Tokens = props => {
...
   }, []);
 
-  return error ? (
-    Tokens.renderError(error)
+  function renderError(errorInfo) {
+    const { result, message } = errorInfo;
+    return <NoMatch title={result} hints={[message]} />;
+  }
+
+  return error || isError ? (
```
natenichols, line 61: where is `error` defined here?

src/containers/Tokens/index.js
mvadari: Up to you :)

src/containers/Tokens/index.js
```diff
@@ -40,8 +53,13 @@ const Tokens = props => {
...
+    renderError(errorInfoRef.current)
```
mvadari, line 62: Why are we using a `ref` for this? We can just only render it if there's an error

src/containers/Tokens/index.js
mukulljangid: > `res?.data?.result !== 'success'` should handle cases where either res or data are `undefined`

Yes, that's better

src/containers/Tokens/index.js
mukulljangid: at the top, it is being deconstructed from props

natenichols (approved):

mvadari (approved):

PR #36, opened 2021-11-10
fix: actually fix the tokens page

mvadari:

This PR actually fixes the problems that https://livenet.xrpl.org/tokens was having. It turns out that the 40 queries that are needed for this page (4 for each of the 10 tokens) triggers the rate limiter. This PR runs those queries slightly more in series, instead of in parallel as it was doing before. It doesn't matter if it takes extra time, because this doesn't run very often.

### Context of Change

Tokens page went down with all the changes to our rippled nodes last week

### Type of Change

### TypeScript/Hooks Update

not worth the effort for a quick bugfix

## Test Plan

CI passes. Works locally.

### DIFF
--- a/server/routes/v1/tokenDiscovery.js
+++ b/server/routes/v1/tokenDiscovery.js
@@ -83,39 +83,59 @@ async function getTokensList() {
   options = { query, location: 'US' };
   const [rankedTokens] = await bigQuery.query(options);
 
-  const promises = [];
   for (let i = 0; i <= NUM_TOKENS_FETCH_ALL; i += 1) {
     const { issuer, currency } = rankedTokens[i];
+    const promises = [];
     promises.push(getAccountInfo(issuer, currency));
     promises.push(getExchangeRate(issuer, currency));
+    try {
+      const [accountInfo, exchangeRate] = await Promise.all(promises);
+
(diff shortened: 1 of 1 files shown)


natenichols (approved):

intelliot (approved):
Remind me, what are the TOTAL ISSUERS and TOTAL TOKENS numbers on the page actually counting? (A couple of question-mark explainers on the page may be useful)

Unrelated to this PR, it may be worth adding a note on the page that these tokens are not vetted nor endorsed by anyone, including XRPL.org, Ripple, nor the XPRLF. (Might want to check in with Legal or Comms for exact wording though)

Just ideas :)

mvadari:
> Remind me, what are the TOTAL ISSUERS and TOTAL TOKENS numbers on the page actually counting? (A couple of question-mark explainers on the page may be useful)

Looks like it's counting only the tokens with at least 50 trustlines.

> Unrelated to this PR, it may be worth adding a note on the page that these tokens are not vetted nor endorsed by anyone, including XRPL.org, Ripple, nor the XPRLF. (Might want to check in with Legal or Comms for exact wording though)

I think that falls under @ledhed2222 's purview.

ledhed2222 (approved):

PR #42, opened 2021-12-21
chore: switch from https/wss to http/ws

mvadari:

This PR switches the JSON RPC and WS connections from HTTPS and WSS to HTTP and WS.

### Context of Change

This is due to a change in the deployment situation.

### Type of Change

### TypeScript/Hooks Update

## Test Plan

CI passes. Works locally and in the dev env.

### DIFF
--- a/server/lib/rippled.js
+++ b/server/lib/rippled.js
@@ -4,8 +4,8 @@ const moment = require('moment');
 const utils = require('./utils');
 
 const HOSTNAME = os.hostname();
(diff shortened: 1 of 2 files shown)


natenichols (approved):

intelliot (changes requested):
Why don't we leave this as https/wss and use s1 or s2 instead?

mvadari:
> Why don't we leave this as https/wss and use s1 or s2 instead?

Because reporting mode requires an admin connection for `gateway_balances`.

khancode (approved):

intelliot (approved):

PR #46, opened 2022-01-05
fix: remove Top Trading Pairs table from non-mainnet Explorers

mvadari:

The `Top Trading Pairs` table shows up on a page for a token (e.g. https://livenet.xrpl.org/token/534F4C4F00000000000000000000000000000000.rsoLo2S1kiGeCcn6hCUXVrCpGMWLrRrLZz). It also uses the top tokens endpoint. This endpoint doesn't exist on testnet/devnet, and the data is also not useful on testnet/devnet, because BigQuery only contains data for mainnet.

### Context of Change

This should have been handled alongside https://github.com/xpring-eng/explorer/pull/97, but it was missed.

### Type of Change

### TypeScript/Hooks Update

## Before / After

Before:
image

After:
image

This change only happens on devnet/testnet; mainnet will still remain the same.

## Test Plan

CI passes. Fix is fairly simple.

### DIFF
--- a/src/containers/Token/index.js
+++ b/src/containers/Token/index.js
@@ -11,8 +11,7 @@ import NoMatch from '../NoMatch';
 import './styles.css';
 import { analytics, ANALYTIC_TYPES, NOT_FOUND, BAD_REQUEST } from '../shared/utils';
 
-const MODE = process.env.REACT_APP_ENVIRONMENT;
-const IS_MAINNET = MODE === 'mainnet';
+const IS_MAINNET = process.env.REACT_APP_ENVIRONMENT === 'mainnet';
 
 const ERROR_MESSAGES = {};
 ERROR_MESSAGES[NOT_FOUND] = {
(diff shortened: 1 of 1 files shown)


src/containers/Token/index.js
```diff
@@ -11,6 +11,9 @@ import NoMatch from '../NoMatch';
 import './styles.css';
 import { analytics, ANALYTIC_TYPES, NOT_FOUND, BAD_REQUEST } from '../shared/utils';
 
+const MODE = process.env.REACT_APP_ENVIRONMENT;
+const IS_MAINNET = MODE === 'mainnet';
```
JST5000, line 15:
```suggestion
const IS_MAINNET = process.env.REACT_APP_ENVIRONMENT === 'mainnet';
```
Since `MODE` isn't used anywhere as far as I can tell, it seems cleaner to do it in one line.

JST5000 (approved):

intelliot (approved):
Kinda-but-not-really related: https://github.com/xpring-eng/explorer/issues/106

PR #48, opened 2022-01-05
feat: set up routing to allow for sidechain URLs

mvadari:

Since sidechains are not one network (unlike mainnet/devnet/testnet), the format of sidechain Explorer URLs is going to be `sidechain.xrpl.org/[sidechain_url]/...`, instead of `livenet.xrpl.org/...`. The current routing infrastructure doesn't support those URLs (the links don't direct properly). This PR modifies the routing structures to support those URLs, while remaining backwards compatible so that it can still be used for mainnet/devnet/testnet.

#### Technical notes:
* The main new files are `UrlContext` and `AppWrapper`
  * `UrlContext` is the actual [React context](https://reactjs.org/docs/context.html) that is used
  * `AppWrapper` is a wrapper around `App` that handles the top-level sidechain-specific routing
* `App` was also edited pretty heavily, to handle the new routing and the new context (alongside the new `AppWrapper` - the old `App/index.js` was moved to `App/App.js`, but git isn't good at handling those changes, so it's not as much change as it looks like)

### Context of Change

This is the first step of getting sidechain support on the Explorer.

### Type of Change

### TypeScript/Hooks Update

The new files are in TypeScript and using hooks (where possible), but the old files weren't touched in this respect (since I touched too many of them).

## Before / After

You can now go to `[explorer-url]/[randomtextthatisntusedyet]` when in sidechain mode.

## Test Plan

CI passes.


src/containers/Accounts/AccountHeader/index.js
```diff
@@ -170,7 +170,8 @@ class AccountHeader extends Component {
             <li key={xAddress.classicAddress}>
               <span className="classicAddress">
                 <span className="label">Classic Address: </span>
-                <Account account={xAddress.classicAddress} /> {/* TODO: Add [copy] button */}
+                <Account account={xAddress.classicAddress} parentPath={parentPath} />{' '}
+                {/* TODO: Add [copy] button */}
```
ledhed2222, line 174: what is this?

mvadari: Not sure - it was there before, the linter just moved it on save

ledhed2222:
two questions:

1) do we need to have `parentPath` as a prop at all? Why can't it be determined from the URL?
2) if we do need to have `parentPath`, since it is likely to be needed throughout the app, it should use React "context" https://reactjs.org/docs/context.html

mvadari:
> two questions:
> 
> 1. do we need to have `parentPath` as a prop at all? Why can't it be determined from the URL?
> 2. if we do need to have `parentPath`, since it is likely to be needed throughout the app, it should use React "context" https://reactjs.org/docs/context.html

1. It can (sort of) be done with the URL, but it gets tricky (e.g. how do you distinguish between mainchain.xrpl.org/network and sidechain.xrpl.org/sidechain_url)? Here's how I did it for the header bar: https://github.com/ripple/explorer/pull/48/files#diff-0f7d55aa457d17d6eea6a2de9056f7235b9255b07fa0b06cd10f7bc29391b49fR27-R48 (which isn't really future-proof)
2. I like the idea of contexts - I think that's probably the way to do it.

src/containers/Token/DEXPairs/index.js
```diff
@@ -122,7 +124,7 @@ const DEXPairs = props => {
         </td>
         <td className="issuer-address">
           {pair.issuer !== undefined ? (
-            <a href={`/token/${pair.token}.${pair.issuer}`}>{pair.issuer}</a>
+            <a href={`${urlLink}/token/${pair.token}.${pair.issuer}`}>{pair.issuer}</a>
```
mvadari, line 127: Tests were timing out when I changed this to an `InternalLink`, and I couldn't figure out what was wrong, so I left it as an href.

mvadari:
It's currently failing tests because it doesn't quite test 70% of branches. This isn't important to resolve right now, and would likely make the PR large, and add code that is unrelated to this PR. Since this isn't being merged into `staging`, it's fine (for now) if the threshold doesn't pass.

khancode:
> * do we need to have `parentPath` as a prop at all? Why can't it be determined from the URL?

I agree with Greg to use React context to determine parentPath.

mvadari:
> > * do we need to have `parentPath` as a prop at all? Why can't it be determined from the URL?
> 
> I agree with Greg to use React context to determine parentPath.

The current version of this PR does.

khancode (approved):
LGTM 👍

ledhed2222 (approved):

PR #49, opened 2022-01-06
Add logic to use a regular node for gateway_balances

JST5000:

We'd like to use a reporting mode server for most things, but `gateway_balances` is an admin only command, so this PR adds functionality to add another server for admin commands.

### Context of Change

Part of upgrading the reliability of the nodes the explorer depends on.

### Type of Change

### TypeScript/Hooks Update

Did not update to TypeScript because I didn't realize that was a goal and have other priorities this week.

## Before / After

* Before all queries were sent to one URL
* Now if you specify a P2P url, it will direct `gateway_balances` to that url (Defaults to the old behavior)

## Test Plan

Manually building the code with and without the new environment variable and with different RIPPLED_HOSTs.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ RIPPLED_WS_PORT=51233
 RIPPLED_RPC_PORT=51234
 RIPPLED_PEER_PORT=51235
 RIPPLED_HOST=your-rippled-host-url-here
+P2P_RIPPLED_HOST=your-rippled-host-that-can-run-admin-only-requests-url-here
 RIPPLED_SECONDARY=s1.ripple.com,s2.ripple.com,r.ripple.com,rippled.xrptipbot.com
 
 REACT_APP_GA_ID=
--- a/.env.example
+++ b/.env.example
@@ -3,7 +3,7 @@ RIPPLED_WS_PORT=51233
 RIPPLED_RPC_PORT=51234
 RIPPLED_PEER_PORT=51235
 RIPPLED_HOST=your-rippled-host-url-here
-#s1 and s2 are setup to make gateway_balances and account_lines admin only
+#s1 and s2 are set up to make gateway_balances and account_lines admin only
 P2P_RIPPLED_HOST=your-rippled-host-that-can-run-gateway_balances-url-here
 RIPPLED_SECONDARY=s1.ripple.com,s2.ripple.com,r.ripple.com,rippled.xrptipbot.com
 
--- a/.env.example
+++ b/.env.example
@@ -3,7 +3,8 @@ RIPPLED_WS_PORT=51233
 RIPPLED_RPC_PORT=51234
 RIPPLED_PEER_PORT=51235
 RIPPLED_HOST=your-rippled-host-url-here
-P2P_RIPPLED_HOST=your-rippled-host-that-can-run-admin-only-requests-url-here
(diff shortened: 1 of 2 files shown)


mvadari (approved):
Nice work!

mvadari:
You should update `.env.example` with your new env variable.

server/lib/rippled.js
```diff
@@ -37,16 +42,24 @@ const formatPaychannel = d => ({
...
   });
 };
 
+// generic RPC query
+function query(...options) {
+  return executeQuery(URL, ...options);
+}
+
+function adminQuery(...options) {
+  return executeQuery(P2P_URL, ...options);
```
natenichols, line 60: Why is adminQuery just query `P2P_URL`?

Is the P2P server (whichever P2P refers to going to open an `admin` port? 

Or is this just for gateway balances?

natenichols: I'm guessing (and hoping) the latter.

server/lib/rippled.js
```diff
@@ -37,16 +42,24 @@ const formatPaychannel = d => ({
...
+    throw new utils.Error(`URL: ${url} - ${message}`, code);
   });
 };
 
+// generic RPC query
+function query(...options) {
+  return executeQuery(URL, ...options);
+}
+
+function adminQuery(...options) {
```
natenichols, line 59: I would call this `queryP2P` and `queryFH`.

JST5000: What does FH stand for?

mvadari: full history

mvadari: perhaps `queryReporting` and `queryP2P`? Or even `query` and `queryP2P`?

JST5000: The Full History is true in this specific case, but not true in general (If the config changes, that could easily become false)

JST5000: I think `query` and `queryP2P` makes the most sense to me

.env.example
```diff
@@ -3,6 +3,7 @@ RIPPLED_WS_PORT=51233
 RIPPLED_RPC_PORT=51234
 RIPPLED_PEER_PORT=51235
 RIPPLED_HOST=your-rippled-host-url-here
+P2P_RIPPLED_HOST=your-rippled-host-that-can-run-admin-only-requests-url-here
```
mvadari, line 6: You should add a note here that s1/s2 are set up to make account_lines and gateway_balances admin-only (it's not reporting-specific)

.env.example
```diff
@@ -3,6 +3,8 @@ RIPPLED_WS_PORT=51233
 RIPPLED_RPC_PORT=51234
 RIPPLED_PEER_PORT=51235
 RIPPLED_HOST=your-rippled-host-url-here
+#s1 and s2 are setup to make gateway_balances and account_lines admin only
```
mvadari, line 6: nit:
```suggestion
#s1 and s2 are set up to make gateway_balances and account_lines admin only
```

intelliot (approved):
Idea for the future:
- Make gateway_balances and account_lines more efficient/scalable so that they don't need to be admin-only

This is probably a task for the Clio / rippled team though.

mvadari:
> This is probably a task for the Clio / rippled team though.

I assume they're working on it and this will be removed at some point

PR #51, opened 2022-01-06
chore: remove CI on node v13.x

mvadari:

Title says it all. While I was here I renamed the job to be more descriptive.

### Context of Change

We don't use node 13.x anymore. It's also deprecated and we shouldn't be using it anymore.

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

CI passes.


.github/workflows/nodejs.yml
```diff
@@ -10,12 +10,12 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  build-and-test:
```
JST5000, line 13: Why does this command need to change?

mvadari: That's just the name of the job. I renamed it to be more descriptive. It's not just building, it's building and testing.

JST5000 (approved):

intelliot (approved):
Would it make sense to add Node.js 16.x or 17.x now?

mvadari:
> Would it make sense to add Node.js 16.x or 17.x now?

Yes, but that was failing (you can see the results in the commit history of this PR) and it wasn't worth the time to figure out what was going wrong right now.

PR #53, opened 2022-01-07
add nft env to dropdown

ledhed2222:

adds a dropdown option for the NFT sandbox env. also refactors the header markup and styling to be more extensible, using flexbox.

### Context of Change

we need to launch the NFT sandbox as it's own explorer instance

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Future Tasks

the header dropdown for env selection is a little wonky now. it needs to be revisited. specificially, the border around the dropdown and the hover color don't work great


public/locales/en-US/translations.json
```diff
@@ -1,7 +1,8 @@
 {
   "testnet_data": "testnet data",
-  "live_data": "live data",
+  "mainnet_data": "live data",
   "devnet_data": "devnet data",
+  "nft_sandbox_data": "NFT sandbox data",
```
mvadari, line 5: IMO this should be lowercase to match the other 3

ledhed2222:
TODO make sure that network pages don't show for sandbox

src/containers/Header/index.js
```diff
@@ -38,38 +41,17 @@ class Header extends Component {
...
-          });
-          window.location = MAINNET_LINK;
-          break;
-      }
+
+    if (mode === MODE) {
+      return;
     }
+
+    const desiredLink = ENV_LINK_MAP[mode] ?? process.env.REACT_APP_MAINNET_LINK;
```
JST5000, line 49: Should this output a log if the desired link doesn't match anything? That seems like an important thing to know about when setting up an environment.

JST5000 (approved):

src/containers/Header/index.js
JST5000: Verdict - yep that probably makes sense, but is beyond the scope of this ticket

public/locales/en-US/translations.json
mvadari: Discussed offline, this doesn't work because of the way capitalization happens (it would be capitalized as `Nft`)

mvadari (approved):

PR #66, opened 2022-01-12
refactor: move rippled connections from server to client

mvadari:

This PR moves the connections to rippled nodes from the server side to the client side.

### Context of Change

This simplifies the logic of the Explorer system. The backend was previously just maintaining a connection to rippled, and was just converting rippled output to a different JSON output, which is unnecessary.

This change is necessary for sidechain work.

### Type of Change

### TypeScript/Hooks Update

This PR touches a lot of files. It wasn't worth the effort to fix all of them.

## Before / After

Before, the rippled connections were all done by the server.

After, the rippled connections are all done by the client.

## Test Plan

#69 will be merged into this PR before it is merged into `staging`. I separated them out into 2 PRs because rewriting tests required a lot of rewriting the mocking framework, so this makes it easier to review.

## Future Tasks
* Improve logging, so that there aren't two different types of logging


server/lib/rippled.js
```diff
@@ -1,41 +1,10 @@
 const os = require('os');
 const axios = require('axios');
-const moment = require('moment');
 const utils = require('./utils');
 
 const HOSTNAME = os.hostname();
 const URL = `http://${process.env.RIPPLED_HOST}:${process.env.RIPPLED_RPC_PORT}`;
 const URL_HEALTH = `http://${process.env.RIPPLED_HOST}:${process.env.RIPPLED_PEER_PORT}/health`;
-const N_UNL_INDEX = '2E8A59AA9D3B5B186B0B9E0F62E6C02587CA74A4D778938E957B6357D364B244';
```
mvadari, line 9: Everything deleted here was moved to `src/rippled/lib/rippled.js`

khancode (approved):
LGTM

src/containers/Token/DEXPairs/index.js
```diff
@@ -11,13 +11,14 @@ import {
...
 import PairStats from './PairStats';
 
 // Hard Coded Pairs that we always check for
 const pairsHardCoded = [
-  'XRP',
-  'USD.rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B', // Bitstamp USD
-  'BTC.rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B', // Bitstamp BTC
+  { currency: 'XRP' },
+  { currency: 'USD', issuer: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B' }, // Bitstamp USD
+  { currency: 'BTC', issuer: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B' }, // Bitstamp BTC
```
ledhed2222, line 21: 👍

src/containers/Token/index.js
```diff
@@ -67,6 +67,7 @@ class Token extends Component {
     const currency = match.params.currency || '';
     const showError = accountId === prevId && error;
 
+    // TODO: title this so it includes the currency code somehow
     document.title = `${t('xrpl_explorer')} | ${accountId.substr(0, 12)}...`;
```
ledhed2222, line 71: why can't you add the code already?

src/containers/shared/utils.js
```diff
@@ -23,7 +23,7 @@ export const BAD_REQUEST = 400;
 export const DECIMAL_REGEX = /^\d+$/;
 export const RIPPLE_ADDRESS_REGEX = /^r[rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz]{27,35}$/;
 export const HASH_REGEX = /[0-9A-Fa-f]{64}/i;
-export const CURRENCY_REGEX = /^[a-zA-Z]{3,}\.r[rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz]{27,35}$/;
+export const CURRENCY_REGEX = /^[a-zA-Z0-9]{3,}\.r[rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz]{27,35}$/;
```
ledhed2222, line 26: 👍  - wondering though whether we can get these regexes from `xrpl`

src/rippled/accountTransactions.js
```diff
@@ -7,15 +7,12 @@
...
-const summarize = require('../../lib/txSummary');
-const log = require('../../lib/logger')({ name: 'account transactions' });
-
-module.exports = (req, res) => {
-  const { id: account, currency } = req.params;
-  const { limit, marker } = req.query;
+const utils = require('./lib/utils');
+const rippled = require('./lib/rippled');
+const summarize = require('./lib/txSummary');
+const log = require('./lib/logger')({ name: 'account transactions' });
```
ledhed2222, line 13: these could be `imports` instead

src/rippled/lib/rippled.js
```diff
@@ -0,0 +1,351 @@
+const os = require('os');
+const axios = require('axios');
+const moment = require('moment');
+const utils = require('./utils');
```
ledhed2222, line 4: can be `imports`

src/containers/Token/index.js
mvadari: I spent a couple minutes thinking about it and couldn't really come up with a good way to do this for the currency codes that were just hex (aka more than 3 characters).

src/containers/shared/utils.js
mvadari: That's a good idea - I honestly just modified these regexes myself

src/rippled/lib/txSummary/NFTokenMint.js
```diff
@@ -1,4 +1,6 @@
-const { convertHexToString } = require('xrpl');
+function convertHexToString(hex, encoding = 'utf8') {
+  return Buffer.from(hex, 'hex').toString(encoding);
+}
```
ledhed2222, line 3: put this in a utils folder please - also put a comment indicating that it will not be required once we do `xrpl` correctly

ledhed2222 (approved):

ledhed2222 (approved):

.github/workflows/nodejs.yml
```diff
@@ -27,3 +27,5 @@ jobs:
     - run: npm run test:all
       env:
         REACT_APP_VALIDATOR: vl.ripple.com
+        REACT_APP_RIPPLED_WS_PORT: 51233
+        REACT_APP_RIPPLED_HOST: fake.ripple.com
```
intelliot, line 31: Is fake.ripple.com just made up? If so, I would prefer to use an `example.com` host (since that's a reserved domain name) so that we don't think there is a mock server or something set up at fake.ripple.com

```suggestion
        REACT_APP_RIPPLED_HOST: fake.rippled.example.com
```

intelliot (approved):
nice refactor!

PR #70, opened 2022-01-19
Add sentinel value of 'unused' for P2P_RIPPLED_HOST

JST5000:

Since we can't add empty environment variables in our gitlab deployment, and using yaml `null` variables causes issues as well, we need a sentinel value for "Don't use this variable" for the P2P_RIPPLED_HOST variable.

### Context of Change

Created as a follow up to #49 

### Type of Change

### TypeScript/Hooks Update

Other tasks are higher priority than updating this file currently. 

## Before / After

* P2P_RIPPLED_HOST with a value of 'unused' now also defaults to using the RIPPLED_HOST value 

## Test Plan

Manual local testing
Testing in dev deployment environments in conjunction with the environment variable changes.

### DIFF
--- a/server/lib/rippled.js
+++ b/server/lib/rippled.js
@@ -7,9 +7,11 @@ const HOSTNAME = os.hostname();
 const URL = `http://${process.env.RIPPLED_HOST}:${process.env.RIPPLED_RPC_PORT}`;
 const URL_HEALTH = `http://${process.env.RIPPLED_HOST}:${process.env.RIPPLED_PEER_PORT}/health`;
 // If there is a separate peer to peer server for admin requests, use it. Otherwise use the default url for everything.
-const P2P_URL = process.env.P2P_RIPPLED_HOST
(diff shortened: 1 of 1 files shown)


mvadari:
Can we use an empty string or something instead of something like `unused`? I don't like the syntax of using a word like that.

JST5000:
> Can we use an empty string or something instead of something like `unused`? I don't like the syntax of using a word like that.

Unfortunately I couldn't figure out a nice way to do it. I tried using `null`, having it empty, "", and removing the variable entirely from the keypair list, but none of them worked.

mvadari:
> The .yaml file interprets empty strings, and null, into a special value of `nil` which then is not accepted by Helm. So, we need a sentinel value which .yaml won't turn to `nil` (See 3rd answer for reference on what values get turned into nil: https://stackoverflow.com/questions/34089496/empty-field-in-yaml)

Can we instead have an `or` somewhere in the helm script that helps it evaluate to `null`?

JST5000:
No longer needed (Solved in a better way :) )

PR #76, opened 2022-01-31
feat: use the info from the sidechain URL to connect to the sidechain

mvadari:

This PR builds upon #48 to actually use the rippled URL in the Explorer URL to connect. If you go to `[explorer-url.net]/[rippled-url.net]` in sidechain mode, it will use the rippled URL as the node to connect to.

Note: this currently only works for nodes with the standard JSON and WS port configurations (51234 and 51233 respectively), and will not work with reporting mode (because that has `gateway_balances` disabled).

This code is also backwards compatible, so it works for the other network setups (mainnet/testnet/devnet) as well.

This branch is currently deployed to the dev env (https://sidechain.dev.explorer.xpring.dev) if you want to play with it.

### Context of Change

sidechain explorer work

### Type of Change

### TypeScript/Hooks Update

Not worth here

## Before / After

No visible changes.

## Test Plan

CI passes (except for the branch threshold). Tested in the dev environment and it worked as expected.

### DIFF
--- a/src/containers/Accounts/AccountHeader/actions.js
+++ b/src/containers/Accounts/AccountHeader/actions.js
@@ -3,7 +3,7 @@ import { getAccountState } from '../../../rippled';
 import { analytics, ANALYTIC_TYPES, BAD_REQUEST } from '../../shared/utils';
 import * as actionTypes from './actionTypes';
 
-export const loadAccountState = accountId => dispatch => {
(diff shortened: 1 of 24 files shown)


khancode (approved):
LGTM

ledhed2222 (approved):

PR #89, opened 2022-02-08
fix: fetch ledger metrics from the backend

mvadari:

One consequence of #66 is that the ledger metrics at the top of the Explorer only use the ledgers that the frontend has fetched. This doesn't provide very accurate numbers, especially within the first few ledgers that it loads. This PR fixes that, and fetches ledger metrics from the backend, which pays attention to all ledgers.

This PR re-adds a rippled WS connection to the backend, to get the ledger information. This re-adds some code that was here before #66. You can take a look at staging for what the previous versions looked like.

### Context of Change

Cleanup from #66

### Type of Change

### TypeScript/Hooks Update

## Test Plan

Tests pass. When tested locally, the metrics work as expected.


server/lib/rippled.js
```diff
@@ -105,3 +105,31 @@ module.exports.getHealth = async () => {
     }
   });
 };
+
+module.exports.getLedger = parameters => {
```
mvadari, line 109: This was here previously.

server/lib/utils.js
```diff
@@ -1,3 +1,33 @@
+const moment = require('moment');
```
mvadari, line 1: All of this was here before - I simplified `summarizeLedger` a bit because it did more processing than needed here.

server/lib/xrpl-ws.js
```diff
@@ -0,0 +1,84 @@
+const WebSocket = require('ws');
```
mvadari, line 1: This was all here previously - I just removed the `server` and `validations` subscription streams.

server/lib/streams.js
```diff
@@ -0,0 +1,109 @@
...
+
+    if (d.total_fees) {
+      fees += d.total_fees;
+      txCount += d.txn_count ?? 0;
+      ledgerCount += 1;
+    }
+  });
+  currentMetric.base_fee = Number(baseFee.toPrecision(4)).toString();
+  currentMetric.txn_sec = time && txCount ? ((txCount / time) * 1000).toFixed(2) : undefined;
+  currentMetric.txn_ledger = ledgerCount ? (txCount / ledgerCount).toFixed(2) : undefined;
```
JST5000, line 93: Should this also check for `txCount` like the line before?
```suggestion
  currentMetric.txn_ledger = ledgerCount && txCount ? (txCount / ledgerCount).toFixed(2) : undefined;
```

server/lib/rippled.js
JST5000: Any idea why these files are appearing as additions if they were here before?

server/lib/streams.js
```diff
@@ -0,0 +1,109 @@
...
+  currentMetric.ledger_interval = timeCount ? (time / timeCount / 1000).toFixed(3) : undefined;
+  currentMetric.avg_fee = txCount ? (fees / txCount).toPrecision(4) : undefined;
```
JST5000, line 95: Do the precisions for base_fee and avg_fee need to be the same? (If so defining a constant might help keep things in sync)

server/lib/rippled.js
mvadari: It's here on `staging`, but is removed in `move-rippled`, so it's supposed to show up as an addition

server/lib/streams.js
mvadari: `txCount` can be 0

server/lib/utils.js
```diff
@@ -1,3 +1,33 @@
...
+    .utc()
+    .format();
+
+module.exports.summarizeLedger = (ledger, txDetails = false) => {
+  const summary = {
+    ledger_index: Number(ledger.ledger_index),
+    ledger_hash: ledger.ledger_hash,
+    parent_hash: ledger.parent_hash,
+    close_time: convertRippleDate(ledger.close_time),
+    total_xrp: ledger.total_coins / 1000000,
```
JST5000, line 19: Might be good to define a constant for `100000` to avoid any off-by-one-zero typos.

server/lib/streams.js
mvadari: The precisions are the same as what is currently used by the Explorer

JST5000: Should that be tracked somewhere central so that they stay in sync then? e.g. a constants file?

server/lib/utils.js
mvadari: turns out the constant already exists and just wasn't used lol

server/lib/streams.js
mvadari: This is where the Explorer is getting that information.

JST5000 (approved):
LGTM!

PR #90, opened 2022-02-11
feat: add sidechains to network toggle

mvadari:

This PR adds sidechains to the network toggle on the top left of the Explorer. This allows users to switch between sidechain networks and other networks on the Explorer more easily.

### Context of Change

Sidechain MVP

### Type of Change

### TypeScript/Hooks Update

## Before / After

Before:
image

After:
(in testnet mode)
image

(in sidechain mode)
image

## Test Plan

CI passes. I tested this in the dev env and made sure everything works as expected.


src/containers/Header/index.js
```diff
@@ -46,29 +66,104 @@ class Header extends Component {
...
+        onClick={handleClick}
+        onKeyUp={handleClick}
+        tabIndex={0}
+      >
+        <span>{text}</span>
+        {/*
+        Adding this space to provide width between name
+        and selected checkmark
+        */}
+        &nbsp;
```
khancode, line 127: Why not just add css padding between name and checkmark?

khancode (changes requested):

src/containers/Header/index.js
mvadari: @ledhed2222 added this

ledhed2222: seemed semantic? 🤷

ledhed2222 (approved):

src/containers/Header/index.js
khancode: Haha ok. I'll approve but we should probably fix this in another PR.

PR #92, opened 2022-02-15
fix: remove server stream

mvadari:

This PR removes the `server` stream from the WS connection. The `server` stream is not supported by reporting mode. It was being used to figure out the load fee metric, so that calculation is redone with a call to `server_info` instead.

### Context of Change

The `server` stream is not supported by reporting mode. This meant that sometimes, when connecting to s2 via WS, we would run into issues 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Test Plan

Tested by hand. Worked successfully.

### DIFF
--- a/src/rippled/lib/streams.js
+++ b/src/rippled/lib/streams.js
@@ -1,4 +1,4 @@
-import { getLedger } from './rippled';
+import { getLedger, getServerInfo } from './rippled';
 import { summarizeLedger, EPOCH_OFFSET } from './utils';
 import logger from './logger';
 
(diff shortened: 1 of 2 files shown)


src/rippled/lib/streams.js
```diff
@@ -44,6 +44,16 @@ const fetchLedger = (ledger, attempts = 0) => {
     });
 };
 
+const fetchServerInfo = () => {
+  return getServerInfo()
+    .then(result => result.info)
+    .then(info => {
```
JST5000, line 50: Can these two lines be combined into one? 

ie `.then(result => { info = result.info ... })`

src/rippled/lib/streams.js
```diff
@@ -44,6 +44,16 @@ const fetchLedger = (ledger, attempts = 0) => {
     });
 };
 
+const fetchServerInfo = () => {
+  return getServerInfo()
```
JST5000, line 48: Instead of `server info`, would it make more sense to use the `fee` request? - https://xrpl.org/fee.html#fee

src/rippled/lib/streams.js
```diff
@@ -44,6 +44,16 @@ const fetchLedger = (ledger, attempts = 0) => {
     });
 };
 
+const fetchServerInfo = () => {
```
JST5000, line 47: Since this just returns the fee, maybe it should have a different name like `fetchServerFee`?

src/rippled/lib/streams.js
mvadari: `fee` doesn't have the load factor. This is why we use `server_info` in xrpl.js for this calculation: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/sugar/getFeeXrp.ts#L39

JST5000 (approved):
LGTM!

PR #96, opened 2022-02-17
chore: add "coming soon" placeholder on network page for other networks

mvadari:

The "networks" page doesn't exist yet for networks that aren't mainnet/devnet/testnet (e.g. sidechains and the NFT-Devnet. So instead of displaying info there that doesn't mean anything/could be misleading, this PR displays a generic error message telling people that the page isn't ready yet.

### Context of Change

The VHS doesn't yet have support for networks that aren't mainnet/devnet/testnet (e.g. sidechains and the NFT-Devnet).

### Type of Change

### TypeScript/Hooks Update

## Before / After

image

## Test Plan

CI passes. Tested locally and it worked as expected.

### DIFF
--- a/public/locales/en-US/translations.json
+++ b/public/locales/en-US/translations.json
@@ -136,6 +136,8 @@
   "transaction_SetFee": "Set Fee",
   "generic_error": "Something bad happened",
   "not_your_fault": "It's probably not your fault",
+  "under_construction": "This page doesn't exist right now",
+  "come_back_later": "It will be available soon",
   "invalid_ledger_id": "The ledger id is invalid",
   "invalid_transaction_hash": "The transaction hash is invalid",
   "ledger_not_found": "Ledger not Found",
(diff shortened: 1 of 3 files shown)


JST5000:
Would it be better to just take away the tab like we do with `tokens` on all but mainnet? (That way users don't wonder if they messed up)

mvadari:
> Would it be better to just take away the tab like we do with `tokens` on all but mainnet? (That way users don't wonder if they messed up)

That feels more confusing to me, because this isn't a permanent change (hence the "it will be available soon" message). We had a few issues opened up on Github for the Explorer when we removed the "Tokens" tab for the other networks, because people were confused about why the tab wasn't there. But I'm open to further discussion/opinions.

JST5000:
> That feels more confusing to me, because this isn't a permanent change (hence the "it will be available soon" message). We had a few issues opened up on Github for the Explorer when we removed the "Tokens" tab for the other networks, because people were confused about why the tab wasn't there. But I'm open to further discussion/opinions.

That makes sense to me. Is it possible to hide the "Uh Oh" since it's not really an error? (If it's a huge lift I think the PR is fine as is, but if it's easy to do I think that'll be less confusing if folks end up there by accident)

ledhed2222 (approved):

PR #97, opened 2022-02-18
refactor: use a sustained rippled WS connection on the frontend

mvadari:

This PR refactors the frontend code to use a sustained WS connection to rippled, instead of using a WS connection created just for the streams and a JSON connection for all non-subscribe requests. 

Main changes:
* I refactored `App.js` to create an `XrplClient` instance that is put into the context (which now contains an XrplClient instance instead of just the rippled URL).
* I renamed all instances of `rippledUrl`/`url` that now refer to the WS connection.
* I replaced all instances of `UrlContext` with the new `SocketContext`.
* I refactored `Streams` to use the WS connection in the context instead of creating a new WS connection.
* I refactored `rippled.js` to use the WS connection in the context instead of using a JSON RPC connection.

Renaming `rippledUrl`/`url` to `rippledSocket` and replacing `UrlContext` with `SocketContext` are 29/32 of the modified files.

### Context of Change

Using JSON RPC connections on the frontend was meant to be a placeholder. 

### Type of Change

### TypeScript/Hooks Update

Too many files were touched for this to be worth it.

## Before / After

No visible changes.

## Test Plan

I tested locally and everything worked well. Tests are fixed in #107, to make reviewing easier.


khancode (approved):
LGTM

PR #99, opened 2022-02-24
chore: switch AccountHeader to hooks

mvadari:

This PR updates `AccountHeader` to use hooks instead of a component.

### Context of Change

repo cleanup

### Type of Change

### TypeScript/Hooks Update

This PR is about updating to hooks. I didn't do TS because this was just something quick I did with a few spare minutes.

## Before / After

No change.

## Test Plan

CI passes.

## Future Tasks
* Use TypeScript
* Remove the use of actions (I spent some time trying to do this, but this will be trickier since it looks like the banner uses the action output as well)


JST5000 (approved):
LGTM

khancode (approved):
LGTM 👌

PR #104, opened 2022-03-15
feat: add sidechain MVP support to Explorer

mvadari:
## High Level Overview of Change

This PR adds basic sidechain support to the Explorer.

This contains the following PRs (which have more detailed information):
* #48 (sets up routing to handle `sidechain.xrpl.org/[sidechain_url]` URLs)
* #76 (uses the URL listed in `sidechain.xrpl.org/[sidechain_url]` to connect to an XRPL node)
* #90 (adds sidechains to the dropdown network toggle)
* #98 (adds a home page for sidechains if no network is selected yet)

### Context of Change

sidechain Explorer support

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

### TypeScript/Hooks Update

All new files use hooks and TS, but old files were not modified.

- [x] Updated files to React Hooks
- [x] Updated files to TypeScript

## Before / After

See smaller PRs for more details.

## Test Plan

CI passes.

PR #105, opened 2022-03-17
fix: re-add sidechain dropdown to non-sidechain explorers

mvadari:

This PR re-adds the sidechain input to the network dropdown in the mainnet/devnet/testnet/NFT-devnet versions of the explorer.

### Context of Change

Because sidechains were deployed for the first time in #104, it was recommended that we hide the sidechain-related features from the already-existing networks, because there would be some lag between deploying and having the URL all set up for sidechains.

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

Also uncommented a test. CI passes.

### DIFF
--- a/src/containers/Header/index.js
+++ b/src/containers/Header/index.js
@@ -204,8 +204,7 @@ class Header extends Component {
                       t(`${network}_data`)
                     );
               })}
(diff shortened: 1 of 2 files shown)


khancode (approved):
LGTM

PR #106, opened 2022-03-24
fix: failover to `ws` if `wss` fails

mvadari:

This PR adds a failover for the streams, so that if `wss` fails, then it attempts to use `ws`.

### Context of Change

Some sidechains may not have certs set up, and may have only `ws` set up. But the Explorer currently expects them to be `wss`.

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

CI passes. Tested locally and it works

## Future Tasks
The startup for the stream takes a bit of time. We should see if we can make it faster somehow.

### DIFF
--- a/src/containers/shared/components/Streams.jsx
+++ b/src/containers/shared/components/Streams.jsx
@@ -231,18 +231,29 @@ class Streams extends Component {
   connect() {
     const rippledUrl = this.context;
     const rippledWsUrl = `wss://${rippledUrl}:${process.env.REACT_APP_RIPPLED_WS_PORT}`;
-    this.ws = new WebSocket(rippledUrl == null ? DEFAULT_WS_URL : rippledWsUrl);
+    this.startWs(rippledUrl == null ? DEFAULT_WS_URL : rippledWsUrl);
+  }
+
+  startWs(url) {
+    this.ws = new WebSocket(url);
     this.ws.last = Date.now();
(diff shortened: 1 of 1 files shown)


src/containers/shared/components/Streams.jsx
```diff
@@ -230,14 +228,22 @@ class Streams extends Component {
...
+    const wsUrl = `${connection}://${rippledUrl}:${process.env.REACT_APP_RIPPLED_WS_PORT}`;
+    this.ws = new WebSocket(wsUrl);
     this.ws.last = Date.now();
     Log.info(`connecting...`);
 
     // handle error
-    this.ws.onclose = () => {
+    this.ws.onclose = e => {
       Log.warn(`ws closed`);
+      if (!e.wasClean && wss) {
```
JST5000, line 244: nit: I think naming `wss` to `useWss` would be a bit more clear so people don't accidentally think `wss` stands for a websocket url.

JST5000 (approved):
LGTM

mvadari:
Does not work in prod:
```
Mixed Content: The page at '<URL>' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws:<URL>/'. This request has been blocked; this endpoint must be available over WSS.
```
This problem might require using the backend as a WS proxy

PR #116, opened 2022-04-18
feat: Add gzipping and cache busting to assets

ckniffen:

When running server in prod mode all static assets will be gzipped.
Additionally all .js and .css will have versioned hashes and set cache headers.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Check the headers of .js and .css files.  If they contain a 20 character has they will be served up with `Cache-Control: max-age=31536000`. Html files will have `Cache-Control: no-cache`


mvadari (changes requested):
This looks awesome - it improves performance so much!

It looks like your PR is changing the package-lock.json back to `lockfileVersion 1` instead of `2`, so your npm might need to be updated to avoid that change.

JST5000 (approved):
LGTM

ckniffen:
> This looks awesome - it improves performance so much!
> 
> It looks like your PR is changing the package-lock.json back to `lockfileVersion 1` instead of `2`, so your npm might need to be updated to avoid that change.

Done.

mvadari (approved):

PR #129, opened 2022-04-26
fix: get sidechain ledger metrics to show up properly

mvadari:
## High Level Overview of Change

This PR fixes some bugs in how the metrics for sidechains are displayed. See Before/After for more details.

### Context of Change

I noticed these bugs while poking around

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### TypeScript/Hooks Update

- [ ] Updated files to React Hooks
- [ ] Updated files to TypeScript

## Before / After

Before:
image

After:
image

## Test Plan

CI passes. Works locally.

PR #130, opened 2022-04-27
refactor: move `streams.js` into `Streams.jsx`

mvadari:

This PR gets rid of the `streams.js` file in the `src/rippled` folder, and integrates it into `Streams.jsx`, which is the only file that uses it. In the process, it removes the use of global objects, which was causing issues with tests and is bad practice.

### Context of Change

Failing a test in #129 

### Type of Change

### TypeScript/Hooks Update

## Before / After

No UI changes.

## Test Plan

CI passes. I tested the leaking tests by copying the last `LedgersPage` test and making sure both copies passed. Works locally.

## Future Tasks
This logic could definitely be simplified a bunch.


src/containers/shared/components/Streams.jsx
```diff
@@ -2,30 +2,69 @@ import { Component } from 'react';
...
   let inThrottle;
+  let queued = false;
   return function throttled(...args) {
     const context = this;
     if (!inThrottle) {
+      queued = false;
       func.apply(context, args);
       inThrottle = true;
       setTimeout(() => {
+        if (queued) {
```
JST5000, line 24: Does this check the outside state `queued`, or does this reference the value of `queued` when the timeout was started?

src/containers/shared/components/Streams.jsx
```diff
@@ -97,6 +139,74 @@ class Streams extends Component {
...
+      ledgerIndex > allValidators[pubkey].ledger_index
+    ) {
+      this.setState(prevState => {
+        const newValidatorData = Object.assign(prevState.allValidators[pubkey], {
+          ledger_hash: ledgerHash,
+          ledger_index: ledgerIndex,
+          last: Date.now(),
+        });
+        const newAllValidators = Object.assign(prevState.allValidators, {
+          [pubkey]: newValidatorData,
```
JST5000, line 195: What does `[pubkey]` syntax mean? (Is that a key that has the value of an array?)

src/containers/shared/components/Streams.jsx
ckniffen: It looks at the current value of queue.  So subsequent calls to the wrapped throttle function will set queue to true if the happen during throttling.

src/containers/shared/components/Streams.jsx
mvadari: It's syntax that essentially means the key will have the value of that variable (without the `[]` it would be a key with the value of the string `"pubkey"`)

src/containers/shared/components/Streams.jsx
```diff
@@ -2,30 +2,69 @@ import { Component } from 'react';
...
+      Log.error(error.toString());
+      if (error.code === 404 && attempts < 5) {
+        Log.info(`retry ledger ${ledger.ledger_index} (attempt:${attempts + 1})`);
+        return sleep(500).then(() => fetchLedger(ledger, rippledSocket, attempts + 1));
+      }
+      throw error;
+    });
+};
+
+const fetchLoadFee = rippledUrl => {
```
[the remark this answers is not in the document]
mvadari: That was a typo - fixed (not sure how it didn't error tbh)

ckniffen (approved):

src/containers/shared/components/Streams.jsx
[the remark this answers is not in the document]
mvadari: * shakes fist * Javascript

ckniffen (approved):

JST5000 (approved):
LGTM

Issue #71, opened 2022-01-24
Handle loading of the `/detailed` tab in transactions better


mvadari:
The transaction at https://livenet.xrpl.org/transactions/9BE43A0058FF4C358015CD89CE754AFEF201A6FF3D7E2C868B8C7E61DB00BD68 takes forever to load the `/detailed` tab, because there are over 2500 affected nodes. While this page processes all the nodes, it just shows a blank page. There should be a better loading sign.

intelliot:
@mvadari @ckniffen the page seems to load ok now. What more is there to do for this issue?

mvadari:
Yeah, I think this was resolved by #132

ckniffen:
It still loads slower than it should but I am okay closing it.

I have code somewhere that renders it in increments instead of all at once.  Much lower priority now.

intelliot:
Cool. We can still improve the speed / efficiency / progressive-loading, but since it's low priority, I'll close this issue for now. Feel free to reopen or create a new one when the time comes. Thanks!

PR #132, opened 2022-04-28
fix: NPE when an offer meta is missing PreviousFields

ckniffen:
This happens in detail view with transaction
[9BE43A0058FF4C358015CD89CE754AFEF201A6FF3D7E2C868B8C7E61DB00BD68](https://livenet.xrpl.org/transactions/9BE43A0058FF4C358015CD89CE754AFEF201A6FF3D7E2C868B8C7E61DB00BD68)

It is also the first step in fixing #71.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Test Plan

Ideally would like to find a transaction that has the same problems but is way smaller.  This specific test case will be useful for performance testing in the future as it has 5291 nodes.


JST5000 (approved):
Hahaha, that's one massive example :P

mvadari (approved):

PR #142, opened 2022-05-04
chore(deps): bump i18next libraries

ckniffen:
- i18next: ^10.6.0 => ^21.6.16
- i18next-browser-languagedetector: ^2.1.1 => ^6.1.4
- i18next-xhr-backend@^1.5.1 => i18next-http-backend@^1.4.0
- react-i18next": ^9.0.3 => ^11.16.7

- translate was changed to withTranslation
- useTranslation hook was added

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/i18n.js
```diff
@@ -1,9 +1,9 @@
 import i18n from 'i18next';
-import XHR from 'i18next-xhr-backend';
-import languageDetector from './languageDetector';
+import Backend from 'i18next-http-backend';
+import LanguageDetector from 'i18next-browser-languagedetector';
 
 const options = {
-  debug: process.env.NODE_ENV === 'development',
+  debug: true,
```
mvadari, line 6: Why was this changed?

mvadari (approved):
lgtm, assuming you checked to make sure the translations were actually working

src/i18n.js
ckniffen: Great catch.

ckniffen:
> lgtm, assuming you checked to make sure the translations were actually working

If I re-enable the picker it works great.

src/containers/Token/TokenTransactionsTable/index.js
```diff
@@ -69,7 +70,7 @@ export const TokenTxTable = props => {
               </div>
             </div>
             <div className={`col-type tx-type ${tx.type}`}>
-              <TxLabel type={tx.type} t={t} />
+              <TxLabel type={tx.type} />
```
ledhed2222, line 73: how come we don't need to pass `t` as a prop anymore? I'm not seeing it used elsewhere here

src/index.js
```diff
@@ -33,15 +33,17 @@ if (!Promise) {
...
-    <I18nextProvider i18n={i18n}>
-      <Provider store={store}>
-        <PersistGate loading={null} persistor={persistor}>
-          <Router>
-            <App />
-          </Router>
-        </PersistGate>
-      </Provider>
-    </I18nextProvider>,
+    <Suspense fallback="Loading">
```
ledhed2222, line 36: nice - just looked up `Suspense`. It has quite an ominous name, but seems very cool

src/containers/Token/TokenTransactionsTable/index.js
ckniffen: I updated TxLabel to use the `useTranslation` hook so it wont need to be passed.  I will have more PRs in the future updating more places.  That hook was added in the major version of react-i18next we are updating to.

src/index.js
ckniffen: Yeah.  I still need to research more.  Adding it prevents the page from rendering until the translations are loaded.  Otherwise a double load of the app happens

src/containers/Tokens/TokensTable/index.js
```diff
@@ -72,15 +71,19 @@ const TokensTable = props => {
           )}
         </td>
         <td className="obligations">
-          {tokenInfo.obligations ? processBigNumber(tokenInfo.obligations, lng) : 0}
+          {tokenInfo.obligations
+            ? processBigNumber(tokenInfo.obligations, i18n.resolvedLanguage)
```
JST5000, line 75: Why is `i18n` used to replace `lng` on line 52, but `i18n.resolvedLanguages` used instead of `lng` here on line 75? 
(What did `lng` represent?)

src/containers/shared/components/Sequence.tsx
```diff
@@ -31,19 +31,17 @@ const Sequence = (props: Props) => {
...
 Sequence.defaultProps = {
-  t: (d: string) => d,
   sequence: PropTypes.number,
   ticketSequence: 0,
   addContextHelp: false,
   account: '',
 };
 
-export default translate()(Sequence);
+export default Sequence;
```
JST5000, line 47: Why is Sequence no longer being translated?

ckniffen: `translate(...)` populated the `t` prop.  Now `useTranslation()` is used which makes t available via a hook.

src/containers/Tokens/TokensTable/index.js
ckniffen: `lng` represents the language code.  Good catch.

JST5000 (approved):
Once you fix that one `lng` replacement, LGTM :)

PR #145, opened 2022-05-05
refactor: filter networks with new VHS API

mvadari:

This PR filters nodes based on network using the new VHS API filters.

### Context of Change

https://github.com/ripple/validator-history-service/pull/32

NOTE: This PR should not be merged/deployed until the changes to the VHS have been deployed

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

No real change, though with the new VHS updates, more validators show up on the proper networks.

## Test Plan

CI passes. Works locally, with the VHS staging API endpoints.

### DIFF
--- a/server/routes/v1/nodes.js
+++ b/server/routes/v1/nodes.js
@@ -1,7 +1,12 @@
 const { default: axios } = require('axios');
 const log = require('../../lib/logger')({ name: 'nodes' });
 
-const DATA_URL = process.env.REACT_APP_DATA_URL;
+const ENV_NETWORK_MAP = {
+  mainnet: 'main',
+  testnet: 'test',
+  devnet: 'dev',
+  'nft-devnet': 'nft-dev',
+};
 
 const cache = {};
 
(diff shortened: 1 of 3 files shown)


server/routes/v1/nodes.js
```diff
@@ -1,7 +1,12 @@
 const { default: axios } = require('axios');
 const log = require('../../lib/logger')({ name: 'nodes' });
 
-const DATA_URL = process.env.REACT_APP_DATA_URL;
+const ENV_NETWORK_MAP = {
+  mainnet: 'main',
+  testnet: 'test',
+  devnet: 'dev',
+  'nft-devnet': 'nft-dev',
```
JST5000, line 8: Why is `nft-devnet`'s key a string when the other ones have no quotes?

server/routes/v1/validators.js
```diff
@@ -1,49 +1,29 @@
...
-      ) {
-        cache.validators = cache.validators.filter(v => v.unl === process.env.REACT_APP_VALIDATOR);
-      } else {
-        cache.validators = cache.validators.filter(
-          v =>
-            v.agreement_30day &&
-            v.agreement_30day.score &&
-            (v.unl === process.env.REACT_APP_VALIDATOR || v.unl === false)
-        );
-      }
```
JST5000, line 43: Is the filtering now happening on the api side so we don't need to manually check here? 
(In other words, why was this deleted?)

server/routes/v1/nodes.js
ckniffen: Having a dash requires quotes in JavaScript since it is an invalid variable character.

server/routes/v1/validators.js
mvadari: Yes - see https://github.com/ripple/validator-history-service/pull/32

khancode (approved):

PR #146, opened 2022-05-06
fix(nft): NFTokenAcceptOffer failure edge case and add NFT summary tests

ckniffen:

NFTokenAcceptOffer's transaction page would fail to load if the transaction failed like in the case of `tecINSUFFICIENT_FUNDS` or `tecOBJECT_NOT_FOUND`. In these situations the only data on the transaction is the offerId so that is all that will be shown.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


khancode (approved):
LGTM

PR #147, opened 2022-05-09
refactor(localizeDate): refactor(localizeDate): Add default options to clean up usage

ckniffen:

DATE_OPTIONS was defined in 8 different places. Made that the default value and removed the need to pass it in.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update


src/containers/shared/components/Tooltip.jsx
```diff
@@ -7,12 +7,6 @@ import '../css/tooltip.css';
 import { ReactComponent as PayStringToolTip } from '../images/paystring_tooltip.svg';
 
 const PADDING_Y = 20;
-const DATE_OPTIONS = {
-  hour: 'numeric',
-  minute: 'numeric',
-  second: 'numeric',
-  hour12: true,
-};
```
JST5000, line 15: This one is different - does that matter?

JST5000 (approved):
LGTM (It's so nice when you can simplify the code like this!)

src/containers/shared/transactionUtils.ts
```diff
@@ -79,17 +79,6 @@ export const CURRENCY_OPTIONS = {
   maximumFractionDigits: 8,
 };
 
-export const DATE_OPTIONS = {
-  hour: 'numeric',
-  minute: 'numeric',
-  second: 'numeric',
-  year: 'numeric',
-  month: 'long',
```
mvadari, line 87: This is different than the others - most of the other ones say `numeric` for the month

src/containers/shared/utils.js
```diff
@@ -135,11 +147,12 @@ export function formatPrice(number, lang = 'en-US', currency = 'USD', decimals =
 
 // Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat
 export const localizeDate = (date, lang = 'en-US', options = {}) => {
-  // TODO: default config
+  const config = options ?? DATE_DEFAULT_OPTIONS;
```
mvadari, line 150:
```suggestion
  const config = {
  ...DATE_DEFAULT_OPTIONS,
  options
};
```
This way, users can specify just part of the list instead of needing to repeat the default options for everything if they only want to display one part differently.

ckniffen: I originally had it that way but then found `dateStyle` didnt work if the options were extended.  I need to think through some of this some more

ckniffen:
Going to close the PR to rethink the approach.  It seems some places use different format rules.  Could be just happenstance.

PR #151, opened 2022-05-10
docs(readme): Clean up setup process

ckniffen:
- Call out node@14
- Xcode no longer required after switching to `sass` npm module
- Ruby is no longer required for sass
- CRA is not required to be installed globally to run the code
- Test run without watchman
- Filled in more blanks in `.env.example`

### Type of Change

- [x] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,9 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 
 ## Basic requirements
 
-### Install node.js and npm with [nvm](https://github.com/creationix/nvm)
+### Install Node and NPM
 
-1.  `curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash`
-1.  `nvm install node@14`
-1.  `nvm alias default node`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,23 +6,12 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 
 ## Basic requirements
 
-### xcode
-
-1.  Install xcode from app store
-1.  Open xcode and agree to license and finish the installation
-
 ### Install node.js and npm with [nvm](https://github.com/creationix/nvm)
```
shadymoses, line 9: I'd suggest simplifying this to point say something like:

> **Install Node and NPM**
> The project requires node v14 or later. Follow installation instructions on [nodejs.org](https://nodejs.org/en/). 

`nvm` is not necessary to run the project, and I think this section makes it look more complicated than it really is to get started.

ckniffen: I previously had issues running it on newer node so will say `node@14` for now.

intelliot (approved):

mvadari (approved):

PR #155, opened 2022-05-12
chore(deps): Remove react-persist and localforage

ckniffen:

We were saving all redux state to indexedDB.  None of the state was useful on reload with the exception of language code but that can be derived from i18next which stores it separately. A case could be made that retaining the transaction stats such as tx/sec and then adding to them upon revisting the site is useful.  However I would counter that that is not a valid average and that making the average be based on the ledgers received during the current visit is a more accurate representation.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update


mvadari (approved):
lgtm - but maybe should be tested in a staging environment for a few hours. Not sure if there would be long-term impacts.

JST5000 (approved):
LGTM

intelliot (approved):

PR #162, opened 2022-05-25
fix(Transaction): Force Transaction page to re-render when hash changes

ckniffen:
This also switches the Transaction component to hooks.

When the hash changed on the transaction page by either updating the url manually or searching for a transaction using the search box the page would act like it loading it new data but would keep the old transactions information.

By leveraging useEffect the new hash will be used to fetch data.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update


src/containers/Transactions/index.js
```diff
@@ -33,12 +34,15 @@ ERROR_MESSAGES.default = {
...
 const getErrorMessage = error => ERROR_MESSAGES[error] || ERROR_MESSAGES.default;
 
-class Transaction extends Component {
-  componentDidMount() {
-    const { t, actions, match, data } = this.props;
+const Transaction = props => {
+  const { identifier = '', tab = 'simple' } = useParams();
+  const match = useRouteMatch();
+  const { t } = useTranslation();
+  const { actions, data, loading } = props;
```
ledhed2222, line 41: extreme nit - usually props destructuring happens as the first line in the component. just a pattern. does not block this pr

ledhed2222:
very cool to fix this bug while moving our style forward and while removing lines of code!

ledhed2222 (approved):

src/containers/Transactions/index.js
```diff
@@ -50,10 +54,9 @@ class Transaction extends Component {
     if (identifier && identifier !== hash) {
       actions.loadTransaction(identifier, rippledSocket);
     }
-  }
+  }, [identifier]);
```
JST5000, line 57: Do we also need to include `data` in this dependency array?

ckniffen: No because this is what triggers the data request.  If we specified data it might create a loop.

JST5000 (approved):
LGTM

PR #165, opened 2022-05-31
refactor(table): Define global table styles to help future table usage

ckniffen:
## High Level Overview of Change

All three tables had their own one-off table styling.  This unifies them and will make future tables easier to implement

This eliminates 66 lines of code.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

- [ ] Updated files to React Hooks
- [ ] Updated files to TypeScript

## Before
Screen Shot 2022-05-31 at 5 21 21 PM
Screen Shot 2022-05-31 at 5 28 54 PM
Screen Shot 2022-05-31 at 5 29 02 PM

## After
Screen Shot 2022-06-01 at 2 59 20 PM
Screen Shot 2022-06-01 at 3 00 25 PM
Screen Shot 2022-06-01 at 2 59 59 PM

PR #175, opened 2022-06-07
fix(validators): Mark any valid agreement < 1 in red

ckniffen:
Any agreement score less than 1 should be marked red.

### Context of Change

I broke this in PR #165.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Screen Shot 2022-06-07 at 5 19 09 PM

Screen Shot 2022-06-07 at 5 18 47 PM

### DIFF
--- a/src/containers/Network/ValidatorsTable.jsx
+++ b/src/containers/Network/ValidatorsTable.jsx
@@ -64,7 +64,7 @@ class ValidatorsTable extends Component {
 
     return d ? (
       <td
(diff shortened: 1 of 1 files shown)


JST5000 (approved):

mvadari (approved):

intelliot (approved):
Unrelated future ideas:
1. I'd be fine with a more lenient threshold, e.g. maybe 0.9999 instead of 1 
2. I think we should consider changing this to % which seems easier to understand than a decimal number

PR #176, opened 2022-06-13
feat(assets): Add NFT and issued token tables to account page

ckniffen:

Account page was updated to use tabs.  One tab for transactions and one for issued tokens and NFTS.

### Context of Change

Allows a user to see all the issued tokens and NFTS and account owns.

Also removed eslint `import/prefer-default-export`. Read more [here](https://www.bundleapps.io/blog/use-named-exports-over-default-exports-in-javascript) about the benefits of named exports.

I wanted to use the redux hook use `useSelector` but our redux version is too old.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before
Screen Shot 2022-06-13 at 4 43 58 PM

## After
Screen Shot 2022-06-13 at 4 43 38 PM
Screen Shot 2022-06-13 at 4 43 32 PM
Screen Shot 2022-06-13 at 4 51 00 PM
Screen Shot 2022-06-13 at 4 46 45 PM


ckniffen:
> It feels weird to have tabs inside tabs. Can there instead just be three top-level tabs (transactions, fungible tokens, non-fungible tokens)?

I think at this time we are going to keep them grouped by asset type.  It does feel a bit weird but when we add more top level tabs like trustlines or offers it will reduce the future clutter.

intelliot (approved):

public/locales/en-US/translations.json
```diff
@@ -1,4 +1,9 @@
 {
+  "assets": "Assets",
+  "assets.nft_tab_title": "Non-Fungible Tokens",
+  "assets.token_tab_title": "Issued Tokens",
+  "assets.no_tokens_message": "No tokens found.",
+  "assets.no_nfts_message": "No NFTS found.",
```
JST5000, line 6: nit:
```suggestion
  "assets.no_nfts_message": "No NFTs found.",
```

public/locales/en-US/translations.json
```diff
@@ -17,6 +22,8 @@
   "xrpl_explorer": "XRPL Explorer",
   "ledgers": "Ledgers",
   "ledger": "Ledger",
+  "taxon": "Taxon",
+  "token_id": "Token Id",
```
JST5000, line 26: This should probably be called `NFTokenId` to match the rippled format

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { ChangeEvent } from 'react';
+import { useTranslation } from 'react-i18next';
+import { connect } from 'react-redux';
+import { useHistory, useParams } from 'react-router';
+import { AccountTokenTable } from '../AccountTokenTable';
+import { AccountNFTTable } from '../AccountNFTTable/AccountNFTTable';
+
+// TODO: Add state types or convert to react query
+interface Props {
+  account: any;
```
JST5000, line 10: Is there a better type we can use than `any`?

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const AccountAssetTabDisconnected = ({ account }: Props) => {
+  const assetTypes = ['token', 'nft'];
```
JST5000, line 14: Would it make sense to rename the asset type here to `issuedToken` so it doesn't get confused with a Non-fungible token when accessed later on?

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { id: accountId, assetType = assetTypes[0] } = useParams<{
+    id: string;
+    assetType: string;
+  }>();
+  const supportsNFT = process.env.REACT_APP_ENVIRONMENT === 'nft_sandbox';
```
JST5000, line 19: `devnet` also supports NFTs currently :)

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
ckniffen: That would be good.  The account is so complex I considered out of scope for the current work.  Would like to do a refactor where we at the very least use the type definitions from xrpl.js since we dont need much of the other code for interacting with stuff.

public/locales/en-US/translations.json
ckniffen: This value is used for the header and trying to keep it as plain of a name as possible.  For raw output we will use the actual field name.  This is also what we call it on the NFTokenMint transaction display.

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
ckniffen: This gets used for the url too.  I can update it be issued.

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
ckniffen: I will need to update this.

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
JST5000: I believe there's an outstanding task for us to separate out the typing so that we can reuse it in multiple projects

src/containers/Accounts/AccountAssetTab/AccountAssetTab.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const AccountAssetTabDisconnected = ({ account }: Props) => {
+  const assetTypes = ['issued', 'nft'];
+  const { id: accountId, assetType = assetTypes[0] } = useParams<{
+    id: string;
+    assetType: string;
+  }>();
+  const supportsNFT = ['nft_sandbox', 'devnet'].includes(
+    process.env.REACT_APP_ENVIRONMENT as string
+  );
```
ledhed2222, line 21: a superior way to do this would be to check what amendments are enabled by the connected server using `ledger_entry`

ckniffen: Agreed.  I think that should happen in another phase and we will at that point also add a page to show all amendments enabled on a network.  Because of the async nature of such a thing we would need to look at rewriting some stuff.

src/containers/shared/css/table.scss
```diff
@@ -53,4 +53,11 @@ table.basic {
       }
     }
   }
+
+  .empty-message {
+    padding: 16px;
+    color: #9ba2b0;
```
ledhed2222, line 59: should we add this as a constant somewhere?

PR #191, opened 2022-07-14
Add 6-digits zero padding for transaction costs

shawnxie999:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Before 
image
After
image

## Test Plan
Tested decimal numbers that include and no include decimals

### DIFF
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -129,7 +129,7 @@ export function formatPrice(number, lang = 'en-US', currency = 'USD', decimals =
     ? localizeNumber(number.toPrecision(decimals), lang, {
         style: 'currency',
         currency,
-        minimumFractionDigits: padding,
+        minimumFractionDigits: number.toPrecision(decimals).includes('.') ? padding : 0,
       })
     : undefined;
 }
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -58,7 +58,7 @@ const NUMBER_DEFAULT_OPTIONS = {
   useGrouping: true,
 };
 
-const FORMAT_PRICE_OPTIONS = {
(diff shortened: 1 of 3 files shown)


src/containers/shared/utils.js
```diff
@@ -124,11 +124,12 @@ export const localizeNumber = (num, lang = 'en-US', options = {}) => {
   return new Intl.NumberFormat(lang, config).format(number);
 };
 
-export function formatPrice(number, lang = 'en-US', currency = 'USD', decimals = 4) {
```
ledhed2222, line 127: why would we ever pad with something other than `0`? if there's no need for it now, having the additional parameter is just noise. Let's get rid of it.

ledhed2222 (changes requested):
see comment

src/containers/shared/utils.js
ckniffen: It isnt what to pad with but how many places to pad.

src/containers/shared/utils.js
```diff
@@ -58,6 +58,13 @@ const NUMBER_DEFAULT_OPTIONS = {
   useGrouping: true,
 };
 
+const FORMAT_PRICE_OPTIONS = {
+  lang: 'en-US',
+  currency: 'USD',
+  decimals: 4,
+  padding: 0,
+};
```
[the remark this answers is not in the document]
ledhed2222: the reason here is that there are not the options for format price, that would be an array:

```javascript
const FORMAT_PRICE_OPTIONS = [
  'lang',
  'currency',
  'decimals',
  'padding',
]
```

these are the defaults for the options.

ledhed2222 (changes requested):

ledhed2222 (approved):

Issue #172, opened 2022-06-01
Switch `typeface-roboto` to `@fontsource/roboto`


ckniffen:
Switch to [@fontsource/roboto](https://www.npmjs.com/package/@fontsource/roboto) which has replaced `typeface-roboto`

_Originally posted by @ckniffen in https://github.com/ripple/explorer/issues/169#issuecomment-1143956176_

jonathanlei:
fixed in #211

PR #211, opened 2022-07-28
add @fontsource/roboto and fixed compile errors

jonathanlei:

switched from typeface-roboto  to @fontsource/roboto

### Context of Change

issue: https://github.com/ripple/explorer/issues/172

### Type of Change

- [X] Refactor (non-breaking change that only restructures code)

## Before / After
changed from typeface-roboto  to @fontsource/roboto in dependency. 
fixed minor compiling issues


ckniffen (approved):

intelliot (approved):

src/containers/shared/hooks.ts
```diff
@@ -1,3 +1,4 @@
 import { useTranslation } from 'react-i18next'
 
-export const useLanguage = () => useTranslation().i18n.resolvedLanguage
+export const useLanguage = () => useTranslation().i18n.resolvedLanguage;
```
ckniffen, line 3: after @ledhed2222 commit this should not have a colon.  I don't know why it passes linting though.

src/index.js
```diff
@@ -13,7 +13,8 @@ import './containers/shared/css/global.css'
 import App from './containers/App'
 import i18n from './i18n'
 
-require('typeface-roboto')
+require('@fontsource/roboto');
```
ckniffen, line 16: same

src/containers/shared/components/TxDetails.tsx
```diff
@@ -417,7 +417,8 @@ const TxDetails = (props: Props) => {
 
   const DetailComponent = transactionTypes[type]?.TableDetail
   if (DetailComponent) {
-    return <DetailComponent instructions={instructions} />
+
+    return <DetailComponent instructions={instructions} />;
```
ckniffen, line 421: same

ckniffen (approved):

src/containers/shared/hooks.ts
ckniffen: Figured out why this doesn't fail the PR.

The PR runs `lint:js` which has `--fix`.  This causes these issues to be fixed but also means there are unstaged changes so the code does not reflect this in the repo.  One solution is to remove `--fix` the other is to do a check for unstaged changed and failed the build if those exist.  Thoughts @mvadari @ledhed2222 .

mvadari: We should add a new `lint:ci` script that doesn't call `--fix`.

ledhed2222: i agree that the linter shouldn't fix by default. i would prefer just no `--fix` and for users to call `npm run lint -- --fix` on their own

ledhed2222: i didn't want to make that change in my previous PR to keep the footprint down

src/containers/shared/components/TxDetails.tsx
```diff
@@ -417,6 +417,7 @@ const TxDetails = (props: Props) => {
 
   const DetailComponent = transactionTypes[type]?.TableDetail
   if (DetailComponent) {
+
```
ckniffen, line 420: Still an extra space

src/index.js
```diff
@@ -13,7 +13,8 @@ import './containers/shared/css/global.css'
 import App from './containers/App'
 import i18n from './i18n'
 
-require('typeface-roboto')
+require('@fontsource/roboto')
+
```
ckniffen, line 17: Still an extra new line

khancode (approved):

PR #218, opened 2022-08-02
css: update transaction simple tab table, sidechain color

mvadari:

This PR makes some CSS changes to transactions and sidechain mode.
* The simple tab table has changed a bit (see screenshots below).
* The color that represents sidechains has changed from green to yellow (since payments use green and we wanted to use the same color for the bridge transactions).

It turns out that the transaction and validator simple tab CSSes were clashing with each other, so I renamed those sections to make them separate.

### Context of Change

Requested by the designers

### Type of Change

- [x] CSS changes

### TypeScript/Hooks Update

N/A

## Before / After

Before:
image
image

After:
image
image

## Test Plan

Works locally. Tests pass.


src/containers/Transactions/simpleTab.scss
```diff
@@ -66,6 +71,8 @@
         padding-left: 5px;
         color: $white;
         float: right;
+        font-size: 14px;
```
ledhed2222, line 74: @ckniffen FYI we really should have a font style library driven by design - do you agree?

src/containers/Transactions/simpleTab.scss
```diff
@@ -1,14 +1,15 @@
...
   .rows {
     width: 100%;
     padding-right: 0px;
     padding-bottom: 40px;
+    padding-left: 16px;
     float: left;
 
     @include for-size(desktop-up) {
-      width: calc(100% - 276px - 32px);
+      width: calc(100% - 276px - 48px);
```
ledhed2222, line 12: i don't love these magic numbers. is there no way to define this relatively using percents and flex?

src/containers/Transactions/simpleTab.scss
```diff
@@ -31,20 +32,24 @@
     .row {
       display: flex;
       overflow: hidden;
-      padding: 20px 10px;
+      padding: 0px 15px;
+      padding-left: 0px;
```
ledhed2222, line 36: FYI it's considered good CSS style to not include a unit for 0 values

ledhed2222 (approved):
few comments for you but LGTM

src/containers/Transactions/simpleTab.scss
ckniffen: I disagree because it could be undoing something.  What is generally good practice is `0` vs `0px`.

src/containers/Transactions/simpleTab.scss
ckniffen: Its all over the place but yes.

src/containers/Transactions/simpleTab.scss
mvadari: There might be. I believe the idea is that you have the 276px for the sidebar with ledger index and stuff and 48px for buffer, and everything else should be taken up by this section. I'm not sure how else to put it, but I'm all ears.

jonathanlei (approved):

ckniffen (approved):

src/containers/shared/css/simpleTab.scss
```diff
@@ -10,6 +10,8 @@ $index-width: 308px;
     width: 100%;
     padding-right: 0px;
     padding-bottom: 40px;
+    padding-left: 16px;
+    float: left;
```
ckniffen, line 14: Is float needed with it being flex?

src/containers/shared/css/simpleTab.scss
```diff
@@ -36,10 +40,10 @@ $index-width: 308px;
...
       .label {
         color: $white;
-        font-size: 12px;
-        line-height: 20px;
-        text-transform: uppercase;
-        @include medium;
+        font-size: 14px;
+        line-height: 21px;
+        text-transform: capitalize;
+        @include regular;
```
ckniffen, line 46: we should not need `@regular` as it is default.

src/containers/shared/css/simpleTab.scss
```diff
@@ -21,11 +23,13 @@ $index-width: 308px;
...
       overflow: hidden;
       flex-wrap: wrap;
-      padding: 20px 10px;
+      padding: 0 15px;
+      padding-left: 0;
       border-bottom: 1px solid $gray-800;
 
       @include for-size(tablet-portrait-up) {
-        padding: 24px;
+        padding: 15px;
```
ckniffen, line 31: use shorthand notation

src/containers/shared/css/simpleTab.scss
```diff
@@ -36,10 +40,10 @@ $index-width: 308px;
 
       .label {
         color: $white;
-        font-size: 12px;
-        line-height: 20px;
-        text-transform: uppercase;
-        @include medium;
+        font-size: 14px;
```
ckniffen, line 43: Move `font-size: 14px;` and `line-height: 20px;` to `.rows` since all the elements share the same styles.

ckniffen (approved):

PR #219, opened 2022-08-02
feat: add nft_info detail page (Phase 1)

shawnxie999:

- `NFTHeader` component is for displaying all  the current NFT info(flags, transfer fee, taxon, etc)

- Currently `NFT`  component is merely a wrapper for `NFTHeader` , but it should encompass more components(ex. NFT transaction history) for the next phase
- Added `react-query v3.39.0` package for making API call without redux 
- Added `CopyToClipboard` shared component for copying content
- Added tooltip for hovering over NFT Id
- Moved the table styling from `TokenHeader` to a new file `token-table.scss` for sharing the same style used in `NFTHeader`

### Context of Change

- Navigate to a NFT detail page through routing `/token/:id`, and a valid NFT example would be 
`0000000025CC40A6A240DB42512BA22826B903A785EE2FA512C5D5A70000000C`

- Fetches `nft_info` from Clio Devnet with the following IP and port in `.env` file

 ```
REACT_APP_RIPPLED_WS_PORT=51233
REACT_APP_P2P_RIPPLED_HOST=clionft.devnet.rippletest.net
```

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After

image
image

Since there is a new redesign of tooltip, the current broken tooltip component of old style is used for now
image

### CODE REVIEW

File: src/containers/shared/css/token-table.scss [hunk 1]
```diff
@@ -0,0 +1,33 @@
+@import './variables';
+
+@mixin token-table {
+  table {
```
ckniffen, line 4: Use a class instead of a mixin.  They way you are using this causes the same class to be output twice and using `table` globally.

src/containers/App/App.js
```diff
@@ -46,6 +47,12 @@ const App = (props) => {
         `wss://${process.env.REACT_APP_P2P_RIPPLED_HOST}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
       ])
     : undefined
+  socket.clioSocket =
```
mvadari, line 50: I'd like to limit the number of new env vars we have. Instead of this, could we just make sure the p2p URL is a Clio node? I assume we want all versions of the explorer to use Clio eventually.

mvadari (commented):
@ledhed2222 @ckniffen aren't we trying to kill the use of redux in this repo? None of the new pages use it

src/containers/App/App.js
ckniffen: Does clio have the gateway call?

mvadari: Yes

ckniffen:
> @ledhed2222 @ckniffen aren't we trying to kill the use of redux in this repo? None of the new pages use it

Which of the new pages don't use it?  While that is the goal I felt it best to be consistent with the previous pages for now.

src/containers/shared/css/token-table.scss
shawnxie999: wouldnt declaring as a class also cause it to output twice? since we are still importing the file in two files?

src/containers/App/App.js
shawnxie999: @mvadari i have tried to use the cliosocket. However it doesnt work with gateway_balances. Also currently this cliosocket could only be connected through ws, not wss. Should we just it leave off for now?

src/containers/NFT/NFTHeader/index.js
```diff
@@ -1,26 +1,48 @@
-import React, { Component, useEffect, useContext, useState } from 'react'
-import { useTranslation, withTranslation } from 'react-i18next'
+import React, { useEffect, useContext, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import { Link } from 'react-router-dom'
 import PropTypes from 'prop-types'
-import { connect } from 'react-redux'
-import { bindActionCreators } from 'redux'
-import { loadNFTState } from './actions'
+import { useQuery } from 'react-query'
```
ledhed2222, line 5: this file is too big. there are really multiple components in here. it should be broken up into multiple files

src/containers/NFT/NFTHeader/index.js
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  const renderSettings = () => {
+    const { flags } = data
+
+    const burnable = flags?.includes('lsfBurnable') ? 'enabled' : 'disabled'
+    const onlyXRP = flags?.includes('lsfOnlyXRP') ? 'enabled' : 'disabled'
+    const trustLine = flags?.includes('lsfTrustLine') ? 'enabled' : 'disabled'
+    const transferable = flags?.includes('lsfTransferable')
+      ? 'enabled'
```
ledhed2222, line 122: i still don't see why you need the `?.` guard

src/containers/NFT/NFTHeader/index.js
```diff
@@ -0,0 +1,214 @@
...
+          </tr>
+        </tbody>
+      </table>
+    )
+  }
+
+  const renderHeaderContent = () => {
+    const { issuer } = data
+    const abbrvIssuer =
+      issuer?.length > 23 ? issuer?.slice(0, 23).concat('...') : issuer
```
ledhed2222, line 151: you definitely do not need the `?.` guard

ledhed2222: also the issuer length is always > 23. https://xrpl.org/accounts.html#addresses

src/containers/NFT/NFTHeader/index.js
```diff
@@ -0,0 +1,214 @@
...
+            <div
+              className="title-content"
+              onMouseOver={(e) => showTooltip(e, { tokenId })}
+              onFocus={() => {}}
+              onMouseLeave={hideTooltip}
+            >
+              {tokenId}
+            </div>
+          </div>
+        )}
```
ledhed2222, line 199: you can render all of this safely even without `data` and even if still `loading`

src/containers/NFT/NFTHeader/styles.scss
```diff
@@ -0,0 +1,201 @@
+@import '../../shared/css/variables';
+@import '../../shared/css/table';
+
+.section {
+  margin: 0 32px;
+  @include for-size(tablet-landscape-up) {
+    width: 552px;
+    margin: 0px auto;
```
ledhed2222, line 8: no `px`

src/containers/NFT/NFTHeader/index.js
ckniffen: Also truncation should happen via css.

src/containers/NFT/NFTHeader/styles.scss
```diff
@@ -0,0 +1,201 @@
...
+        color: $gray-400;
+        font-size: 14px;
+        text-transform: uppercase;
+        @include semibold;
+      }
+
+      .value {
+        display: flex;
+
+        .value-content {
```
ledhed2222, line 87: this class could be named `content` since the `value` part is inferred from the nesting

src/containers/NFT/index.js
```diff
@@ -0,0 +1,71 @@
...
+  const [error, setError] = useState(null)
+
+  document.title = `${t('xrpl_explorer')} | ${tokenId.substr(0, 12)}...`
+
+  useEffect(() => {
+    analytics(ANALYTIC_TYPES.pageview, { title: 'NFT', path: '/token/:id' })
+    return () => {
+      window.scrollTo(0, 0)
+    }
+  }, [])
```
ledhed2222, line 46: nice call to get analytics and scroll to top

src/containers/NFT/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  .loader {
+    min-height: 100px;
+    background: $black;
+    opacity: 0.8;
+    pointer-events: none;
+    transition: opacity 0.3s ease;
+  }
```
ledhed2222, line 56: shouldn't this be a general style somewhere? cc @ckniffen

src/containers/shared/css/table.scss
```diff
@@ -20,7 +20,8 @@ table.basic {
     border-bottom: 1px solid $gray-800;
   }
 
-  td, th {
+  td,
+  th {
```
ledhed2222, line 24: it always feels weird to me to target elements by type. also wouldn't this apply to all `td` and `th` since the scss is minified into a single file and this is an SPA? @ckniffen

src/containers/shared/css/table.scss
```diff
@@ -61,3 +62,33 @@ table.basic {
...
+    a {
+      color: $green;
+      text-decoration: none;
+    }
+
+    .copy {
+      position: absolute;
+      width: 17px;
+      height: 17px;
+      margin-left: 8.42px;
```
ledhed2222, line 91: is this coming from the design? seeing fractional pixels doesn't make sense to me. i don't even know what that would mean. looking around, this is going to get rounded anyway: https://stackoverflow.com/questions/9080633/can-a-css-pixel-be-a-fraction

src/rippled/lib/rippled.js
```diff
@@ -289,6 +289,20 @@ const getAccountNFTs = (rippledSocket, account, marker = '', limit = 20) =>
     limit,
   })
 
+const getNFTInfo = (rippledSocket, tokenId) =>
```
ledhed2222, line 292: this argument is not correctly named, since you cannot call this API against `rippled` but only `clio`

src/containers/App/App.js
```diff
@@ -85,42 +88,55 @@ const App = (props) => {
...
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        refetchOnWindowFocus: false,
+        refetchOnReconnect: false,
+        refetchOnMount: false,
+        retry: false,
+      },
+    },
+  })
```
ledhed2222, line 100: what is the purpose of react query here? what does it add over just calling the API directly ?

ledhed2222 (changes requested):
see comments :)

ledhed2222:
also looks like tests are failing

src/containers/App/App.js
ckniffen: This just sets up react-query for the app. @shadymoses might be able to expound on its benefits better than I.

shawnxie999:
> also looks like tests are failing

ahh its because i removed all the redux tests, so i dont have tests. its not meeting test coverage :(

src/containers/NFT/NFTHeader/index.js
ledhed2222: you're still not rendering this if the component is loading. i don't see why?

src/containers/shared/css/table.scss
shawnxie999: yeah i got 8.42 by cursoring on the figma, I guess i will  just put it a 8px then

src/rippled/lib/rippled.js
shawnxie999: it is still accepting a rippledSocket tho. it is only inside `queryP2P` where it checks for clio/p2p socket

src/containers/NFT/NFTHeader/styles.scss
shawnxie999: ahh yes, but ill probably rename it to something else, since the 'content' classname is conflicting with another style

src/containers/NFT/NFTHeader/index.js
ledhed2222: also - _why_ are we doing this truncation? it's pretty critical to know the full issuer address. i added a comment to you and natalia on figma WRT topic

src/containers/App/App.js
shadymoses: From their website:

> gives you declarative, always-up-to-date auto-managed queries and mutations that directly improve both your developer and user experiences

Makes your requests reusable hooks with a shared data cache. No more manual fetching or re-fetching, or messing with something like redux to manage server state. The explorer's current redux store is made up of 2 things:

1. some app measuring data (which isn't really being used and could be done much simpler)
2. request data that is mostly not shared across pages

All of the (necessary) app state is server state. This is not a good use case for redux. `react-query` makes handling server state significantly nicer in every regard.

ledhed2222: yeah i saw their site and i _completely_ agree that we have no need for redux in our application. my question is why we _do_ need react query, and i'm still not seeing it. i don't see why we have a need for anything special in the app storage world

shawnxie999: i think it is having to avoid the need of creating hooks to manage states like loading, error, etc

shadymoses: Well, you have state, so it has to be managed somehow. The explorer is fairly data heavy.

It's a matter of putting in the work to roll your own:
- context providers with custom hooks, caching, managing request lifecycle states, and all the other features you might need
- or copy pasta to add manual request handling everywhere in the app that needs data

Or you could use an amazingly well crafted existing solution. Savings in time and quality all around.

src/containers/shared/css/table.scss
ckniffen: Usually when it doesnt is calling out span or div.  In the case of tables it completely does especially in the context of a style meant to be used many places

ledhed2222: not following your response @ckniffen

ckniffen: Styling tables using tbody, th, td, tr, tfoot, thead are all good ways to style elements.

src/containers/NFT/styles.scss
ckniffen: I will investigate the usages.

ckniffen: @ledhed2222 Created #227.

src/containers/shared/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
...
+const CopyToClipboard = ({ text, className }: Props) => (
+  <input
+    className={className}
+    type="image"
+    src={copy}
+    alt="Copy"
+    onClick={() => navigator.clipboard.writeText(text)}
+  />
+)
+CopyToClipboard.propTypes = {
```
ckniffen, line 19: There is need to specify propTypes or defaultProps for TSX components.

src/containers/shared/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
...
+  text: PropTypes.string,
+  className: PropTypes.string,
+}
+
+CopyToClipboard.defaultProps = {
+  text: '',
+  className: null,
+}
+
+export default CopyToClipboard
```
ckniffen, line 29: Default exports are not preferred.

src/containers/shared/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import copy from '../images/copy.png'
+
+interface Props {
+  text: string
+  className?: string
+}
+
+const CopyToClipboard = ({ text, className }: Props) => (
```
ckniffen, line 10: Please add a test for this component.  Also this one could easily be its own pull request to not have it mired in this one.

src/containers/shared/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import copy from '../images/copy.png'
```
ckniffen, line 3: You should make this an svg.  It will be smaller when gzipped.  Also checkout https://jakearchibald.github.io/svgomg/.

src/containers/NFT/NFTHeader/Details.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import PropTypes from 'prop-types'
+import CopyToClipboard from '../../shared/components/CopyToClipboard'
+import './styles.css'
+
+const Details = ({ data }) => {
```
ckniffen, line 7: Please make this tsx.

src/containers/NFT/NFTHeader/Settings.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import './styles.css'
+import { useTranslation } from 'react-i18next'
+
+const Settings = ({ flags }) => {
```
ckniffen, line 6: Please make this tsx.

src/containers/NFT/NFTHeader/index.js
```diff
@@ -0,0 +1,119 @@
...
+import {
+  analytics,
+  ANALYTIC_TYPES,
+  BAD_REQUEST,
+  HASH_REGEX,
+} from '../../shared/utils'
+import Details from './Details'
+import Settings from './Settings'
+
+const NFTHeader = (props) => {
```
ckniffen, line 22: Don't use index.js for components. It makes it very hard to find components.

ckniffen (approved):

src/containers/shared/components/CopyToClipboard.tsx
shawnxie999: could you elaborate please? i thought i included them? or do you mean i dont need them?

src/containers/NFT/NFTHeader/Details.js
```diff
@@ -0,0 +1,95 @@
...
+          <tr className="row">
+            <td className="col1">Minted</td>
+            <td className="col2">{minted}</td>
+          </tr>
+        )}
+        {domain && (
+          <tr className="row">
+            <td className="col1">{t('domain')}</td>
+            <td className="col2">
+              <a
```
ckniffen, line 28: Created #228 to clean this up.

src/containers/NFT/NFTHeader/Details.js
```diff
@@ -0,0 +1,95 @@
...
+  const abbrvEmail =
+    emailHash?.length > 20 ? emailHash?.slice(0, 20).concat('...') : emailHash
+  const abbrvURI = uri?.length > 20 ? uri?.slice(0, 20).concat('...') : uri
+
+  return (
+    <table className="token-table">
+      <tbody>
+        {minted && (
+          <tr className="row">
+            <td className="col1">Minted</td>
```
ckniffen, line 20: Add i18n.

src/containers/shared/components/CopyToClipboard.tsx
shawnxie999: would we need to use something like test-library then? I dont know if we can mock something like this with wrapper.

src/containers/shared/components/CopyToClipboard.tsx
shawnxie999: huh, isnt it the convention throughout the repo to use this? or are we moving away from it?

src/containers/NFT/NFTHeader/index.js
shawnxie999: huh, isnt it the convention throughout the repo to use this? or are we moving away from it?

src/containers/shared/components/CopyToClipboard.tsx
ckniffen: Opps sorry. I meant there is no need to specify proptypes

src/containers/App/App.js
```diff
@@ -40,9 +42,10 @@ const App = (props) => {
   const hasP2PSocket =
     process.env.REACT_APP_P2P_RIPPLED_HOST != null &&
     process.env.REACT_APP_P2P_RIPPLED_HOST !== ''
+  const p2pConnection = process.env.REACT_APP_P2P_UNENCRYPTED ?? 'wss'
```
ckniffen, line 45: I dont think this is needed since we now have a clio for the nft devnet setup with wss.

shawnxie999: which clio would that be? im not sure if it supports nft_info

src/containers/NFT/NFTHeader/styles.scss
```diff
@@ -0,0 +1,201 @@
+@import '../../shared/css/variables';
+@import '../../shared/css/table';
+
+.section {
```
ckniffen, line 4: This naming feels like a global style.  I want to make sure this either gets moved into a shared location or is renamed to be more specific.

src/containers/NFT/NFTHeader/Settings.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useTranslation } from 'react-i18next'
+
+interface Props {
+  flags: string[]
+}
+
+const Settings = ({ flags }: Props) => {
+  const { t } = useTranslation()
+
+  const burnable = flags.includes('lsfBurnable') ? 'enabled' : 'disabled'
```
ckniffen, line 12: Created #233 for future cleanup.

ckniffen (approved):

src/containers/shared/css/tooltip.scss
```diff
@@ -17,6 +17,10 @@
     vertical-align: middle;
   }
 
+  .nft {
+    font-size: 13px;
```
ckniffen, line 21: Does line 10 no make it `13px`?

src/containers/NFT/NFTHeader/styles.scss
shawnxie999: this is being used by the token currency page as well. im not sure if we want to put it global when this style is specific only to these 2 places

ckniffen: It will be made global anyway so we will have two definitions of `.section` last one will win.

public/locales/en-US/translations.json
```diff
@@ -4,6 +4,20 @@
   "assets.issued_tab_title": "Issued Tokens",
   "assets.no_issued_message": "No tokens found.",
   "assets.no_nfts_message": "No NFTs found.",
+  "check_nft_id": "Please check your NFT ID",
```
mvadari, line 7: nit: this should be added to the end of the file - seems a bit weird to be here

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,102 @@
...
+          <td className="col1">{t('taxon_id')}</td>
+          <td className="col2">{NFTTaxon}</td>
+        </tr>
+        {uri && (
+          <tr className="row">
+            <td className="col1">{t('uri')}</td>
+            <td className="col2">
+              <a href={uri} target="_blank" rel="noopener noreferrer">
+                {abbrvURI}
+              </a>
```
mvadari, line 89: @ledhed2222 isn't it a security issue to have external links?

ckniffen: We currently like the domain for validators and I think account pages.  I do have a PR open that will add an external link icon to all external links.  That could be enough.

ckniffen: I can pull out that from https://github.com/ripple/explorer/pull/240.

ckniffen: It is now in a separate pr #247

ledhed2222: yes, i think we should not link out for now. doing so invites all kinds of potential security issues that we haven't thought through enough. we can always add this back

public/locales/en-US/translations.json
```diff
@@ -354,5 +354,20 @@
...
+  "uri": "URI",
+  "transfer_fee": "Transfer Fee",
+  "burnable":"Burnable",
+  "only_xrp":"Only XRP",
+  "trust_line":"Trust Line",
+  "transferable":"Transferable",
+  "buy_offers":"Buy Offers",
+  "sell_offers":"Sell Offers",
+  "offer_index": "Offer Id",
+  "owner": "Offerer's Account",
```
ledhed2222, line 371: shouldn't this be `Owner's Account`?

public/locales/en-US/translations.json
```diff
@@ -354,5 +354,20 @@
   "signature_reward": "Signature Reward",
   "min_account_create_amount": "Minimum Account Creation Amount",
   "other_chain_account": "Other Chain Account",
-  "xchain_claim_id": "XChain Claim ID"
+  "xchain_claim_id": "XChain Claim ID",
+  "check_nft_id": "Please check your NFT ID",
+  "get_nft_state_failed":"Unable to load NFT state",
```
ledhed2222, line 359: seems like this could be:

```
'get_nft_failed': 'Unable to load NFT',
```

public/locales/en-US/translations.json
```diff
@@ -354,5 +354,20 @@
...
+  "minted":"Minted",
+  "taxon_id": "Taxon ID",
+  "uri": "URI",
```
ledhed2222, line 362: I don't see why this would ever be translated. do we need this w/in locales?

public/locales/en-US/translations.json
```diff
@@ -354,5 +354,20 @@
...
+  "transfer_fee": "Transfer Fee",
+  "burnable":"Burnable",
+  "only_xrp":"Only XRP",
+  "trust_line":"Trust Line",
```
ledhed2222, line 366: need to review this copy. does this refer to the `tfTrustLine` flag? If so, this isn't clear. something like `Creates Trust Lines for Isuser`

public/locales/en-US/translations.json
```diff
@@ -354,5 +354,20 @@
...
+  "transferable":"Transferable",
+  "buy_offers":"Buy Offers",
+  "sell_offers":"Sell Offers",
+  "offer_index": "Offer Id",
```
ledhed2222, line 370: `Offer ID`

src/containers/NFT/NFT.tsx
```diff
@@ -0,0 +1,70 @@
...
+    hints: ['check_nft_id'],
+  },
+  default: {
+    title: 'generic_error',
+    hints: ['not_your_fault'],
+  },
+}
+
+const getErrorMessage = (error: any) =>
+  ERROR_MESSAGES[error] ?? ERROR_MESSAGES.default
```
ledhed2222, line 30: the way that you're using this, I don't think that the default error should be part of `ERROR_MESSAGES` because it is special. It should be its own constant.

src/containers/NFT/NFT.tsx
```diff
@@ -0,0 +1,70 @@
...
+    )
+  }
+
+  return error ? (
+    renderError()
+  ) : (
+    <div className="token-page">
+      {tokenId && <NFTHeader tokenId={tokenId} setError={setError} />}
+      {!tokenId && (
+        <div style={{ textAlign: 'center', fontSize: '14px' }}>
```
ledhed2222, line 62: no inline styles!!! use a class and a CSS file

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { CopyToClipboard } from '../../shared/components/CopyToClipboard'
+import './styles.scss'
+import { formatPrice } from '../../shared/utils'
+
+interface Props {
+  data: {
+    NFTId: string
+    ledgerIndex: number | undefined
```
ledhed2222, line 10: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties

these should all be defined using optional properties:

```typescript
data: {
  NFTId: string
  ledgerIndex?: number
  ...
}
```

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+    issuer: string
+    NFTTaxon: number
+    NFTSequence: number | undefined
+    uri: string | undefined
+    validated: boolean | undefined
+    status: string | undefined
+    warnings: string[] | undefined
+    minted: string | undefined
+    domain: string | undefined
+    emailHash: string | undefined
```
ledhed2222, line 24: really, the shape of this should be divided. this looks like the intersection of an account and an nft, so it should be:

```typescript:
interface data = {
  NFT: NFT
  Account: Account
}

interface NFT = {
  ID: string
  ...
}

interface Account = {
  ID: string
  ...
}
```

then the `NFT` and `Account` definitions should be in another file that allows them to be referenced by lots of components and just imported here

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ? `${transferFee}%`
+      : `${formatPrice(transferFee / 1000, {
+          currency: 'none',
+          decimals: 5,
+          padding: 3,
+        })}%`
+
+  const abbrvEmail =
+    emailHash &&
+    (emailHash.length > 20 ? emailHash.slice(0, 20).concat('...') : emailHash)
```
ledhed2222, line 43: this is too complex for an expression in my opinion, since you have a ternary nested within a boolean. it should be its own function. furthermore, what is the expected behavior if `emailHash` is falsey? if it is `false` or `null` then `abbrvEmail` will be `false` or `null`. that means the type of this variable could be:

`undefined | null | false | string`

that's not right.

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const abbrvURI =
+    uri && (uri.length > 20 ? uri.slice(0, 20).concat('...') : uri)
```
ledhed2222, line 45: same as above

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+        {domain && (
+          <tr className="row">
+            <td className="col1">{t('domain')}</td>
+            <td className="col2">
+              <a
+                href={`https://${domain}`}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {domain}
```
ledhed2222, line 65: i'm wary about doing this without further investigation. i think for now we should not wrap in an anchor

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+              </a>
+            </td>
+          </tr>
+        )}
+        {emailHash && (
+          <tr className="row">
+            <td className="col1">{t('email_hash')}</td>
+            <td className="col2">
+              {abbrvEmail}
+              <CopyToClipboard className="copy" text={emailHash} />
```
ledhed2222, line 75: i don't think that copying someone's email to the clipboard is useful

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+          <tr className="row">
+            <td className="col1">{t('uri')}</td>
+            <td className="col2">
+              <a
+                href={uri}
+                target="_blank"
+                rel="noopener noreferrer"
+                className="external"
+              >
+                {abbrvURI}
```
ledhed2222, line 93: we should not link to these

public/locales/en-US/translations.json
ckniffen: This key should already exist since we show it for the transactions.

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { t } = useTranslation()
+
+  const formattedFee =
+    transferFee === 0
+      ? `${transferFee}%`
+      : `${formatPrice(transferFee / 1000, {
+          currency: 'none',
+          decimals: 5,
+          padding: 3,
+        })}%`
```
ledhed2222, line 39: why do we need to special case `0`? I don't think there's anything wrong with `0.0000%`

src/containers/NFT/NFTHeader/Details.tsx
ckniffen: this should be done in css anyway basing it on character counts means it is tied for font size and other styles.

src/containers/NFT/NFTHeader/Settings.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const onlyXRP = flags.includes('lsfOnlyXRP') ? 'enabled' : 'disabled'
+  const trustLine = flags.includes('lsfTrustLine') ? 'enabled' : 'disabled'
```
ledhed2222, line 14: need to name this better potentially to describe what this flag does. see comment above.

src/containers/NFT/NFTHeader/styles.scss
ledhed2222: this doesn't sound like it was resolved. we should rename this, right?

ledhed2222 (changes requested):
this is a lot of work! nice job. there's a few things to update however

src/containers/NFT/NFTHeader/Details.tsx
mvadari: If that's the case, then we don't even need the CopyToClipboard component, which should simplify the code

src/containers/NFT/NFTHeader/Details.tsx
mvadari: Unresolving because this still has the anchor

mvadari (commented):
For future reference, each new component should have been a different PR - it would make it much easier to review. These PRs can be dependent on each other.

src/containers/NFT/NFT.tsx
```diff
@@ -0,0 +1,70 @@
...
+    title: 'assets.no_nfts_message',
+    hints: ['check_nft_id'],
+  },
+  [BAD_REQUEST]: {
+    title: 'invalid_xrpl_address',
+    hints: ['check_nft_id'],
+  },
+}
+
+const DEFAULT_ERROR: any = {
```
ledhed2222, line 25: is typescript complaining if you don't type this as `any`? use of `any` is not good. if you need to type this:

```typescript
interface ErrorMessage {
  title: string,
  hints: string[],
}

const ERROR_MESSAGES: { [string]: ErrorMessage } = ...
```

src/containers/NFT/NFTHeader/NFTHeader.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const NFTHeader = (props: Props) => {
+  const { t } = useTranslation()
+  const { tokenId, setError } = props
+  const rippledSocket = useContext(SocketContext)
+  const [tooltip, setTooltip] = useState(null)
+  const { data: rawData, isFetching: loading } = useQuery(
+    ['getNFTInfo'],
+    async () => getNFTInfo(rippledSocket, tokenId),
```
ledhed2222, line 33: why is this async? does useQuery require that?

src/containers/NFT/NFTHeader/styles.scss
ledhed2222: @ckniffen so should this be renamed?

ckniffen: We created an issue to address this later.

src/containers/NFT/NFTHeader/NFTHeader.tsx
ckniffen: I don't think it requires it.  Just that a promise is returned and getNFTInfo does that.

src/containers/shared/Interfaces.tsx
```diff
@@ -0,0 +1,36 @@
...
+  tick?: number
+  rate?: number
+  domain?: string
+  emailHash?: string
+  flags?: string[]
+  balance?: string
+  gravatar?: any
+  previousTxn?: string
+  previousLedger?: number
+}
```
ledhed2222, line 15: so these are the return values of the `rippled` APIs?

ledhed2222: if so the name is confusing - these are the required props for a component called `AccountInfo` but are the return values of the `account_info` request

ledhed2222 (approved):

PR #237, opened 2022-08-13
fix(transaction): Details page had displayed `{{code}}` for failures

### LINKED ISSUE #236: The details tabs for transactions that fail contain erroneous {{code}}
The summary for failed transactions contain unevaluated `{{code}}` text.

```
This transaction failed with a status code of {{code}}tecINSUF_RESERVE_OFFER, and validated in ledger [73492249](https://livenet.xrpl.org/ledgers/73492249) on August 5, 2022 at 4:46:30 PM UTC
```

 Example Transaction: https://livenet.xrpl.org/transactions/887F024573D3FC9990B80DAE0D3FE95D4C106C400895002436241E85A8C4C31C/detailed

Screen Shot 2022-08-12 at 6 25 34 PM

ckniffen:

The tecCode was appended instead of being incorporated into the translation.

Closes #236

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### Before
Screen Shot 2022-08-12 at 6 25 34 PM

### After
Screen Shot 2022-08-12 at 6 25 14 PM


src/containers/Transactions/DetailTab.jsx
```diff
@@ -20,21 +21,21 @@ class DetailTab extends Component {
...
     if (TransactionResult === SUCCESSFULL_TRANSACTION) {
       line1 = t('successful_transaction')
     } else {
       line1 = (
-        <>
-          {t('fail_transaction')}
-          <span className="tx-result fail">{TransactionResult}</span>
-        </>
+        <Trans i18nKey="fail_transaction" values={{ code: TransactionResult }}>
+          <span className="tx-result fail">{{ code }}</span>
```
mvadari, line 32: Why create a separate variable instead of doing this?
```suggestion
          <span className="tx-result fail">{{ '' }}</span>
```

ckniffen: I did not see that as an option in the react-i18n. I will try that out

mvadari: Oh, I see what's happening - you're passing in an object. This is confusing to read, I suggested doing this instead:
```suggestion
          <span className="tx-result fail">{ { code: '' } }</span>
```
(and same spacing for the other places where you place in a constructed object

ckniffen: I am actually not placing in an object.  It is doing interpolation when i18n-next evaluates it.  Type checking gets in the way and thinks it is ts code.

ckniffen: > Why create a separate variable instead of doing this?

`<span className="tx-result fail">{{ '' }}</span>` turns into `<span className="tx-result fail"/>` due to linter but works.

`<span className="tx-result fail">{{ code: '' }}</span>` also works but I found it more confusing.  I added an assertion to make sure the span gets populated with the code properly.

mvadari (approved):

pdp2121 (approved):

Issue #224, opened 2022-08-04
Make links more distinguishable


ckniffen:
Update links within the explorer to a green color to more easily tell when they are links.

PR #240, opened 2022-08-15
refactor(paystring): Update to use common table and link styles

ckniffen:

- Use `.basic` table styles
- Make address links green
- Remove lots of code ~160 lines
- Dynamically add external link icon to all links that arent in the footer
- Fix addresses for networks without an explorer acting like links

Related to #165 and #224

### Context of Change

The Paystring page was overlooked when working on #165 and #224.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before
Screen Shot 2022-08-15 at 6 50 12 PM

## After
Screen Shot 2022-08-15 at 6 41 01 PM


src/containers/Footer/footer.scss
```diff
@@ -46,7 +46,10 @@
     letter-spacing: 0em;
     line-height: 21px;
     text-align: left;
-    text-decoration: none;
+
+    &::after {
```
shawnxie999, line 50: What would be the purpose to this style?

ckniffen: This prevents the external link icon from displaying.  I will add a comment.

src/containers/shared/css/global.scss
```diff
@@ -152,4 +152,20 @@ a {
   &:hover {
     color: $green-primary;
   }
+
+  &[href^="http://"],
+  &[href^="https://"],
+  &.external {
```
shawnxie999, line 158: is the external class used anywhere? couldnt find it

ckniffen: I don't think it is.  I can just remove.

ckniffen:
Requires #247

shawnxie999 (approved):

JST5000 (approved):
LGTM - Only note is in the example when you click the bottom link it goes to an "Account Missing" page which I'm pretty sure is due to the account having been deleted on the XRPL, but may be worth double checking

ckniffen:
> LGTM - Only note is in the example when you click the bottom link it goes to an "Account Missing" page which I'm pretty sure is due to the account having been deleted on the XRPL, but may be worth double checking

If you access it from the testnet it will do this weirdness (happens in prod).  I think it is tied to this line https://github.com/ripple/explorer/blob/staging/src/containers/PayStrings/PayStringMappingsTable/index.js#L101

It should check if the paystring is MAINNET but the network the app is running against is testnet it needs to link to the main net url.

PR #241, opened 2022-08-16
fix(amount): Get XRP_BASE from somewhere else and remove eslint disable

ckniffen:

Error is being thrown during compilation

```
You attempted to import ../../../../server/lib/utils which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
You can either move it inside src/, or add a symlink to it from project's node_modules/.
```

### Context of Change

The app would not compile when importing XRP_BASE from server folder since it is outside of the source root aka `src`. Getting it from transactionUtils.ts fixes it.

react/require-default-props won't fail if you export a component's props

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/containers/shared/components/Amount.tsx
+++ b/src/containers/shared/components/Amount.tsx
@@ -1,11 +1,10 @@
 import React from 'react'
 import { useTranslation } from 'react-i18next'
-import { XRP_BASE } from '../../../../server/lib/utils'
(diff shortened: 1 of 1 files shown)


mvadari (approved):
Whoops

shawnxie999 (approved):

PR #252, opened 2022-08-18
feat: add support for `XChainAccountCreate` tx

mvadari:

This PR adds support for the `XChainAccountCreate` transaction type. It also handles some renames.

### Context of Change

cross-chain bridge tx support

### Type of Change

### TypeScript/Hooks Update

New files use TypeScript and Hooks.

## Test Plan

Added a basic test and a test for a failed tx.


mvadari:
Waiting on opening this PR because there are going to be some renames

src/containers/shared/components/Transaction/XChainAccountCreateCommit/parser.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function parser(tx: any, meta: any) {
+  const affectedNodes = meta.AffectedNodes
+  const bridgeMeta = affectedNodes.filter(
+    (node: any) =>
+      node.ModifiedNode && node.ModifiedNode.LedgerEntryType === 'Bridge',
+  )[0]
+  // TODO: somehow get the bridge owner via ledger_entry
+  // AffectedNodes won't contain the bridge owner if the transaction fails
+  return {
```
ckniffen, line 11: I was thinking it would be nice to make the parser async so that additional info could be looked up.

src/containers/shared/components/Transaction/XChainAccountCreateCommit/Simple.tsx
```diff
@@ -0,0 +1,44 @@
...
+      },
+    },
+  } = props
+
+  return (
+    <>
+      <SimpleRow label={t('send')} data-test="send">
+        <Amount value={amount} />
+      </SimpleRow>
+      <XChainBridge
```
ckniffen, line 32: Out of scope but would it be easier to keep these as one value that is passed in.  They seem to always be passed in together.

ckniffen (approved):

src/containers/shared/components/Transaction/XChainAccountCreateCommit/Simple.tsx
mvadari: Oh huh I didn't even think about that - yes, I could do that, but I'll save it for another PR.

jonathanlei (approved):

Issue #108, opened 2022-03-28
Unable to see full URI for long NFT URI's


JST5000:
Quoting @andrew-keig-nomena who posted this issue on March 28th in the xrpl.js repo here: https://github.com/XRPLF/xrpl.js/issues/1944

When viewing my transaction on [nft-devnet.xrpl.org](http://nft-devnet.xrpl.org/), I am not able to see the complete IPFS link. Also would it be possible to provide a link to IPFS from here.

Example below:

https://nft-devnet.xrpl.org/transactions/DB45197C3165450FC3A1035C5517E09D0724DC9CF461E18899D2D883AC6FD227

mvadari:
The link is purposefully not provided, because that could be a potential security issue

JST5000:
Not hyperlinking is fine, but this is about how the text just runs off the screen without any way to copy + paste it. 

If our goal is to not show the link at all, then we should have a better user interface to demonstrate that. (Ex. "URI Hidden for security") 

(Although I think that having a plaintext version of the link be fully visible would be better than hiding the URI entirely)

ledhed2222:
Added to our board

ckniffen:
fixed by #261

PR #261, opened 2022-08-23
fix(transactions): Decrease URI font on NFTTokenMint

shawnxie999:
URI link overflows, changed its tags to prevent it from doing so

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Before
image

After
image

### DIFF
--- a/src/containers/Transactions/Simple/NFTokenMint.jsx
+++ b/src/containers/Transactions/Simple/NFTokenMint.jsx
@@ -26,8 +26,10 @@ const NFTokenMint = (props) => {
       </div>
       <div className="row">
         <div className="label">URI</div>
-        <div className="value" data-test="token-uri">
-          {uri}
+        <div className="value">
+          <div className="dt" data-test="token-uri">
+            {uri}
+          </div>
         </div>
       </div>
     </>


src/containers/Transactions/Simple/NFTokenMint.jsx
```diff
@@ -26,8 +26,10 @@ const NFTokenMint = (props) => {
       </div>
       <div className="row">
```
ckniffen, line 27: if you add `flex-wrap` here it will also make it so even more of the url can be shown in case it is even longer.

I have a PR coming up that will remove the need to add `flex-wrap` and will apply it to all simple tab rows.  You can choose to wait for that PR if you want.

ckniffen (approved):

ckniffen:
This will fix #108

src/containers/Transactions/Simple/NFTokenMint.jsx
shawnxie999: oh just saw that PR. I can close this one, doesnt seem necessary

ledhed2222 (approved):

PR #270, opened 2022-08-25
feat(NFT): Add NFT Transaction, buy and sell offers (Phase 2)

shawnxie999:
- added get methods for `nft_tx`, `nft_buy_offers`, and `nft_sell_offers`
- created tab switching between Transactions, Buy Offers, and Sell Offers
- created a shared component for the Transactions Table to display NFT transactions
- created `Offers` component to display both types of offers
- transformed all default exports to named exports within NFT

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Transactions
image

image

Buy/Sell Offers
image

image

image


src/rippled/lib/rippled.js
```diff
@@ -303,6 +303,74 @@ const getNFTInfo = (rippledSocket, tokenId) =>
...
+    if (resp.error === 'objectNotFound') {
+      throw new Error('NFT not found', 404)
+    }
+    if (resp.error_message) {
+      throw new Error(resp.error_message, 500)
+    }
+    return resp
+  })
+
+const getSellNFToffers = (rippledSocket, tokenId, limit = 50, marker = '') =>
```
ckniffen, line 322: These two should be shortcuts for one method that just takes the command since otherwise they are identical

src/containers/shared/components/TransactionTable/TransactionTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+        ) : (
+          transactions.map(renderListItem)
+        )}
+      </ol>
+      {loading ? <Loader /> : renderLoadMore()}
+    </div>
+  )
+}
+
+TransactionTable.defaultProps = {
```
ckniffen, line 66: Default Props is not needed for tsx.  export the props type and the warnings will go away.

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <SuccessIcon className="successful" title={t('success')} />
+              ) : (
+                <FailIcon className="failed" title={t('fail')} />
+              )}
+              <span className="status">{status}</span>
+            </span>
+          </div>
+          <div className="col-date">{date}</div>
+        </div>
+        {detailsEnabled && (
```
ckniffen, line 67: Having a prop is the wrong approach.  TxDetails will be null if there are no details. Instead do 

```
const details =  <TxDetails
    language={i18n.resolvedLanguage}
    type={tx.type}
     instructions={tx.details.instructions}
/>
```

And then later do

```
{details && (
  <div className="details">{details}</div>
)}
```

shawnxie999: This was my first approach, but it seems that nft_tx has details too...

shawnxie999: Nevermind, I realized its something with the `summarizeTransaction`, and there is a flag to not parse details. I can just turn it off and it will work I think. 

It will lose some of the details about the tx(like  the amount of xrp for tx), but  these arent being displayed by the table anyways

shawnxie999: Ill create a seperate issue for this 👍

src/containers/NFT/NFTTransactions/NFTTransactions.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Tabs from '../../shared/components/Tabs'
+import { getBuyNFToffers, getSellNFToffers } from '../../../rippled/lib/rippled'
+import Offers from './Offers'
+import Transactions from './Transactions'
+
+interface Props {
+  tokenId: string
+}
+
+const NFTTransactions = (props: Props) => {
```
ckniffen, line 13: This should not be named NFTransactions it contains more than just transactions.  Maybe NFTTabs

src/containers/NFT/NFTTransactions/NFTTransactions.tsx
```diff
@@ -0,0 +1,66 @@
...
+        )
+        break
+      default:
+        body = <Transactions tokenId={tokenId} />
+        break
+    }
+    return <div>{body}</div>
+  }
+  return (
+    <div className="nft-transactions">
```
ckniffen, line 59: Same as above in relation to this class name.

src/containers/NFT/NFTTransactions/NFTTransactions.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <div>{renderTabs()}</div>
+      <div className="tab-body">{renderTransactions()}</div>
+    </div>
+  )
+}
+
+export default NFTTransactions
```
ckniffen, line 66: Avoid using default exports.  It makes exploring code very difficult.

src/containers/NFT/NFTTransactions/styles.scss
```diff
@@ -0,0 +1,119 @@
...
+    @include for-size(tablet-landscape-up) {
+      width: 552px;
+      margin: 0px auto;
+    }
+
+    @include for-size(desktop-up) {
+      width: 1140px;
+    }
+
+    .react-json-view {
```
ckniffen, line 34: Do we show json on the nft page?

src/containers/shared/components/TransactionTable/styles.scss
```diff
@@ -0,0 +1,277 @@
+@import '../../../shared/css/variables';
+
+.account-transactions {
```
ckniffen, line 3: This class should not be named this.  It should be a more general name like .transaction-table especially since these styles are used on the nft.

src/containers/NFT/NFTTransactions/NFTTransactions.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const { id = '', tab = 'transactions' } = useParams<{
+    id: string
+    tab: string
+  }>()
+  const match = useRouteMatch()
+  const { tokenId } = props
+
+  function renderTabs() {
+    const { path = '/' } = match
+    const tabs = ['transactions', 'buy_offers', 'sell_offers']
```
ckniffen, line 23: use dashes instead of underscores since these are used in the url's path.

src/containers/NFT/NFTTransactions/styles.scss
shawnxie999: forgot to delete👍

src/containers/shared/components/TransactionTable/styles.scss
shawnxie999: true, will change

public/locales/en-US/translations.json
```diff
@@ -364,8 +364,8 @@
   "only_xrp":"Only XRP",
   "trust_line":"Creates Trust Lines for Issuer",
   "transferable":"Transferable",
-  "buy_offers":"Buy Offers",
-  "sell_offers":"Sell Offers",
+  "buy-offers":"Buy Offers",
+  "sell-offers":"Sell Offers",
```
ledhed2222, line 368: why did you change to a dash?

src/containers/App/App.js
```diff
@@ -125,7 +125,7 @@ const App = (props) => {
                 />
                 <Route exact path="/paystrings/:id?" component={paystrings} />
                 <Route exact path="/token/:currency.:id" component={token} />
-                <Route exact path="/token/:id" component={NFT} />
+                <Route exact path="/token/:id/:tab?" component={NFT} />
```
ledhed2222, line 128: that's clever

src/containers/NFT/NFTTabs/NFTTabs.tsx
```diff
@@ -0,0 +1,64 @@
...
+            fetchOffers={getBuyNFToffers}
+            offerType="BuyOffers"
+          />
+        )
+        break
+      default:
+        body = <Transactions tokenId={tokenId} />
+        break
+    }
+    return <div>{body}</div>
```
ledhed2222, line 56: not a big deal, but in this case I think it would be easier to have the function just `return` what you want instead of setting the back of `body`.

src/containers/NFT/NFTTabs/Offers.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Link } from 'react-router-dom'
+import Loader from '../../shared/components/Loader'
+import './styles.scss'
+import noInfo from '../../shared/images/no_info.svg'
+import SocketContext from '../../shared/SocketContext'
+import { Amount } from '../../shared/components/Amount'
+import { analytics, ANALYTIC_TYPES } from '../../shared/utils'
+import '../../shared/components/TransactionTable/styles.scss' // Reuse load-more-btn
+
+const formatAmount = require('../../../rippled/lib/txSummary/formatAmount')
```
ledhed2222, line 13: why do we `require` instead of `import`?

src/containers/NFT/NFTTabs/Offers.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface Props {
+  tokenId: string
+  offerType: string
+  fetchOffers: (
+    socket: any,
+    id: string,
+    limit: number | undefined,
```
ledhed2222, line 21: you should write this as `limit?: number`

public/locales/en-US/translations.json
shawnxie999: Because we are using the same string for locale and the url inside Tabs component. And I dont think we can have underscore in url by convention

`const Tabs = (props: Props) => {
  const { t } = useTranslation()
  const { tabs, selected, path } = props
  const items = tabs.map((title) => {
    const className = selected === title ? 'tab selected' : 'tab'
    return (
      <Link
        className={className}
        title={t(title)}
        key={title}
        to={`${path}/${title}`}
      >
        {t(title)}
      </Link>
    )
  })
  return <div className="tabs">{items}</div>
}`

src/containers/NFT/NFTTabs/Offers.tsx
shawnxie999: Fixed.

src/containers/NFT/NFTTabs/Offers.tsx
shawnxie999: I cant do that. I pass in either `getBuyNFToffers` or `getBuyNFToffers` into `fetchOffers`. Both of these two functions are in a javascript file, and js files dont support optional param for function

ckniffen: They should support default values

shawnxie999: Yes  we need to have undefined type in order to use default param

```
const getBuyNFToffers = (rippledSocket, tokenId, limit = 50, marker = '') =>
  getNFToffers('nft_buy_offers', rippledSocket, tokenId, limit, marker)

```

public/locales/en-US/translations.json
ckniffen: you could do `t(title.replaceAll('-', '_'))`

src/containers/NFT/NFTHeader/NFTHeader.tsx
```diff
@@ -48,7 +64,40 @@ const NFTHeader = (props: Props) => {
     }
   }, [tokenId])
 
-  const data = rawData && formatNFTInfo(rawData)
+  const data = useMemo(() => rawData && formatNFTInfo(rawData), [rawData])
```
ckniffen, line 67: Why is this needed?

src/containers/NFT/NFTTabs/NFTTabs.tsx
ckniffen: Yeah and the the wrapping element be where this function is called.

src/containers/NFT/NFTHeader/NFTHeader.tsx
shawnxie999: I added it to avoid re-computing the format function, because this component has several api responses that will re-render. I can remove it I dont think it has any noticeable difference

public/locales/en-US/translations.json
shawnxie999: i dont think replaceAll is supported in some browser versions lol. And also my tests is throwing error about not recognzing replaceAll 🤣. ill find an alternative

ckniffen (approved):

src/containers/NFT/NFTTabs/test/NFTTabs.test.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import { I18nextProvider } from 'react-i18next'
+import { MemoryRouter as Router, Route } from 'react-router-dom'
+import { QueryClient, QueryClientProvider } from 'react-query'
+import { NFTTabs } from '../NFTTabs'
+import i18n from '../../../../i18nTestConfig'
+
+describe('NFT Transactions tab container', () => {
+  const queryClient = new QueryClient({
```
ckniffen, line 10: The defining of the client should be pulled out into its own file since it is the same as in App.js and in every test that has react-query in it.

ckniffen (approved):

src/containers/NFT/NFTHeader/Details.tsx
```diff
@@ -13,12 +13,12 @@ interface Props {
...
 }
 
-const Details = ({ data }: Props) => {
+export const Details = ({ data }: Props) => {
   const { minted, domain, NFTTaxon, uri, transferFee } = data
   const { t } = useTranslation()
   const language = useLanguage()
   const formattedFee =
-    transferFee &&
+    transferFee !== undefined &&
```
ledhed2222, line 21: i prefer `transferFee != null` but nbd

src/containers/NFT/NFTHeader/NFTHeader.tsx
```diff
@@ -48,7 +68,33 @@ const NFTHeader = (props: Props) => {
...
+    ['getAccountInfo'],
+    async () => {
+      const info = await getAccountInfo(rippledSocket, data?.issuer)
+      return formatAccountInfo(info, {})
+    },
+    { enabled: !!data },
+  )
+
+  const mintedDate =
+    firstTransaction && firstTransaction.transaction?.type === 'NFTokenMint'
```
ledhed2222, line 91: i don't see why there would ever be a `firstTransaction` but that transaction would not have a type. If that's true, you should write _either_ :

```js
const mintedDate = firstTransaction && firstTransaction.transaction.type === 'NFTokenMint' ....
```

or:

```js
firstTransaction?.transaction?.type === 'NFTokenMint' ...
```

personally I prefer the latter.

src/containers/NFT/NFTHeader/NFTHeader.tsx
```diff
@@ -59,27 +105,33 @@ const NFTHeader = (props: Props) => {
...
+                ...data,
+                domain: accountData?.domain,
+                minted: mintedDate,
+              }}
+            />
           </div>
           <div className="settings">
             <div className="title">{t('settings')}</div>
-            <Settings flags={data.flags} />
+            <Settings flags={data?.flags ?? []} />
```
ledhed2222, line 134: isn't the `data!` syntax to let ts know that `data` is never null-ish? If that's true, then why would you need the `data?.flags` syntax instead of `data.flags`?

src/containers/NFT/NFTTabs/NFTTabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+    tab: string
+  }>()
+  const match = useRouteMatch()
+  const { tokenId } = props
+
+  function renderTabs() {
+    const { path = '/' } = match
+    const tabs = ['transactions', 'buy-offers', 'sell-offers']
+    // strips :url from the front and the identifier/tab info from the end
+    const mainPath = `${path.split('/:')[0]}/${id}`
```
ledhed2222, line 25: even nicer would be:

```js
const mainPath = [path.split('/:')[0], id].join('/')
```

src/containers/shared/QueryClient.tsx
```diff
@@ -0,0 +1,12 @@
...
+export const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+      refetchOnReconnect: false,
+      refetchOnMount: false,
+      retry: false,
+    },
+  },
+})
```
ledhed2222, line 12: nice

src/containers/shared/components/TransactionTable/TransactionTable.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { MouseEventHandler } from 'react'
+import { useTranslation } from 'react-i18next'
+import { TransactionTableRow } from './TransactionTableRow'
+import Loader from '../Loader'
+import { LoadMoreButton } from '../../LoadMoreButton'
+import './styles.scss'
+
+export type TransactionTableProps = React.HTMLAttributes<HTMLElement> & {
+  transactions: any[] | undefined
```
ledhed2222, line 9: again we can't use `transactions?: any[]`?

src/containers/shared/components/TransactionTable/TransactionTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+  emptyMessage?: string
+  loading: boolean
+  onLoadMore: MouseEventHandler
+  hasAdditionalResults: boolean | undefined
```
ledhed2222, line 13: ditto

ledhed2222: also this is weird - seems like the `undefined` case should coalesce to `false` somewhere

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Link } from 'react-router-dom'
+import TxLabel from '../TxLabel'
+import { ReactComponent as SuccessIcon } from '../../images/success.svg'
+import { ReactComponent as FailIcon } from '../../images/ic_fail.svg'
+import TxDetails from '../TxDetails'
+import { localizeDate } from '../../utils'
+import './styles.scss'
+import { useLanguage } from '../../hooks'
+
+const TIME_ZONE = 'UTC'
```
ledhed2222, line 12: should we have a global timezone setting maybe?

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -0,0 +1,78 @@
...
+const DATE_OPTIONS = {
+  hour: 'numeric',
+  minute: 'numeric',
+  second: 'numeric',
+  year: 'numeric',
+  month: 'numeric',
+  day: 'numeric',
+  hour12: true,
+  timeZone: TIME_ZONE,
+}
```
ledhed2222, line 22: and DT options?

src/containers/shared/components/TransactionTable/styles.scss
```diff
@@ -0,0 +1,281 @@
...
+    font-size: 14px;
+  }
+
+  .col-type,
+  .col-account,
+  .col-status,
+  .col-date {
+    display: inline-block;
+    overflow: hidden;
+    // padding: 4px 8px 4px 4px;
```
ledhed2222, line 37: looks like you meant to remove this

src/rippled/NFTTransactions.js
```diff
@@ -0,0 +1,43 @@
...
+      return {
+        transactions,
+        marker: data.marker,
+      }
+    })
+    .then((d) => d)
+    .catch((error) => {
+      log.error(error.toString())
+      throw error
+    })
```
ledhed2222, line 24: do you prefer this syntax to just `await`? i don't

ledhed2222 (changes requested):
see comments

src/containers/NFT/NFTTabs/styles.scss
```diff
@@ -0,0 +1,75 @@
...
+.nft-tabs {
+  margin-top: 60px;
+}
+
+.offers-table {
+  position: relative;
+  min-height: 150px;
+
+  table {
+    .no-info-panel {
```
ckniffen, line 13: Is this supposed to have different styling than the no results display used on the assets table?

shawnxie999: The “no information” for offers table is unique afaik

src/containers/shared/components/TransactionTable/TransactionTable.tsx
ckniffen: Having it optional is fine and default to false

src/containers/NFT/NFTHeader/Details.tsx
shawnxie999: reverted back to the original, explicitly check for null or undefined isnt needed here

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
shawnxie999: Whats DT?

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
shawnxie999: i think there is one inside `transacactionUtils.ts` that renders date in non-numeric format (eg, "December 20, 2020 at 3:23:16 AM").  But I also see many modules define their own date-options for numeric dates (eg, "12/20/2020, 3:23:16 AM"), which indeed have duplicated code. So I think it should be part of another issue if we want to clean up and rename the date options. @ckniffen correct me if im wrong

src/rippled/NFTTransactions.js
shawnxie999: are you talking about then/catch? most other files use this instead of await...

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
ckniffen: Let make it part of something else

src/rippled/NFTTransactions.js
```diff
@@ -0,0 +1,42 @@
...
+        const txn = formatTransaction(tx)
+        return summarize(txn, true)
+      })
+
+      return {
+        transactions,
+        marker: data.marker,
+      }
+    })
+    .then((d) => d)
```
ckniffen, line 20: This `then` is doing nothing.

src/rippled/NFTTransactions.js
```diff
@@ -0,0 +1,42 @@
...
+    .then((data) => {
+      const transactions = data.transactions.map((tx) => {
+        const txn = formatTransaction(tx)
+        return summarize(txn, true)
+      })
+      return {
+        transaction: transactions?.length > 0 ? transactions[0] : undefined,
+      }
+    })
+    .then((d) => d)
```
ckniffen, line 38: This `then` is doing nothing.

src/rippled/NFTTransactions.js
ledhed2222: doesn't matter - you can use whichever you want. with `await` you have the benefit of  traditional `try` and `catch`

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
ledhed2222: datetime

ledhed2222 (approved):

Issue #258, opened 2022-08-23
Add ability to connect to insecure rippled and clio servers


ckniffen:
In an effort to make local development easier there should be a simple way to connect to insecure/local servers instead of directly editing code.

Right now `wss` is the implied protocol that must be used when defining [REACT_APP_RIPPLED_HOST](https://github.com/ripple/explorer/blob/staging/.env.example#L5) and [REACT_APP_P2P_RIPPLED_HOST](https://github.com/ripple/explorer/blob/staging/.env.example#L7). These are used [here](https://github.com/ripple/explorer/blob/staging/src/containers/App/App.js#L29-L47).  

One option is to make the host environment variable values incorporate the protocol.  The other is to have a new environment variable that is used to swap between using `wss/https` and `ws/http`.

rashidmya:
can I try doing this?

ckniffen:
> can I try doing this?

Sure thing.  Fork the repo and when you are done create a pull request.

ckniffen:
@rashidmya also feel free to ask any questions in the ticket.

rashidmya:
> One option is to make the host environment variable values incorporate the protocol. 
Is this what you meant? #266

If not, maybe i can just add a new environment variable to swap between secure and insecure connections

ckniffen:
REACT_APP_ENVIRONMENT is used to distinguish the various networks such as

- https://livenet.xrpl.org/
- https://testnet.xrpl.org/
- https://devnet.xrpl.org/
- https://nft-devnet.xrpl.org/
- https://sidechain.xrpl.org/

This task is for a scenario like "As a user I want to connect to a testnet rippled or clio node that does not have ssl enabled" Usually this is done to hit a local server or one on another machine in the same private network. In that scenario the `REACT_APP_ENVIRONMENT === 'testnet'` and I want it to use `ws`.  However in production for https://testnet.xrpl.org/ `REACT_APP_ENVIRONMENT === 'testnet'` will also be true but we want to use `wss`.  This makes `REACT_APP_ENVIRONMENT` not a good variable to base this off of.

Additionally for local development I typically point to a production rippled node so relying on `NODE_ENV` is not feasible either.

rashidmya:
> Additionally for local development I typically point to a production rippled node so relying on `NODE_ENV` is not feasible either.

so an environment variable to switch between secure and insecure connections would be a solutions to this i guess

ckniffen:
[...]

PR #272, opened 2022-08-26
added new env variable to switch between ws & wss

rashidmya:
- added a new environment variable `REACT_APP_RIPPLED_SECURE` to switch between `ws/http` and `wss/https`
- if the env variable is not defined or not set to true, the default value would be false which means the connection is not secure
 

### Context of Change
#258 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
code added:
```js
  const isSecureConnection = process.env.REACT_APP_RIPPLED_SECURE === 'true' || false
  const wsConnectionProtocol = isSecureConnection ? 'wss' : 'ws'
  const wsConnectionPort = isSecureConnection ? '443' : '80'
```

- before
```js
  if (rippledHost.includes(':')) {
    wsUrls.push(`wss://${rippledHost}`)
  } else {
    wsUrls.push.apply(wsUrls, [
      `wss://${rippledHost}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
      `wss://${rippledHost}:443`,
    ])
  }
```
```js
  socket.p2pSocket = hasP2PSocket
    ? new XrplClient([
        `wss://${process.env.REACT_APP_P2P_RIPPLED_HOST}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
      ])
    : undefined
```

- after
```js
  if (rippledHost.includes(':')) {
    wsUrls.push(`${wsConnectionProtocol}://${rippledHost}`)
  } else {
    wsUrls.push.apply(wsUrls, [
      `${wsConnectionProtocol}://${rippledHost}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
      `${wsConnectionProtocol}://${rippledHost}:${wsConnectionPort}`,
    ])
  }
```
```js
  socket.p2pSocket = hasP2PSocket
    ? new XrplClient([
        `${wsConnectionProtocol}://${process.env.REACT_APP_P2P_RIPPLED_HOST}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
      ])
    : undefined
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
 NODE_ENV=production
 REACT_APP_RIPPLED_WS_PORT=51233
 REACT_APP_RIPPLED_PEER_PORT=51235
+REACT_APP_RIPPLED_SECURE=true
 # Your Rippled Primary Host
 REACT_APP_RIPPLED_HOST=s2.ripple.com
 # Your Rippled Host That Can Run Gateway Balances URL
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -25,17 +25,21 @@ const MODE = process.env.REACT_APP_ENVIRONMENT
 const App = (props) => {
   const { actions, location, match } = props
 
+  const isSecureConnection = process.env.REACT_APP_RIPPLED_SECURE === 'true' || false
+  const wsConnectionProtocol = isSecureConnection ? 'wss' : 'ws'
+  const wsConnectionPort = isSecureConnection ? '443' : '80'
+
   const {
     params: { rippledUrl = null },
   } = match
   const rippledHost = rippledUrl ?? process.env.REACT_APP_RIPPLED_HOST
   const wsUrls = []
   if (rippledHost.includes(':')) {
-    wsUrls.push(`wss://${rippledHost}`)
+    wsUrls.push(`${wsConnectionProtocol}://${rippledHost}`)
   } else {
     wsUrls.push.apply(wsUrls, [
-      `wss://${rippledHost}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
-      `wss://${rippledHost}:443`,
+      `${wsConnectionProtocol}://${rippledHost}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
+      `${wsConnectionProtocol}://${rippledHost}:${wsConnectionPort}`,
     ])
   }
   const socket = new XrplClient(wsUrls)
@@ -44,7 +48,7 @@ const App = (props) => {
     process.env.REACT_APP_P2P_RIPPLED_HOST !== ''
   socket.p2pSocket = hasP2PSocket
     ? new XrplClient([
-        `wss://${process.env.REACT_APP_P2P_RIPPLED_HOST}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
+        `${wsConnectionProtocol}://${process.env.REACT_APP_P2P_RIPPLED_HOST}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
       ])
     : undefined
 


mvadari:
I don't like the idea of having a new env variable for this - it clutters up the env vars, and we already have a lot.

JST5000:
> I don't like the idea of having a new env variable for this - it clutters up the env vars, and we already have a lot.

To avoid that, It might be better to make the `REACT_APP_RIPPLED` and `REACT_APP_P2P_RIPPLED_HOST` include the `ws`/`wss` parts of their url instead of just the middle part. (Ex. `s.altnet.rippletest.net` -> `wss://s.altnet.rippletest.net`)

mvadari:
> To avoid that, It might be better to make the `REACT_APP_RIPPLED` and `REACT_APP_P2P_RIPPLED_HOST` include the `ws`/`wss` parts of their url instead of just the middle part. (Ex. `s.altnet.rippletest.net` -> `wss://s.altnet.rippletest.net`)

That makes URLs really un-user-friendly for sidechains, since the `/` needs to be URL-encoded.

ckniffen:
> > To avoid that, It might be better to make the `REACT_APP_RIPPLED` and `REACT_APP_P2P_RIPPLED_HOST` include the `ws`/`wss` parts of their url instead of just the middle part. (Ex. `s.altnet.rippletest.net` -> `wss://s.altnet.rippletest.net`)
> 
> That makes URLs really un-user-friendly for sidechains, since the `/` needs to be URL-encoded.

That could be the exception to the rule.  Sidechains already have support for optionally supplying the whole protocol

rashidmya:
> > I don't like the idea of having a new env variable for this - it clutters up the env vars, and we already have a lot.
> 
> To avoid that, It might be better to make the `REACT_APP_RIPPLED` and `REACT_APP_P2P_RIPPLED_HOST` include the `ws`/`wss` parts of their url instead of just the middle part. (Ex. `s.altnet.rippletest.net` -> `wss://s.altnet.rippletest.net`)

should I implement this? @ckniffen

mvadari:
> > > To avoid that, It might be better to make the `REACT_APP_RIPPLED` and `REACT_APP_P2P_RIPPLED_HOST` include the `ws`/`wss` parts of their url instead of just the middle part. (Ex. `s.altnet.rippletest.net` -> `wss://s.altnet.rippletest.net`)
> > 
> > 
> > That makes URLs really un-user-friendly for sidechains, since the `/` needs to be URL-encoded.
> 
> That could be the exception to the rule. Sidechains already have support for optionally supplying the whole protocol

That feels like a worse UX in prod, because in prod you can't use any `ws` connection due to CORS.

intelliot:
Personally I think it's fine to include the protocol in the ENV variables, as ckniffen suggested. @mvadari thoughts?

mvadari:
> Personally I think it's fine to include the protocol in the ENV variables, as ckniffen suggested. @mvadari thoughts?

I guess it's fine, yeah. I was hoping that we could have some sort of elegant solution where the Explorer would automatically detect if you were running locally instead of in production, and change the protocol automatically.

ckniffen:
> I guess it's fine, yeah. I was hoping that we could have some sort of elegant solution where the Explorer would automatically detect if you were running locally instead of in production, and change the protocol automatically.

Having the full protocol will also help when pointing to a server we haven't set up SSL on yet that isn't local too. That was the case for some servers that were set up for clio nft calls for weeks.

mvadari:
> Having the full protocol will also help when pointing to a server we haven't set up SSL on yet that isn't local too. That was the case for some servers that were set up for clio nft calls for weeks.

We can't connect to servers that don't have SSL in prod, due to CORS, since the production website is deployed on HTTPS.

ckniffen:
> We can't connect to servers that don't have SSL in prod, due to CORS, since the production website is deployed on HTTPS.

This was doing local explorer development and pointing to a ip address for the p2p socket but using a production server over ssl for the main socket.  Normally while doing local development I am pointing to both to a production server.

mvadari:
> > We can't connect to servers that don't have SSL in prod, due to CORS, since the production website is deployed on HTTPS.
> 
> This was doing local explorer development and pointing to a ip address for the p2p socket but using a production server over ssl for the main socket. Normally while doing local development I am pointing to both to a production server.

Sorry, I meant that I wish we could detect if the _Explorer_ is running locally, and switch to `ws` in that case.

mvadari:
Closed in favor of #394

mvadari:
Closed in favor of https://github.com/ripple/explorer/pull/394

PR #276, opened 2022-08-29
refactor: domainLink component

connorjchen:

Created DomainLink component to reduce repeated code for domain links in 5 areas of the repo

### Context of Change

Code for displaying domain links were repeated in the repo. New change abstracts a new component out to make code more readable.

### Type of Change

## Before / After

Created separate DomainLink component that can be used across repo for domain link displays.

## Test Plan

Ran 'npm run lint' and 'npm test'


src/containers/Transactions/Description/AccountSet.jsx
```diff
@@ -10,14 +11,7 @@ const AccountSet = (props) => {
...
         {t('set_domain_description')}{' '}
-        <a
-          className="domain"
-          rel="noopener noreferrer"
-          target="_blank"
-          href={`http://${decodeHex(data.tx.Domain)}`}
-        >
-          {decodeHex(data.tx.Domain)}
-        </a>
+        <DomainLink className="domain" decode domain={data.tx.domain} />
```
JST5000, line 14: My guess is your test failure is on this line - `data.tx.domain` should be `data.tx.Domain` with a capital D :) 

(My thought process to get there was since the test failure said `decodeHex doesn't work on undefined` to look back through until I find something that could be undefined by accident)

connorjchen: Yes, that seemed to fix the problem. Thanks haha.

JST5000: Meta: Once you've resolved all the comments from a reviewer, you can ask them to re-review in the top right corner of this page by clicking the "cycle" icon next to their username. 

image

JST5000 (approved):
Looks good to me :) (LGTM)

src/containers/shared/components/DomainLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+    </a>
+  )
+}
+
+DomainLink.defaultProps = {
+  className: '',
+  decode: false,
+}
+
+DomainLink.propTypes = {
```
ckniffen, line 31: If you export Props you do not have define prop types.  Also avoid using default export and use named exports instead.

src/containers/shared/components/DomainLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+      rel="noopener noreferrer"
+      target="_blank"
+      href={`http${decode ? '' : 's'}://${decode ? decodeHex(domain) : domain}`}
+    >
+      {decode ? decodeHex(domain) : domain}
+    </a>
+  )
+}
+
+DomainLink.defaultProps = {
```
ckniffen, line 26: default these in the component.

src/containers/shared/components/DomainLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+  decode?: boolean
+  domain: string
+}
+
+const DomainLink = (props: Props) => {
+  const { className, decode, domain } = props
+
+  return (
+    <a
+      className={className}
```
ckniffen, line 16: I would default the className of "domain" and use className to add additional classes.  `clx` is a good tool for that.

ckniffen:
Please add a unit test for `DomainLink`

src/containers/shared/components/DomainLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+      rel="noopener noreferrer"
+      target="_blank"
+      href={`http${decode ? '' : 's'}://${decode ? decodeHex(domain) : domain}`}
```
ckniffen, line 19: Always use `https`.  I think this was a bug that some places did it while others did not.

src/containers/shared/components/DomainLink.tsx
connorjchen: Aren't there some places (like src/containers/Token/TokenHeader/index.js) where we don't want className="domain"?

src/containers/shared/components/DomainLink.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { decodeHex } from '../transactionUtils'
+
+export interface Props {
+  className?: string
+  decode?: boolean
+  domain: string
+}
+
+const DomainLink = (props: Props) => {
```
mvadari, line 10: maybe this should be called `ExternalLink` for a slightly more general name?

ckniffen: We could have ExternalLink and have DomainLink use that.  The DomainLink component would then handle the decode and adding https and pass in the className of "domain".  External link would then handle rel and target.

ckniffen (approved):

PR #295, opened 2022-09-07
feat(build): dedupe bn.js during compilation

ckniffen:

This will reduce the bundle size by 55.44 KB gzipped and the fix was borrowed from https://github.com/XRPLF/xrpl.js/pull/1750

### Context of Change

The dependency tree contains two versions of bn.js, 4.12.0 and 5.2.0. Since there are more than one npm will only dedupe one.

In this case npm is choosing to dedupe 5.2.0 which is used by two dependencies which leaves 4.12.0 to be included 6 times.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


intelliot (approved):
I didn't test it, but it looks okay

pdp2121 (approved):

Issue #203, opened 2022-07-27
Refactor NFT transactions to use new transaction scheme


ckniffen:
Update NFTokenAcceptOffer, NFTokenBurn, NFTokenAcceptOffer, NFTokenCancelOffer, NFTokenCreateOffer, NFTokenCreateOffer  and NFTokenMint transactions to use scheme defined in https://github.com/ripple/explorer/tree/staging/src/containers/shared/components/Transaction.

Also we need to update components to use tsx, flesh out tests, and leverage the `<Amount>` and `<SimpleRow>` components.

ckniffen:
Closed by #341, #342, #343,  #344, and #345

PR #278, opened 2022-08-30
refactor(transactions) Switch SignerListSet to new transaction scheme

ckniffen:
## High Level Overview of Change
This PR primarily moves SignerListSet to the new transaction schema, hooks, typescript, and more tests in addition to the following:

- Allow transaction components to be more correctly typed
- Rename `.unsetregularkey` to `.unset` to make it more generic
apply to SetRegularKey and SignerListSet.
- Fix bug related to transactions that remove all signers #277
- Add clsx a small (228B) library for dynamic classNames

Fixes #277 and completes #205

### Context of Change

This change also allows each transaction to define its own fields and instructions.  It will help clean up the massive Instructions type.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

- [x] Updated files to React Hooks
- [x] Updated files to TypeScript

PR #302, opened 2022-09-08
refactor: NFTokenMint, NFTokenCreateOffer, NFTokenCancelOffer, NFTokenAcceptOffer, and NFTokenBurn

connorjchen:

OVERALL CHANGES FOR ALL NFTOKEN TRANSACTIONS
Split up into separate PRs:

- NFTokenAcceptOffer #341 
- NFTokenBurn #342 
- NFTokenCancelOffer #343 
- NFTokenCreateOffer #344 
- NFTokenMint #345 

### Context of Change

https://github.com/ripple/explorer/issues/203

### Type of Change

### TypeScript/Hooks Update

## Before / After

Refactored .jsx to .tsx and used SimpleRow and Amount

## Test Plan

Ran npm test


src/containers/Transactions/Simple/NFTokenCancelOffer.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
```
ckniffen, line 1: Can you use i18n throughout this file?

src/containers/Transactions/Simple/NFTokenCancelOffer.tsx
```diff
@@ -0,0 +1,71 @@
...
+        amount: { currency: string; amount: number; issuer: string }
+        offerID: string
+        tokenID: string
+        offerer: string
+      }[]
+    }
+  }
+}
+
+const NFTokenCancelOffer = (props: Props) => {
```
ckniffen, line 20: Use TransactionSimpleProps as the type here.  In #278 I add a way to define the types for the instructions.

src/containers/Transactions/Simple/NFTokenCancelOffer.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <>
+      {cancelledOffers.map(
+        ({
+          amount: { amount, issuer, currency },
+          offerID,
+          tokenID,
+          offerer,
+        }) => (
+          <>
+            <SimpleRow label="Offer ID">
```
ckniffen, line 37: In #278 I add a way to provide a className to `<SimpleRow>` which will add a css class to the `.value` element.  This will remove the need for an extra div and you can then move data-test up to `<SimpleRow>`

src/containers/Transactions/Simple/NFTokenCreateOffer.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <div className="dt">{tokenID}</div>
+        </div>
+      </SimpleRow>
+      <SimpleRow label="Amount">
+        <div data-test="amount">
+          <Amount
+            value={{
+              issuer: amount.issuer,
+              currency: amount.currency,
+              amount: amount.amount.toString(),
```
ckniffen, line 59: As part of #270 the need to pull these out individually will be fixed.  You will then be able to pass in `amount` to `value`.  I will pull out that fix into its own PR.

ckniffen:
Please move these files along with the parser to their new locations in `shared/components/Transaction/{TransactionType}`.

src/containers/Transactions/Simple/NFTokenCreateOffer.tsx
ckniffen: Created PR #303.

shawnxie999:
@ckniffen  Should we change to named export here?

src/containers/Transactions/Simple/NFTokenCancelOffer.tsx
```diff
@@ -0,0 +1,71 @@
...
+              </div>
+            </SimpleRow>
+          </>
+        ),
+      )}
+    </>
+  )
+}
+
+export default NFTokenCancelOffer
```
shawnxie999, line 71: @ckniffen Should we use named export here?

src/containers/Transactions/Simple/NFTokenAcceptOffer.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <>
+      {acceptedOfferIDs.map((offer) => (
+        <SimpleRow label="Offer ID" className="dt" data-test="offer-id">
+          {offer}
+        </SimpleRow>
+      ))}
+      {tokenID && (
+        <>
+          <SimpleRow label="Seller" className="account" data-test="seller">
```
mvadari, line 35: Labels should be lowercase

mvadari (commented):
This PR looks like it's just switching to TypeScript, and isn't using the new transactions scheme.

ckniffen:
T

> This PR looks like it's just switching to TypeScript, and isn't using the new transactions scheme.

There is more coming.  He was waiting on some stuff from #278

src/containers/shared/components/Transaction/NFTokenCreateOffer/parser.ts
```diff
@@ -1,6 +1,12 @@
-const formatAmount = require('./formatAmount')
+import { NFTokenCreateOfferInstructions } from './types'
+import { TransactionParser } from '../types'
 
-module.exports = (tx, meta) => {
+const formatAmount = require('../../../../../rippled/lib/txSummary/formatAmount')
```
ckniffen, line 4: use import instead of require.

connorjchen: When should require be used over import? I saw you use require for requiring utils for one of parsers in SignerListSet PR I think so thought this was similar

ckniffen: That must have been an oversight on my part

mvadari: It's really a JS vs TS thing

ckniffen: Oh i did it to avoid a circular dependency.  Will break out the function i was calling to its own file

ckniffen: I put in a PR to remove require from SignerListSet #326

PR #304, opened 2022-09-08
feat: Bar chart upgrade status color code stable version

pdp2121:

Add stable version indicator and upgrade status color code to version bar chart.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### Before
Screen Shot 2022-09-08 at 11 06 09 AM (2)

### After
Screen Shot 2022-09-15 at 5 52 16 PM


mvadari:
Is this PR supposed to be based on the other bar chart PR?

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  payload,
+  label,
+}: TooltipProps<ValueType, NameType>) => {
+  const { t } = useTranslation()
+  if (active) {
+    return (
+      <div className="custom-tooltip">
+        <p className="label">{t('version_display', { version: label })}</p>
+        <p className="value">
+          {t('nodes_count', { count: payload?.[0].payload.count })}
```
JST5000, line 42: If there's no payload, should `count` be 0 instead of `undefined`? (Currently I think this will default to `undefined`)

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+    )
+  }
+  return null
+}
+
+const renderLegend = (stableVersion: string | null, t: any) => (
+  <div className="legend">
+    <div className="legend-text">
+      <span>{t('current_stable_version')}:</span>
+      <span style={{ color: '#19FF83' }}>
```
JST5000, line 54: This color should use the constants defined below - might be worth bringing their definitions up to a higher level to accommodate that.

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+              dx={45}
+              dy={65}
+            />
+          </YAxis>
+          <Bar
+            dataKey="value"
+            barSize={30}
+            fill={purple}
+            radius={[4, 4, 0, 0]}
+            onMouseOver={(dataY) => {
```
JST5000, line 109: I'm a bit confused by the name `dataY` - is there a more descriptive name you could use here?

JST5000: maybe `topOfBar`?

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+            {stableVersion &&
+              data.map((_entry, index) => (
+                <Cell
+                  fill={
+                    // eslint-disable-next-line no-nested-ternary
+                    data[index].label === stableVersion
+                      ? green
+                      : data[index].label < stableVersion
+                      ? red
+                      : blue
```
JST5000, line 122: Nested ternary's are pretty hard to read - could you split one of the ternary operators into an `if/else` block?

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+          </Bar>
+          <Legend
+            align="right"
+            verticalAlign="top"
+            content={renderLegend(stableVersion, t)}
+          />
+          <Tooltip
+            content={<CustomTooltip />}
+            cursor={false}
+            position={{ x: posData.x - 45, y: posData.y - 90 }}
```
JST5000, line 135: Do any of these offsets have to be the same as offsets above? If so, can you extract them into variables so that no one accidentally updates one, but no the other, later on when modifying this? (Applies to all the magic numbers in this file)

src/containers/Network/NetworkTabs.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Tabs from '../shared/components/Tabs'
+
+interface Props {
+  selected: string
+}
+
+const NetworkTabs = (props: Props) => {
+  const { selected } = props
+  const { path = '/' } = useRouteMatch()
+  const tabs = ['nodes', 'validators', 'upgrade-status']
```
JST5000, line 12: This is defined manually in multiple spots - is there a single place we could define these tabs? (That way the code can follow the [DRY principle](https://www.digitalocean.com/community/tutorials/what-is-dry-development), and it's easier to update later on :) )

src/containers/Network/Nodes.jsx
```diff
@@ -41,7 +41,7 @@ class NodesPage extends Component {
   render() {
     const { nodes, locations, unmapped } = this.state
     const { path, t, language } = this.props
-    const tabs = ['nodes', 'validators']
+    const tabs = ['nodes', 'validators', 'upgrade-status']
```
JST5000, line 44: Affected by above comment :)

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+import BarChartVersion from './BarChartVersion'
+import NetworkTabs from './NetworkTabs'
+import Streams from '../shared/components/Streams'
+import Hexagons from './Hexagons'
+import { localizeNumber } from '../shared/utils'
+import { useLanguage } from '../shared/hooks'
+import Log from '../shared/log'
+
+export const UpgradeStatus = () => {
+  const [vList, setVList] = useState<any>([{}])
```
JST5000, line 13: Is there a better type than `any` to use here? (It's fine if there isn't :) )

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const { t } = useTranslation()
+  const language = useLanguage()
+
+  const aggregateData = (v: any[]) => {
+    if (!v) {
+      return []
+    }
+    let total = 0
+    const tempData: any[] = []
+    v.reduce((res, val) => {
```
JST5000, line 26: Are there more descriptive names than `res` and `val` you could use here? I'm finding it a bit easy to lose track of what the data actually is supposed to be.

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+export const UpgradeStatus = () => {
+  const [vList, setVList] = useState<any>([{}])
+  const [validations, setValidations] = useState([])
+  const [unlCount, setUnlCount] = useState(0)
+  const [stableVersion, setStableVersion] = useState<string | null>(null)
+  const { t } = useTranslation()
+  const language = useLanguage()
+
+  const aggregateData = (v: any[]) => {
```
JST5000, line 20: Can you use a more descriptive name than `v` here? - Maybe `versions`?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+    if (!v) {
+      return []
+    }
+    let total = 0
+    const tempData: any[] = []
```
JST5000, line 25: Ditto - what kind of data is this?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+        tempData.push(res[val.server_version])
+      }
+      res[val.server_version].count += 1
+      total += 1
+      return res
+    }, {})
+
+    return tempData
+      .map((item) => ({
+        label: item.server_version ? item.server_version : ' NA ',
```
JST5000, line 41: nit: I think the `/` is usually included for the Not Applicable acronym
```suggestion
        label: item.server_version ? item.server_version : ' N/A ',
```

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+        value: (item.count * 100) / total,
+        count: item.count,
+      }))
+      .sort((a, b) => (a.label > b.label ? 1 : -1))
+  }
+
+  useEffect(() => {
+    fetchStableVersion()
+    fetchData()
+    const interval = setInterval(fetchData, 5000)
```
JST5000, line 51: Should we really be fetching data so often? - If there's not very much downside to the request it's fine, but I'm not sure how costly the look up is. 

@ckniffen any thoughts on what a reasonable interval for this would be? (I would imagine that validators don't update their version very often, but if you just updated you may want to see that number update. Maybe just checking once when you load the page would be enough - and they have to refresh to see more?)

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }, [])
+
+  const fetchData = () => {
+    const url = '/api/v1/validators?verbose=true'
+
+    axios
+      .get(url)
+      .then((resp) => {
+        const newValidatorList: any = {}
+        resp.data.forEach((v: any) => {
```
JST5000, line 64: Could you give `v` a more descriptive name?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+          newValidatorList[v.signing_key] = v
+        })
+
+        setVList(() => newValidatorList)
+        setUnlCount(resp.data.filter((d: any) => Boolean(d.unl)).length)
```
JST5000, line 69: Could you give `d` a more descriptive name?

src/containers/Network/NetworkTabs.tsx
ckniffen: In this case it would have been 4 and is now 2.  The way this page works is very weird usually the content above the tabs does not change when you change tabs.

JST5000: What would have been 4? Not quite sure what the numbers are in response to

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+          setStableVersion(release.tag_name)
+          return false
+        }
+        return true
+      })
+    })
+  }
+
+  const updateValidators = (newValidations: any[]) => {
+    // @ts-ignore
```
JST5000, line 88: Why is there a `@ts-ignore` here? (Usually if we add thse, we include a comment on the same line explaining the purpose :) - it's also better to avoid wherever possible)

src/containers/Network/NetworkTabs.tsx
ckniffen: It was in 3 places before this PR.  A new tab was added which would have been a new copy paste but this component was create to reuse code for 3 of the instances.

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -0,0 +1,143 @@
...
+    })
+  }
+
+  const validatorCount = Object.keys(vList).length
+
+  return (
+    <div className="network-page">
+      <Streams validators={vList} updateValidators={updateValidators} />
+      {validatorCount && (
+        // @ts-ignore
```
JST5000, line 109: Ditto

src/containers/Network/Validators.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const language = useLanguage()
+  const { t } = useTranslation()
+  const [vList, setVList] = useState<any>([{}])
+  const [validations, setValidations] = useState([])
+  const [metrics, setMetrics] = useState({})
+  const [unlCount, setUnlCount] = useState(0)
+
+  useEffect(() => {
+    fetchData()
+    const interval = setInterval(fetchData, 5000)
```
JST5000, line 22: This interval is the same as before - is it worth factoring out into a single constant, or should it legitimately be configured separately?

JST5000: Oh nevermind - took me a sec to realize this is the same file just ts version - oops!

src/containers/Network/BarChartVersion.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const BarChartVersion = (props: Props) => {
+  const { data, stableVersion } = props
+  const { t } = useTranslation()
+  const [posData, setposData] = useState<BarCoordinates>({ x: 0, y: 0 })
+  const grey = '#9BA2B0'
+  const purple = '#8884d8'
+  const blue = '#1AA4FF'
+  const red = '#FF1A8B'
+  const green = '#19FF83'
```
JST5000, line 70: Should these be defined in the css instead of the React code?

src/containers/Network/BarChartVersion.tsx
pdp2121: It is used only at this place. Since recharts already has well-defined built-in components, those numbers are for just for fine-tuned this specific chart. Please let me know what you think

src/containers/Network/NetworkTabs.tsx
pdp2121: I fixed the Nodes.jsx file so these definitions are now only in one place.

src/containers/Network/UpgradeStatus.tsx
pdp2121: vList is retrieved from VHS and is quite complex so I put any for it.

src/containers/Network/UpgradeStatus.tsx
pdp2121: same as vList above

jonathanlei (approved):

src/containers/Network/BarChartVersion.tsx
pdp2121: Actually some numbers like chart height etc. can be put in one level higher for future use. Will make that update now.

src/containers/Network/BarChartVersion.tsx
pdp2121: Some colors will be assigned based on logic and also some field like stroke works differently when inside recharts compared to when set in css so I feel it is cleaner this way. I move those to shared/utils.js where constant are defined also.

mvadari: These should use CSS variables instead. It would probably be cleaner to change the CSS class name instead of defining colors in the code.

public/locales/en-US/translations.json
```diff
@@ -378,5 +379,10 @@
   "validator_history.date": "Date",
   "validator_history.missed": "Missed",
   "validator_history.score": "Score",
-  "other_chain_destination": "Other Chain Destination"
+  "other_chain_destination": "Other Chain Destination",
+  "% of total nodes":"% of Total Nodes",
```
mvadari, line 383: This should have `_` in it

src/containers/Network/index.js
```diff
@@ -38,7 +39,14 @@ class Network extends Component {
...
     const base = path.split('/:')[0]
-    return tab === 'nodes' ? <Nodes path={base} /> : <Validators path={base} />
+    // eslint-disable-next-line no-nested-ternary
+    return tab === 'upgrade_status' ? (
+      <UpgradeStatus path={base} />
+    ) : tab === 'validators' ? (
+      <Validators path={base} />
+    ) : (
+      <Nodes path={base} />
+    )
```
JST5000, line 49: I'd recommend turning this into an `if/else if/else` instead of a nested ternary

src/containers/shared/utils.js
```diff
@@ -21,6 +21,29 @@ export const NOT_FOUND = 404
 export const SERVER_ERROR = 500
 export const BAD_REQUEST = 400
 
+export const FETCH_INTERVAL = 5000
```
JST5000, line 24:
```suggestion
export const FETCH_INTERVAL_MILLIS = 5000
```

It's often helpful to include the units, even if they seem obvious from the way the number is formatted :)

src/containers/shared/utils.js
```diff
@@ -21,6 +21,29 @@ export const NOT_FOUND = 404
...
+  dx: -7,
+  height: 90,
+  tickLine: false,
+  minTickGap: -1,
+  interval: 0,
+}
+export const BAR_SIZE = 30
+export const BAR_RADIUS = [4, 4, 0, 0]
+export const BAR_TOOLTIPS = { cursor: false, posX: 45, posY: 70, offset: -10 }
+
```
JST5000, line 46: Since these constants aren't used in any other files, I wouldn't store them in the `shared/utils.js` - It's cleaner in my opinion to keep them in the same file they're used, and if they are eventually needed elsewhere we can move them up then :)

JST5000: (With my previous comment about raising things up - I was mostly talking about if there were specific numbers which had to stay in sync within these - such as dx/posX since they seem to have the same number it'd be worth having a variable to guarantee that. If they just happen to be the same though it is not required)

pdp2121: Oh I thought you meant those can be recycled for future charts. Yes it is just for styling and positioning this specific chart. I will just keep those numbers there like it was before and we can move some up in the future when new charts are in (Caleb have a stacked bar chart in mind now).

JST5000 (approved):
LGTM

PR #316, opened 2022-09-12
fix: npm 7 in ci/cd

justinr1234:

In CI/CD, we use node 14 which does not come with version 7 of npm. The build breaks for non-cached CI/CD builds based on the upgraded package.json `engines` setting.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


pdp2121 (approved):

.github/workflows/nodejs.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Setup npm version 7
+        run: npm i -g npm@7 --registry=https://registry.npmjs.org
```
ckniffen, line 27: Is setting registry required?

ckniffen (approved):

.github/workflows/nodejs.yml
justinr1234: Not sure, I need to do some more testing for caching

PR #317, opened 2022-09-12
feat(NFT): add nft_id to search bar

shawnxie999:
- Added search support for NFT id by making call to `nft_info` to check if its NFT (since `transactions` and `nft_id` share the same 64-digit hex syntax)
- Updated a few functions to be `async` within `Search` component to support for async API call 
- Updated test cases to support asynchronous operations that would otherwise fail. Used promise flushing(suggested in https://github.com/facebook/jest/issues/2157#issuecomment-279171856)
- Changed URL for NFTs from `token/:id` to `nft/:id`

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After

image

After clicking Enter
image


src/containers/Header/Search.jsx
```diff
@@ -17,16 +17,30 @@ import {
   HASH_REGEX,
 } from '../shared/utils'
 import './search.scss'
+import { getNFTInfo } from '../../rippled/lib/rippled'
+import SocketContext from '../shared/SocketContext'
 
-const getIdType = (id) => {
+const isNFTOrTransactions = async (id, rippledContext) => {
```
mvadari, line 23: Is there not a difference in the length/structure that can be used to identify whether something is an NFT or a transaction?

shawnxie999: Not that Im aware of, both seem to follow the same length. This was the approach discussed with @ledhed2222. But I would be happy to know if there exists any hacks around this

src/containers/Header/Search.jsx
```diff
@@ -17,16 +17,30 @@ import {
...
+  try {
+    await getNFTInfo(rippledContext, id)
```
ckniffen, line 25: Is it faster to look up an nft or a transaction?  We should choose the faster one.  At somepoint there should be a check to see if the hash matches both a transaction and an nft.

src/containers/Header/Search.jsx
```diff
@@ -17,16 +17,30 @@ import {
...
   }
   if (isValidClassicAddress(id)) {
     return 'accounts'
   }
   if (HASH_REGEX.test(id)) {
-    return 'transactions'
+    // Transactions and NFTs share the same syntax
+    // We must make an api call to ensure if it's one or the other
+    const idType = await isNFTOrTransactions(id, rippledContext)
+    return idType
```
ckniffen, line 43: No need for the extra variable.

src/containers/Header/Search.jsx
shawnxie999: will investigate 👍

src/containers/Header/Search.jsx
mvadari: NFTs do have to start with `000`, but a transaction hash could potentially also start with `000` (though it would be pretty rare - 1/4096 chance). So I guess this is the best option.

src/containers/Header/Search.jsx
```diff
@@ -17,16 +17,29 @@ import {
...
+import { getTransaction } from '../../rippled/lib/rippled'
+import SocketContext from '../shared/SocketContext'
 
-const getIdType = (id) => {
+const isNFTOrTransactions = async (id, rippledContext) => {
+  try {
+    await getTransaction(rippledContext, id)
+    return 'transactions'
+  } catch (e) {
+    return 'token'
```
mvadari, line 28: The key for NFTs should really be something like `nft` or `nftoken`, instead of using `token`. It's a different page type.

ckniffen: Agreed

src/containers/Header/Search.jsx
ckniffen: Why do NFTs always start with 000?

src/containers/Header/Search.jsx
shawnxie999: do you mean we should change the url to something like `/nft/:id`?

src/containers/Header/Search.jsx
shawnxie999: Im not sure if it sounds like a good approach tho. still could collide...

mvadari: If you look at how an NFT ID is derived, it starts with the the flags (the first 4 characters, in hex). There are no flags on an NFT with value greater than 8, so therefore the first 6 bytes (and the first 3 characters) are `0`.

src/containers/Header/Search.jsx
ledhed2222: yeah i agree

ledhed2222: also i don't like the name of this function with relation to its return type. this name should return a boolean and be named something like `isTransaction`. Given what this does, it should be named something like `determineTypeOfHash`

ledhed2222 (changes requested):
see comments

src/containers/Header/Search.jsx
ckniffen: @ledhed2222 soon we will have AMM Instances added to the list of 64 character hexs that we have to determine the type of.

src/containers/Header/Search.jsx
mvadari: Although this won't work as a general case, we could save some time by checking the first 3 characters - if it doesn't start with `000`, then we know it's a transaction.

ckniffen: 👍

shawnxie999: @ckniffen @mvadari i see there is a reserved flag for the first 3 bits that could change in the future, so im skeptical about checking the first 3 bits

shawnxie999: We could check the first 3 bits now , and remember to change it if the reserved flag is used

ckniffen (approved):

ledhed2222 (approved):

PR #325, opened 2022-09-13
Fix incorrect parsing of token ID in some NFTokenMint transactions

ledhed2222:

In some cases, when parsing an NFTokenMint transaction for its created token ID, we were incorrectly determining this value. This would happen in cases where the mint transaction resulted in a single page being split into two, instead of when a single page had the new token appended, or when a single page was created to hold the sole token.

### Context of Change

See above

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After

If you examine transaction `C4E598099A8B13C5C8D2B8C86385A37B64C2F62BFA1FB87196401BB6ACB67A69` in the NFT sandbox, you'll see that the parsed token ID is `0008000085D33F9C5481D3515029C9904D16F0109414D3A00000099A00000000`. This is incorrect; it should be `0008000085D33F9C5481D3515029C9904D16F0109414D3A0DCBA29BA00000020`.

## Test Plan

I created an account on the sandbox and minted 33 tokens for them. The final mint transaction previously would report the wrong token ID. Now it reports the correct token ID.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ src/**/*.css.map
 .env.test.local
 .env.production.local
 .env
+.node-version
 
 # Logs
 npm-debug.log*
--- a/package-lock.json
+++ b/package-lock.json
@@ -99,7 +99,7 @@
       },
       "engines": {
         "node": ">=14.15.0 <15.0.0",
-        "npm": ">=7.14.0 <8.0.0"
+        "npm": ">=7.14.0 <9.0.0"
       }
     },
     "../xrplf-typescript-style/packages/eslint-config": {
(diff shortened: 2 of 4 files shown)


ckniffen:
Can you add a test case with that example transaction?

ckniffen:
package-lock.json should not have changed.

.gitignore
```diff
@@ -16,6 +16,7 @@ src/**/*.css.map
 .env.test.local
 .env.production.local
 .env
+.node-version
```
ckniffen, line 19: Files like `.node-version` or `.nvmrc` should be commited to a repo.

ledhed2222:
> package-lock.json should not have changed.

yeah - it looks like it's because i'm using npm 8 instead of npm 7, which is allowed by our config. there's only a one line change in the package-lock so i think it's correct

ckniffen (approved):

shawnxie999 (approved):

PR #342, opened 2022-09-21
refactor: NFTokenBurn

connorjchen:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update


src/containers/shared/components/Transaction/NFTokenBurn/test/NFTokenBurnSimple.test.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
```
JST5000, line 1: nit: The folder name for this test is inconsistent with the other test folder (One is called `tests` and one is called `test`) - Not sure which one is normal, but we should standardize that.

src/containers/shared/components/Transaction/NFTokenBurn/types.ts
```diff
@@ -0,0 +1,10 @@
+import { TransactionCommonFields } from '../types'
+
+export interface NFTokenBurn extends TransactionCommonFields {
+  NFTokenID: string
+  Owner?: string
```
JST5000, line 5: The `Owner` is never displayed anywhere, is that intentional?

src/containers/shared/components/Transaction/NFTokenBurn/test/NFTokenBurnSimple.test.jsx
ckniffen: Great catch.  Looks like all other places it is called `test`.  The old location must have been a one off.

src/containers/shared/components/Transaction/NFTokenBurn/types.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export interface NFTokenBurnInstructions {
```
JST5000, line 8: I'm a bit confused why this is called `NFTokenBurnInstructions` - what is it instructions for? (Is there a simpler name?)

src/containers/shared/components/Transaction/types.ts
```diff
@@ -76,3 +76,19 @@ export interface TransactionMapping {
...
+  AccountTxnID?: string
+  Flags?: number
+  LastLedgerSequence?: number
+  Memos?: object[]
+  Signers?: object[]
+  SourceTag?: number
+  SignerPubKey?: string
+  TicketSequence?: number
+  TxnSignature?: string
+}
```
JST5000, line 94: @ckniffen for something like this would it make sense to import the types from xrpl.js to avoid duplication?

ckniffen: Yeah.  I want to do that.  Need to play around with getting just want we want out of xrpl.js and tree shaking the rest.  Will need some experimentation

src/containers/shared/components/Transaction/NFTokenBurn/types.ts
ckniffen: "Instructions" is what the legacy code called the result of the parser functions.  I choose to keep it though open to better names

src/containers/shared/components/Transaction/NFTokenBurn/types.ts
ckniffen: Probably not.  We should add that to the list of missed fields.

connorjchen: I added it because NFToken has Owner? in XRPL.org description. However, it doesn't seem like we use it anywhere. Does adding it to list of missed fields mean that we might display it on Explorer page in future? Also, should I keep the Owner in for this PR?

src/containers/shared/components/Transaction/NFTokenBurn/test/NFTokenBurnSimple.test.jsx
connorjchen: Do we prefer `test` going forward?

src/containers/shared/components/Transaction/NFTokenBurn/types.ts
ckniffen: Keeping it is fine because it is correct.  We should create a ticket to add the Owner field to the simple view

ckniffen: Created #356

ckniffen (approved):

public/locales/en-US/translations.json
```diff
@@ -386,6 +386,12 @@
   "validator_history.date": "Date",
   "validator_history.missed": "Missed",
   "validator_history.score": "Score",
+  "seller": "Seller",
+  "buyer": "Buyer",
+  "offerer": "Offerer",
+  "token_taxon": "Token Taxon",
+  "uri": "URI",
+  "owner_label": "Owner",
```
JST5000, line 394: Can you replace the original usages of `owner` -> `Owner's Account` with something like `owner_account`? Then have `owner` -> "Owner"

Otherwise I think the `owner_label` is a bit of a confusing name that other folks won't know to search for sinc emost of the other names for things in the translations are far more direct. (If it would change 5+ files, probably not worth it, but if it's just a couple places I think the consistency would be worth it)

connorjchen: I don't even see owner -> Owner's Account used anywhere.. guess I can change it? @ckniffen

JST5000 (approved):
LGTM

PR #348, opened 2022-09-21
fix: render bar chart only after fetched

pdp2121:

Only show the bar chart after data is fully fetched. Otherwise, a loading icon will be shown.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### Before
When data is not fully fetched, bar chart looks like this:
Screen Shot 2022-09-21 at 3 52 09 PM

### After
A loading icon will be shown when data is not fully fetched:
Screen Shot 2022-09-22 at 1 50 23 PM (2)


jonathanlei (approved):

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -136,7 +138,7 @@ export const UpgradeStatus = () => {
       <div className="wrap">
         <NetworkTabs selected="upgrade_status" />
         <div className="upgrade_status">
-          {vList && (
+          {validatorCount && colorLoaded && (
```
ckniffen, line 141: You could check if there is a current version instead of adding a new prop

ckniffen (approved):

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -111,32 +111,28 @@ export const UpgradeStatus = () => {
...
+            )}
+          </span>
+        </>
       </div>
-
       <div className="wrap">
         <NetworkTabs selected="upgrade_status" />
         <div className="upgrade_status">
-          {vList && (
+          {validatorCount && stableVersion && (
```
mvadari, line 135: There should be a "Loading" component if the network is still loading. That's a different state than "we couldn't fetch the data for some reason" and should be reflected accordingly.

pdp2121: Thanks Mayukha, update the code and also the PR description.

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -111,37 +115,31 @@ export const UpgradeStatus = () => {
   return (
     <div className="network-page">
       <Streams validators={vList} updateValidators={updateValidators} />
-      {validatorCount && (
+      {
         // @ts-ignore - Work around for complex type assignment issues
         <Hexagons data={validations} list={vList} />
```
mvadari, line 120: What's the type difficulty here?

ckniffen: There is weirdness with `connect` where typescript interprets the component as having no props

PR #353, opened 2022-09-22
Fix: bar chart version sort fix

pdp2121:

Fix sorting version for bar chart

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### Before
Screen Shot 2022-09-22 at 5 55 16 PM

### After
Screen Shot 2022-09-22 at 5 55 32 PM

### DIFF
--- a/src/containers/Network/BarChartVersion.tsx
+++ b/src/containers/Network/BarChartVersion.tsx
@@ -25,6 +25,7 @@ import {
   PURPLE,
   GREY_600,
   GREY_800,
+  isLaterVersion,
 } from '../shared/utils'
 import './css/barchart.scss'
 
@@ -68,8 +69,8 @@ const renderLegend = (stableVersion: string | null, t: any) => (
 
 const stableColorCode = (dataLabel: string, stableVersion: string) => {
   if (dataLabel === stableVersion) return GREEN
-  if (dataLabel < stableVersion) return RED
-  return BLUE
+  if (isLaterVersion(dataLabel, stableVersion)) return BLUE
+  return RED
 }
 
 const BarChartVersion = (props: Props) => {
--- a/src/containers/Network/UpgradeStatus.tsx
+++ b/src/containers/Network/UpgradeStatus.tsx
@@ -6,7 +6,11 @@ import BarChartVersion from './BarChartVersion'
 import NetworkTabs from './NetworkTabs'
 import Streams from '../shared/components/Streams'
 import Hexagons from './Hexagons'
-import { localizeNumber, FETCH_INTERVAL_MILLIS } from '../shared/utils'
+import {
+  localizeNumber,
+  FETCH_INTERVAL_MILLIS,
+  isLaterVersion,
+} from '../shared/utils'
 import { useLanguage } from '../shared/hooks'
 import Log from '../shared/log'
 
@@ -44,11 +48,11 @@ export const UpgradeStatus = () => {
 
     return tempData
       .map((item) => ({
-        label: item.server_version ? item.server_version : ' N/A ',
+        label: item.server_version ? item.server_version.trim() : ' N/A ',
         value: (item.count * 100) / total,
         count: item.count,
       }))
-      .sort((a, b) => (a.label > b.label ? 1 : -1))
+      .sort((a, b) => (isLaterVersion(a.label, b.label) ? 1 : -1))
   }
 
   useQuery(
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -78,6 +78,54 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
   padding: 0,
 }
 
+export const isLaterVersion = (source, target) => {
+  if (source === ' N/A ') return false
+  if (target === ' N/A ') return true
+  const sourceDecomp = source.split('.')
+  const targetDecomp = target.split('.')
+  // Compare major version
+  if (parseInt(sourceDecomp[0], 10) > parseInt(targetDecomp[0], 10)) return true
+  // Compare minor version
+  if (
+    parseInt(sourceDecomp[0], 10) === parseInt(targetDecomp[0], 10) &&
+    parseInt(sourceDecomp[1], 10) > parseInt(targetDecomp[1], 10)
+  )
+    return true
+  const sourcePatch = sourceDecomp[2].split('-')
+  const targetPatch = targetDecomp[2].split('-')
+  // Compare patch version
+  if (
+    parseInt(sourceDecomp[0], 10) === parseInt(targetDecomp[0], 10) &&
+    parseInt(sourceDecomp[1], 10) === parseInt(targetDecomp[1], 10)
+  ) {
+    if (parseInt(sourcePatch[0], 10) > parseInt(targetPatch[0], 10)) return true
+    if (
+      parseInt(sourcePatch[0], 10) === parseInt(targetPatch[0], 10) &&
+      sourcePatch.length > 1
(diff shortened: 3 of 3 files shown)


src/containers/shared/utils.js
```diff
@@ -78,6 +78,54 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
   padding: 0,
 }
 
+export const isLaterVersion = (source, target) => {
```
ckniffen, line 81: Please add a unit tests for this.  It is complicated and coverage

src/containers/shared/utils.js
```diff
@@ -78,6 +78,54 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
...
+  if (source === ' N/A ') return false
```
ckniffen, line 82: why does N/A have extra spaces?

JST5000: (Also, if you're going to use this special spacing in multiple files, please make it a shared constant :) )

src/containers/shared/utils.js
```diff
@@ -78,6 +78,54 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
...
+  if (target === ' N/A ') return true
+  const sourceDecomp = source.split('.')
+  const targetDecomp = target.split('.')
+  // Compare major version
+  if (parseInt(sourceDecomp[0], 10) > parseInt(targetDecomp[0], 10)) return true
+  // Compare minor version
+  if (
+    parseInt(sourceDecomp[0], 10) === parseInt(targetDecomp[0], 10) &&
```
JST5000, line 90: Instead of using `parseInt` over and over again you can extract these into variables :) 

```
const left = parseInt(sourceDecomp[0], 10)
const right = parseInt(sourceDecomp[1], 10)
```

General rule: If you are doing the same computation over and over again, can you calculate once and reuse throughout?

src/containers/shared/utils.js
```diff
@@ -78,6 +78,54 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
...
+    parseInt(sourceDecomp[1], 10) > parseInt(targetDecomp[1], 10)
+  )
+    return true
+  const sourcePatch = sourceDecomp[2].split('-')
+  const targetPatch = targetDecomp[2].split('-')
+  // Compare patch version
+  if (
+    parseInt(sourceDecomp[0], 10) === parseInt(targetDecomp[0], 10) &&
+    parseInt(sourceDecomp[1], 10) === parseInt(targetDecomp[1], 10)
```
JST5000, line 99: This logic can be simplified in two main ways:
1. Since you're returning booleans, you can return conditionals instead of returning `true` and `false` directly. 
2. You can rely on your return values in order to avoid checking for the same conditions twice.

Ex. 
Psuedocode:
1. Are the major versions the same? If NOT -> Return left > right (Since they aren't the same, this should give you the same output as currently)
2. Are the minor versions the same? If NOT -> Return minorLeft > minorRight
3. Are the patch version the same? If NOT -> Return patchLeft > patchRight

For the `rcb` - you may want to just rely on alphanumeric comparison, since if they release a release candidate `c`, that should be later anyway.

(This was a lot of words, I may have misunderstood some complexity in the problem - if so sorry!)

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -44,11 +48,11 @@ export const UpgradeStatus = () => {
 
     return tempData
       .map((item) => ({
-        label: item.server_version ? item.server_version : ' N/A ',
+        label: item.server_version ? item.server_version.trim() : ' N/A ',
         value: (item.count * 100) / total,
         count: item.count,
       }))
-      .sort((a, b) => (a.label > b.label ? 1 : -1))
+      .sort((a, b) => (isLaterVersion(a.label, b.label) ? 1 : -1))
```
mvadari, line 55: Will there be a big performance hit as a result of this? Is there any way we can mitigate that, maybe with some extra backend processing?

src/containers/shared/utils.js
pdp2121: Nice fix! Thanks Jackson! For the `rcb`, I cannot rely solely on alphanumeric sort since 1.10.0-b10 will comes before 1.10.0-b9 even though it should be the other way. I fixed part of the code accordingly to your suggestion.

src/containers/shared/utils.js
pdp2121: I added it for sorting and spacing. With the new isLaterVersion sort, I can just remove those extra spaces now.

src/containers/Network/UpgradeStatus.tsx
pdp2121: I dont think it would affect the performance by much since isLaterVersion has constant runtime.

pdp2121: @ckniffen please correct me if I'm wrong

ckniffen: @pdp2121 best way is to put a breakpoint and see how often it is hit

jonathanlei (approved):

src/containers/Network/UpgradeStatus.tsx
pdp2121: I fixed it so it ran only when refetching happens (every 5 seconds).

mvadari: It might have constant runtime, but it still has to do more processing than just comparing `label`, and it has to do it for each element in the list. 
Does this run nlogn times due to the sort, or is it just n times?

src/containers/Network/BarChartVersion.tsx
```diff
@@ -68,8 +69,8 @@ const renderLegend = (stableVersion: string | null, t: any) => (
 
 const stableColorCode = (dataLabel: string, stableVersion: string) => {
   if (dataLabel === stableVersion) return GREEN
-  if (dataLabel < stableVersion) return RED
-  return BLUE
+  if (isLaterVersion(dataLabel, stableVersion)) return BLUE
+  return RED
```
mvadari, line 73: Looking at this diff, would it make more sense to call the method `isEarlierVersion` and have it return the opposite result? I think that might make the code a little more readable, because right now I have to think about it a little backwards re sorting

src/containers/shared/utils.js
```diff
@@ -78,6 +78,61 @@ const FORMAT_PRICE_DEFAULT_OPTIONS = {
   padding: 0,
 }
 
+export const isLaterVersion = (source, target) => {
```
mvadari, line 81: How does this method handle versions that include a commit hash?

src/containers/Network/UpgradeStatus.tsx
pdp2121: Yes, this runs nlogn times due to the sort

src/containers/shared/utils.js
pdp2121: Is there a structural/comparison rule for that? I just followed the structure described here:
https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/BuildInfo.h#L46-L72

pdp2121: As we discussed, versions with commit hash will be handled in a separate PRs to remove the hash.

src/containers/Network/UpgradeStatus.tsx
pdp2121: @ckniffen any ideas?

ckniffen (approved):

PR #354, opened 2022-09-22
feat: show tooltip only on mouse enter

pdp2121:

Currently, tooltips appear when hover on bar chart cursor region. Change its visibility to only when hovering on bar region to improve user experience.
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### Before
Screen Shot 2022-09-22 at 6 37 40 PM (2)

#### Mouse hover outside of bar region
Screen Shot 2022-09-22 at 6 40 43 PM (2)
#### Mouse hover inside of bar region
Screen Shot 2022-09-22 at 6 41 50 PM (2)


src/containers/Network/BarChartVersion.tsx
```diff
@@ -129,6 +137,7 @@ const BarChartVersion = (props: Props) => {
               backgroundColor: GREY_600,
               borderRadius: 8,
               border: ['1px solid', GREY_800].join(' '),
+              opacity: showTooltips ? '100%' : '0',
```
JST5000, line 140: This might end up being confusing because there are multiple ways of indicating whether a tooltip should appear (Even if you do the function, the opacity will basically override it) - Maybe it'd be better to have a single method of deciding to show/hide the tooltip and using it across the board?

pdp2121: @ckniffen

src/containers/Network/BarChartVersion.tsx
```diff
@@ -129,6 +137,7 @@ const BarChartVersion = (props: Props) => {
               backgroundColor: GREY_600,
               borderRadius: 8,
               border: ['1px solid', GREY_800].join(' '),
```
ckniffen, line 139: This could use a template string

src/containers/Network/BarChartVersion.tsx
ckniffen: What are the other places?  Also this spot uses recharts Tooltip.

ckniffen (approved):

src/containers/Network/BarChartVersion.tsx
```diff
@@ -108,7 +110,14 @@ const BarChartVersion = (props: Props) => {
               dy={80}
             />
           </YAxis>
-          <Bar dataKey="value" barSize={30} fill={PURPLE} radius={[4, 4, 0, 0]}>
+          <Bar
+            dataKey="value"
+            barSize={30}
+            fill={PURPLE}
+            radius={[4, 4, 0, 0]}
+            onMouseOver={() => setShowTooltips(true)}
```
mvadari, line 118: What are the performance implications of this change? Is it really slow/does it make other things really slow?

src/containers/shared/utils.js
```diff
@@ -49,6 +49,8 @@ export const GREEN = '#19FF83'
 export const GREY_600 = '#656E81'
 export const GREY_800 = '#383D47'
 
+export const TOOLTIPS_BORDER = ['1px solid', GREY_800].join(' ')
```
ckniffen, line 52: Use a template string here

src/containers/Network/BarChartVersion.tsx
ckniffen: It shouldn't be slow these events don't continually fire.

mvadari (approved):

PR #384, opened 2022-10-04
feat: Link to token page when a currency is shown

### LINKED ISSUE #72: Add link to token page from transaction
For example, there is no real way to go from [this transaction page](https://livenet.xrpl.org/transactions/82FA369A7727DAF1289D092812BAE34FD0B26636EF33B14ADDA4A5763D6C1FB6) to the [page for the token that is traded in this transaction](https://livenet.xrpl.org/token/53686962614E4654000000000000000000000000.rnRXAnVZTyattZXEpKpgTyvdm17DpjrzSZ), without digging through the raw JSON yourself. 

In fact, the token isn't even fully written out - it just has the hex of the token and not the issuer listed on the "simple" tab.

ckniffen:

The `<Currency>` component will now craft links that point a token's page instead of the issuer account page.  Any places that use the `<Amount>` component ensure that issuer is correctly passed into `<Currency>`.

With the Payment PR out created the most common transactions will be covered.

Closes #72

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After

TODO: Add screenshots


src/containers/shared/components/Currency.tsx
```diff
@@ -1,47 +1,33 @@
 import React from 'react'
-import PropTypes from 'prop-types'
-import Account from './Account'
+import { Link } from 'react-router-dom'
 
 // https://xrpl.org/currency-formats.html#nonstandard-currency-codes
 const NON_STANDARD_CODE_LENGTH = 40
 
-interface Props {
+export interface Props {
```
mvadari, line 7: Why is this exported?

mvadari (approved):

src/containers/shared/components/Currency.tsx
ckniffen: So that PropTypes arent needed.  Otherwise it complains about no default proptypes set.

jonathanlei (approved):
LGTM

PR #386, opened 2022-10-04
chore: Tweak tabs, cleanup PropTypes, and default exports

ckniffen:

Update tab title attributes to be a translated value and remove unnecessary PropTypes and a few default exports.

### Context of Change

Several components had both prop interfaces and PropTypes defined. Switched more locations to explicit exports. This uncovered several places that imported components using the name "Account" that weren't correct.

### Type of Change


JST5000:
Can you make the `_` versions of the site also redirect to the same page? (That way old links to specific transactions won't break)

ckniffen:
> Can you make the `_` versions of the site also redirect to the same page? (That way old links won't break)

Only the new upgrade status page is affected

jonathanlei (approved):

pdp2121 (approved):

PR #394, opened 2022-10-07
feat: use `ws` connection when connecting to a local node

mvadari:

The Explorer usually uses a `wss` connection to connect to nodes. To connect to a node that doesn't have this setup, such as a node running locally, you need to switch to a `ws` connection. This PR solves that problem.

This PR also edits the sidechain URL display to include the port (this is confusing in localhost, when running multiple local nodes).

There is also a new environment variable that can be defined, when working with an external network that doesn't have SSL set up.

### Context of Change

When running a local node, you usually don't have a secure connection setup, so the `wss` connection fails.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After
Before:
image

After:
image

## Test Plan

Works locally. I was able to connect to both a local node (with a `ws` connection) and an external node (with a `wss` connection).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,7 +23,7 @@ REACT_APP_ENVIRONMENT=mainnet
 #VHS endpoint url
 REACT_APP_DATA_URL=https://data.xrpl.org/v1/network
 
-# Whether to use ws instead of wss
+# Whether to use ws instead of wss (boolean)
 # Only used locally (the deployed Explorer requires wss)
 REACT_APP_INSECURE_WS=0
 
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,10 @@ REACT_APP_ENVIRONMENT=mainnet
 #VHS endpoint url
 REACT_APP_DATA_URL=https://data.xrpl.org/v1/network
 
+# Whether to use ws instead of wss
+# Only used locally (the deployed Explorer requires wss)
+REACT_APP_INSECURE_WS=0
+
 #Google Credentials for BigQuery, DO NOT COMMIT THESE
 GOOGLE_APP_PROJECT_ID=
 GOOGLE_APP_PRIVATE_KEY=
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -23,6 +23,10 @@ REACT_APP_ENVIRONMENT=mainnet
 #VHS endpoint url
 REACT_APP_DATA_URL=https://data.xrpl.org/v1/network
 
+# Whether to use ws instead of wss
+# Only used locally (the deployed Explorer requires wss)
+REACT_APP_INSECURE_WS=0
```
JST5000, line 28:
```suggestion
# Whether to use ws instead of wss
# Only used locally (the deployed Explorer requires wss)
# 1 = ws, 0 = wss (default)
REACT_APP_USE_INSECURE_WS=0
```

Two suggestions here:
1. Clarifying that we're using 0 as a boolean with the examples of how to use it
2. Adding a word to the name to clarify that this is not a string variable like the other `REACT_APP_...` and is instead a boolean

ckniffen (approved):

src/containers/shared/SocketContext.ts
```diff
@@ -0,0 +1,43 @@
...
+    wsUrls.push.apply(wsUrls, [
+      `${prefix}://${rippledHost}:${process.env.REACT_APP_RIPPLED_WS_PORT}`,
+      `${prefix}://${rippledHost}:443`,
+    ])
+  }
+  const socket = new XrplClient(wsUrls)
+  const hasP2PSocket =
+    process.env.REACT_APP_P2P_RIPPLED_HOST != null &&
+    process.env.REACT_APP_P2P_RIPPLED_HOST !== ''
+  // @ts-ignore - will be removed eventually
```
JST5000, line 30: When will this be removed and why? (Just so future readers can tell if this still makes sense to ignore)

JST5000 (approved):
LGTM - just one comment on a comment

src/containers/shared/SocketContext.ts
ckniffen: It will be removed when we support clio only.

PR #395, opened 2022-10-07
feat(nft): add Issuer field to NFTokenMint tx

shawnxie999:
show issuer and transfee fee for NFTokenMint transactions, if they exist

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
image


shawnxie999:
@ledhed2222 @ckniffen just confirming, within the nftokenmint type, transfer fee is optional, is this expected? some test cases dont have this field either, just curious because i thought transfer fee always exists
```
export interface NFTokenMint extends TransactionCommonFields {
  NFTokenTaxon: number
  Issuer?: string
  TransferFee?: number
  URI?: string
}
```

ckniffen:
> @ledhed2222 @ckniffen just confirming, within the nftokenmint type, transfer fee is optional, is this expected? some test cases dont have this field either, just curious because i thought transfer fee always exists
> 
> ```
> export interface NFTokenMint extends TransactionCommonFields {
>   NFTokenTaxon: number
>   Issuer?: string
>   TransferFee?: number
>   URI?: string
> }
> ```

The docs also list it as being optional.

src/containers/shared/components/Transaction/NFTokenMint/Simple.tsx
```diff
@@ -4,12 +4,23 @@ import { useTranslation } from 'react-i18next'
...
 export const Simple: TransactionSimpleComponent = ({
   data,
 }: TransactionSimpleProps<NFTokenMintInstructions>) => {
-  const { tokenID, tokenTaxon, uri } = data.instructions
+  const { tokenID, tokenTaxon, uri, transferFee, issuer } = data.instructions
   const { t } = useTranslation()
+  const language = useLanguage()
+  const formattedFee =
+    transferFee &&
+    `${localizeNumber((transferFee / 1000).toPrecision(5), language, {
```
ckniffen, line 19: TransferFee fraction of the amount a NFToken is transferred for.  It shouldn't be treated as a currency.

src/containers/shared/components/Transaction/NFTokenMint/Simple.tsx
```diff
@@ -26,6 +37,20 @@ export const Simple: TransactionSimpleComponent = ({
       <SimpleRow label={t('uri')} className="dt" data-test="token-uri">
         {uri}
       </SimpleRow>
+      {transferFee && (
+        <SimpleRow
+          label={t('transfer_fee')}
+          className="dt"
```
ckniffen, line 43: `dt` does nothing now used this way.

src/containers/shared/components/Transaction/NFTokenMint/Simple.tsx
```diff
@@ -26,6 +37,20 @@ export const Simple: TransactionSimpleComponent = ({
...
+          data-test="token-fee"
+        >
+          {formattedFee}
+        </SimpleRow>
+      )}
+      {issuer && (
+        <SimpleRow label={t('issuer')} className="dt" data-test="token-issuer">
```
ckniffen, line 50: Same here.

ledhed2222:
> > @ledhed2222 @ckniffen just confirming, within the nftokenmint type, transfer fee is optional, is this expected? some test cases dont have this field either, just curious because i thought transfer fee always exists
> > ```
> > export interface NFTokenMint extends TransactionCommonFields {
> >   NFTokenTaxon: number
> >   Issuer?: string
> >   TransferFee?: number
> >   URI?: string
> > }
> > ```
> 
> The docs also list it as being optional.

If not provided, the transfer fee will be 0. I'm not sure how this should be displayed in this case, since if you don't provide it it shouldn't be shown in the tx summary. it will be shown in the NFT page itself.

ledhed2222:
@shawnxie999 @ckniffen - i'm not sure this makes sense to add, IMO. Here's why:

people who want to verify _every_ field in a transaction can already do that using the "raw" tab. With `transferFee`, if it's not specified the default is 0. Similarly with issuer, if not specified the issuer is the account that submitted the tx. These data points can also be found in the NFT page itself.

I think that by adding them to the simple view, it could be confusing for users why these values appear sometimes and not othertimes, despite the fact that an NFT will always have an issuer and a transfer fee, even if neither are specified in the transaction.

I'm also not sure who it helps to add to the simple view.

mvadari:
> @shawnxie999 @ckniffen - i'm not sure this makes sense to add, IMO. Here's why:
> 
> people who want to verify _every_ field in a transaction can already do that using the "raw" tab. With `transferFee`, if it's not specified the default is 0. Similarly with issuer, if not specified the issuer is the account that submitted the tx. These data points can also be found in the NFT page itself.
> 
> I think that by adding them to the simple view, it could be confusing for users why these values appear sometimes and not othertimes, despite the fact that an NFT will always have an issuer and a transfer fee, even if neither are specified in the transaction.
> 
> I'm also not sure who it helps to add to the simple view.

Perhaps a better option would be to add it to the Detailed view.

ckniffen:
I like it on the simple view because it will show all parties in involved in the transaction and link to the account where the NFT will be after the transaction

mvadari (approved):

PR #398, opened 2022-10-08
refactor: Normalize headers on Token and NFT pages

ckniffen:

Updated several places to use `<h2>`s and defined a global style. The only place that previously used a `<h2>` was the address on the account page. That is more an h1 so update it there. In the future more pages will properly use `<h1>` tags.

### Context of Change

This has the benefit of 33 less lines of code and fixes weird spacing on the token page for dex pairs.

NFT and Token pages both used the class `.token-page` so I updated the NFT page to `.nft-page` because there was bleed between them.

The `TokenTransactionsTable` had its header moved out of the component to pave the way for it to use the `TransactionTable`.

### Type of Change

## Before
H2 Token - Before
H2 Account - Before
)
H2 NFT - Before

## After
H2 Token - After
H2 Account - After
H2 NFT - After

## Future Tasks

Normalize the styles for metrics in the headers and `.header-container` and `.nft-header-container`.


pdp2121 (approved):
LGTM

PR #401, opened 2022-10-11
feat: rename `sidechain.xrpl.org` and branding to `custom.xrpl.org`

mvadari:

This PR renames everything sidechain-focused.

### Context of Change

Prepping for renaming sidechain.xrpl.org to custom.xrpl.org

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

CI passes.


.env.example
```diff
@@ -14,10 +14,10 @@ REACT_APP_TESTNET_LINK=
 REACT_APP_DEVNET_LINK=
 REACT_APP_AMM_LINK=
 REACT_APP_NFTSANDBOX_LINK=
-REACT_APP_SIDECHAIN_LINK=
+REACT_APP_CUSTOMNETWORK_LINK=
```
JST5000, line 17: I'd recommend splitting up the words here like you do elsewhere to preserve the intuitive spelling.
```suggestion
REACT_APP_CUSTOM_NETWORK_LINK=
```

mvadari: I kept them smushed together because `NFTSANDBOX` does the same. It was more intuitive to me to have the format be `REACT_APP_[network type]_LINK`

src/containers/Header/index.js
```diff
@@ -154,15 +154,15 @@ const Header = (props) => {
...
   const urlLinkMap = {
     ...STATIC_ENV_LINKS,
-    // TODO: store previous sidechains in cookies, add them here
+    // TODO: store previous custom networks in cookies, add them here
   }
 
   const rippledUrl = getSocketUrl(rippledSocket)
 
-  if (process.env.REACT_APP_ENVIRONMENT === 'sidechain') {
+  if (process.env.REACT_APP_ENVIRONMENT === 'custom') {
```
JST5000, line 162: Since for the most part you use `custom network` together, I think it makes sense to keep that convention throughout
```suggestion
  if (process.env.REACT_APP_ENVIRONMENT === 'custom_network') {
```

src/containers/Header/test/Header.test.js
```diff
@@ -97,16 +97,16 @@ describe('Header component', () => {
...
+    it('redirect on custom network input works', () => {
       const wrapper = createWrapper()
 
-      const sidechainInput = wrapper.find('[className="sidechain_input"]')
-      sidechainInput.prop('onKeyDown')({
+      const customInput = wrapper.find('[className="custom_network_input"]')
+      customInput.prop('onKeyDown')({
         key: 'Enter',
-        currentTarget: { value: 'sidechain_url' },
+        currentTarget: { value: 'custom_url' },
```
JST5000, line 106: Ditto
```suggestion
        currentTarget: { value: 'custom_network_url' },
```

src/containers/shared/css/variables.scss
```diff
@@ -57,6 +57,7 @@ $white-transparent: rgb(255 255 255 / 75%);
 $testnet: #e54d00;
 $devnet: #008ae5;
 $nft-sandbox: #791aff;
+$custom: #faff1a;
```
JST5000, line 60: ditto

src/containers/Header/index.js
mvadari: The URL will be `custom.xrpl.org`, which is why I kept the env value `custom`

src/containers/Header/test/Header.test.js
mvadari: This is just a placeholder for the test, so it doesn't matter what the value is

.env.example
JST5000: Stylistic difference - I don't have much of a preference (Although I did just notice that you could read the NFTSANDBOX as NFTS AND BOX which I find funny :P )

src/containers/shared/components/Streams.jsx
```diff
@@ -118,7 +118,7 @@ class Streams extends Component {
   componentDidMount() {
     this.connect()
     this.updateNegativeUNL()
-    if (process.env.REACT_APP_ENVIRONMENT !== 'sidechain') {
+    if (process.env.REACT_APP_ENVIRONMENT !== 'custom') {
```
ckniffen, line 121: Should we hide these metric sections from the homepage if we don't retrieve the data for them?

A change for another PR.

ckniffen:
@buggsbunnies @nataliacebotarii does the old sidechains logo make sense with this rebrand?

https://github.com/ripple/explorer/blob/7d4b9d08e1cc9f42be529c9e6a266d43e321690e/src/containers/shared/images/custom_network_logo.svg

src/containers/shared/components/Streams.jsx
mvadari: We don't retrieve data from the server, it's generated via the data the Explorer stores locally. See line 502

ckniffen (approved):

JST5000 (approved):
LGTM

PR #405, opened 2022-10-12
fix(css): clean up spacing on home page

mvadari:

This PR changes the spacing on the home page (where the ledgers scroll by). Now, if there isn't very much on the page, there won't be a scroll bar.

### Context of Change

The need for a scroll bar was more noticeable in iframes in a page I'm building for the sidechain CLI.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

Before:
image

After:
image

## Test Plan

Works locally.

### DIFF
--- a/src/containers/App/app.scss
+++ b/src/containers/App/app.scss
@@ -1,7 +1,7 @@
 @import '../shared/css/variables';
 
 .app {
-  position: relative;
+  flex-grow: 1;
 
   .content {
     display: flex;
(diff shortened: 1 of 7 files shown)


src/containers/App/app.scss
```diff
@@ -1,7 +1,7 @@
 @import '../shared/css/variables';
 
 .app {
-  position: relative;
+  flex-grow: 1;
 
   .content {
     display: flex;
```
ckniffen, line 7: You can get rid of the following because it is covered by `desktop-up`
```
   @include for-size(big-desktop-up) {
      min-height: $min-height;
    }
 ```

ckniffen (approved):

justinr1234 (approved):

PR #416, opened 2022-10-19
fix: ledger metrics centered instead of right-justified

mvadari:

This PR fixes the justification of the ledger metrics on the main page. Instead of being right-justified, it is centered (as previous).

### Context of Change

#408 changed too much, this is a fix for that.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

Before:
image

After:
image

## Test Plan

Works locally.

### DIFF
--- a/src/containers/Ledgers/css/ledgerMetrics.scss
+++ b/src/containers/Ledgers/css/ledgerMetrics.scss
@@ -7,7 +7,6 @@
   text-align: center;
 
   @include for-size(tablet-landscape-up) {
-    display: flex;
     width: calc(100% - 124px);
     justify-content: space-between;
     padding: 0px 12px;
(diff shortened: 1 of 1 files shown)


src/containers/Ledgers/css/ledgerMetrics.scss
```diff
@@ -6,9 +6,11 @@
   background: $black;
   text-align: center;
 
-  @include for-size(desktop-up) {
-    display: block;
-    max-width: inherit;
```
mvadari, line 11: This is the same as what gets automatically inherited now, so it's unnecessary

ckniffen (approved):

JST5000 (approved):
LGTM

PR #417, opened 2022-10-25
Bug(tx): Fetch TokenID from only where PreviousFields.NFTokens is specified

shawnxie999:
- fetch only from the NFTokenPage that has `NFTokens` in its `PreviousFields`

### Context of Change
There is an edge case problem when parsing TokenID. It occurs because we have two NFTokenPage modified, and only one of them is used to find the TokenID

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Before 
image

After
image


ledhed2222:
@ckniffen see my comments to you in slack. i'm not understanding why the existing code didn't work

ledhed2222 (approved):
excellent ty

PR #418, opened 2022-10-25
feat(tx): Add NFTokenID link to NFT transaction pages

shawnxie999:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Before
image
image
image

After
image
image
image


ckniffen:
Can you use `NFTokenLink` on the NFT tab for an account too?

src/containers/shared/components/NFTokenLink.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+export interface NFTokenLinkProps {
+  link?: boolean
```
ckniffen, line 5: Where is this set to true?

ckniffen (approved):

src/containers/shared/components/NFTokenLink.tsx
ledhed2222: it looks like it's defaulted to true regardless. so where is it set to _false_ is the question? And why would be it be set to false?

also, this isn't a good name for an argument. `showLink` or `displayLink` would be better. Furthermore, the name of the component - `NFTokenLink` tells me that this is always a link. Why would I use this without a link?

ledhed2222 (changes requested):

src/containers/shared/components/NFTokenLink.tsx
shawnxie999: yeah fair, i was thinking of following the same convention used in `Account` component. but i dont think non-link will be used anywhere. I will remove it

ledhed2222 (approved):

PR #443, opened 2022-11-02
fix: improve barchart load time

mvadari:

This PR improves the loading of the Upgrade Status bar chart, from ~5 seconds to almost instantaneously.

It was effectively waiting for the second `fetch` before loading the bar chart, due to a bug.

I also fixed a few other minor issues along the way, like the status codes of the Explorer backend, and adding a key to a map.

### Context of Change

The bar chart took about 5 seconds to load, despite not needing the time for processing. 

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

The bar chart now loads almost instantaneously.

## Test Plan

Works locally.

### DIFF
--- a/server/routes/v1/tokenDiscovery.js
+++ b/server/routes/v1/tokenDiscovery.js
@@ -133,10 +133,13 @@ async function cacheTokensList() {
 // Starts the caching process for bigquery
 function startCaching() {
   // Only run if on mainnet (the tokens page doesn't exist on devnet/testnet)
-  if (process.env.REACT_APP_ENVIRONMENT !== 'mainnet') {
-    return
-  }
+  // TODO: uncomment below code when token discovery is used again
+  // if (process.env.REACT_APP_ENVIRONMENT !== 'mainnet') {
+  //   return
+  // }
+  return
   // Initialize the cache
+  // eslint-disable-next-line no-unreachable -- unused code for now
   cacheTokensList()
   // Cache every TIME_INTERVAL ms (only starts after one interval)
   const intervalId = setInterval(() => cacheTokensList(), TIME_INTERVAL)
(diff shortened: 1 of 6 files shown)


server/routes/v1/tokenDiscovery.js
```diff
@@ -133,10 +133,13 @@ async function cacheTokensList() {
 // Starts the caching process for bigquery
 function startCaching() {
   // Only run if on mainnet (the tokens page doesn't exist on devnet/testnet)
-  if (process.env.REACT_APP_ENVIRONMENT !== 'mainnet') {
-    return
-  }
+  // TODO: uncomment below code when token discovery is used again
```
ckniffen, line 136: We use token discovery still for token detail page.

mvadari: oh right, uncommenting...

server/routes/v1/validators.js
```diff
@@ -70,8 +72,10 @@ module.exports = (req, res) => {
...
 
   if (req.query.unl === process.env.REACT_APP_VALIDATOR) {
-    return res.send(validatorSummary.filter((val) => val.unl === req.query.unl))
+    return res
+      .status(200)
+      .json(validatorSummary.filter((val) => val.unl === req.query.unl))
   }
 
-  return res.send(validatorSummary)
+  return res.status(200).json(validatorSummary)
```
ckniffen, line 80: Does this ensure that the correct headers are set?

mvadari: I don't think headers are used at all.

ckniffen (approved):

pdp2121 (approved):
LGTM

PR #447, opened 2022-11-03
Add Destination field to NFTokenCreateOffer if exists

shawnxie999:
### Context of Change
Its useful to show the Destination on sell offer if it is in raw data

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
Before 
image

After
image


ckniffen (approved):
Killer

mvadari (approved):

ledhed2222:
What about buy offer?

ckniffen:
> What about buy offer?

This should work on both but worth looking into.

shawnxie999:
> What about buy offer?

Just curious how does it apply to buy offer? I thought it only makes sense for a sell offer where the seller only wants a specific account to buy the nft only

ledhed2222:
That was previously true, but a robust use of brokered mode requires use of
destination on both buy and sell offers where the destination in both cases
is the broker. Otherwise the seller can select whichever buy offer they
want and stiff the broker out of any fees.

PR #448, opened 2022-11-04
Clean rows for token-table

shawnxie999:
### Context of Change
Cleaning up duplicate code for .token-table rows

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update


src/containers/shared/components/TokenTableRow.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const TokenTableRow = (props: TokenTableRowProps) => {
+  const { shouldDisplay = true, label, value } = props
+  return (
+    <>
+      {shouldDisplay && (
+        <tr className="row">
+          <td className="col1">{label}</td>
+          <td className="col2">{value}</td>
+        </tr>
```
mvadari, line 17: I think this is the same as `SimpleRow` for transactions. Perhaps those should be merged into something that's just `Row` or `TableRow` or something.

shawnxie999: Not sure, the token-table  is using `<table>` tag, I assume  refactoring the entire token table would be need if we want to use the SimpleRow

mvadari: Oh, good point - nvm

src/containers/shared/components/TokenTableRow.tsx
```diff
@@ -0,0 +1,21 @@
...
+  shouldDisplay?: boolean
+  label: string
+  value: any
+}
+
+export const TokenTableRow = (props: TokenTableRowProps) => {
+  const { shouldDisplay = true, label, value } = props
+  return (
+    <>
+      {shouldDisplay && (
```
mvadari, line 13: Why not just make this
```suggestion
      {value && (
```

shawnxie999: Yeah good point, I thought some of these conditions were independent of the value but there wasnt. Will change it

shawnxie999: oh the `is_burned` field is a boolean, but I cant print that in, so I had to pass in a  string

ckniffen (approved):

mvadari (approved):

PR #455, opened 2022-11-15
fix: remove empty validators on Validators page

pdp2121:

There are a few empty validators that has no data. Need to investigate where they come from and remove them.

Cause: the `validations` stream from `Streams` component returns some validations with `pubkey` not in the validators key list.
Solution: only update the validators list (`vList`) and validation list (`validations`) with valid validation's `pubkey`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Before
Explorer Validators page has empty validators and inconsistent validators count.
Screenshot 2022-11-08 at 4 02 19 PM

### After
Those empty validators are not added to the table and validators count stay consistent.

Screenshot 2022-11-15 at 2 18 14 PM

### DIFF
--- a/src/containers/shared/components/Streams.jsx
+++ b/src/containers/shared/components/Streams.jsx
@@ -295,9 +295,11 @@ class Streams extends Component {
         unl: vList && vList[data.pubkey] && vList[data.pubkey].unl,
       })
 
-      validators[data.pubkey] = data
-      this.updateLedgers(ledgers)
-      this.updateValidators(validators)
+      if (vList[data.pubkey]) {
+        validators[data.pubkey] = data
+        this.updateLedgers(ledgers)
+        this.updateValidators(validators)
+      }
       return { ledgers, validators, maxLedger }
     })
   }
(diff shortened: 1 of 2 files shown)


src/containers/shared/components/Streams.jsx
```diff
@@ -295,9 +295,11 @@ class Streams extends Component {
         unl: vList && vList[data.pubkey] && vList[data.pubkey].unl,
       })
 
-      validators[data.pubkey] = data
-      this.updateLedgers(ledgers)
-      this.updateValidators(validators)
+      if (vList[data.pubkey]) {
```
mvadari, line 298:
```suggestion
      if (vList[data.pubkey] != null) {
```

ckniffen:
We should have enough information to show the pubkey and server_version from the validation stream.

pdp2121:
> We should have enough information to show the pubkey and server_version from the validation stream.

@ckniffen could you clarify what you meant?

ckniffen:
> @ckniffen could you clarify what you meant? Do you mean we just remove those validators from the validators table but since show their validations?

Instead of hiding them I was suggesting that we show them still but make sure we show the key. Turns out I was wrong about the version.  That is only only on every 256th ledger. 

I would think it would render something like this. 

Screen Shot 2022-11-17 at 4 03 15 PM

ckniffen:
Closed in favor of #467

PR #460, opened 2022-11-17
fix: white flicker on Network Validators page

pdp2121:

The `Validators` page of the Explorer is flickering white due to the VHS gets updated slower than the streams. This PR will fix the issue by prioritize the `ledger_hash` and `ledger_index` to use those of the stream and keep the index uniformly increasing. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

After the change, the `ledger` column in Validators table will increase uniformly and there will be no white flicker.


src/containers/Network/Validators.tsx
```diff
@@ -34,7 +46,7 @@ export const Validators = () => {
           newValidatorList[v.signing_key] = v
```
ckniffen, line 46: can this code be incorporated into the other loop so we dont have to iterate twice?

ckniffen:
I wonder if it makes sense for the VHS to store ledger_hash in addition to ledger_index so that we can show the ledger colors upon page load.

ckniffen:
Closed in favor of #467

Issue #452, opened 2022-11-09
Undefined values in transaction raw JSON visualization


zgrguric:
https://livenet.xrpl.org/transactions/1C480B388ECADC1AED30161F9F50B72B5A5A86894480B3B2B2AFCD056EBF097F/raw

![Screenshot_1](https://user-images.githubusercontent.com/108837406/200799702-417c997e-0dba-4d03-b7c0-5fca96380517.jpg)

ckniffen:
Related to #358

ckniffen:
Closing as it is a duplicate of #358

Issue #358, opened 2022-09-23
Update "Raw" transaction tab to show real raw response


ckniffen:
The "Raw" tab shows a json view of transaction after we have performed some processing.  It should be the the result section of the `tx` websocket response.

https://github.com/ripple/explorer/blob/staging/src/rippled/transactions.js#L11-L15

PR #464, opened 2022-11-20
Undefined values in transaction raw JSON visualization solved #358 & …

bluntbrain:
…#452

##  Undefined values in transaction raw JSON visualization #452 & #358 Bug Fix
Places values from tx instead of undefined in src/rippled/lib/utils.js file.

### Bug Fix
This is a Bug fix, mentioned in Issue #358 & #452, showing correct values in place of undefined.
This I've verified in both mainnet and testnet

## Before 
before-fix

## After
after-fix


bluntbrain:
Kindly review the change and merge this PR
cc @ckniffen @mvadari @shawnxie999

shawnxie999:
Should we have a test case for this?

bluntbrain:
It's just mapping of data already present on the page, hardly 5 line change. Can you check once and merge this PR @shawnxie999 
cc @ckniffen @mvadari

ckniffen:
This is good as an intermediary step.  Eventually we want the data shown in the Raw tab to exactly reflect what rippled returned.

bluntbrain:
Noted. @ckniffen Kindly merge this PR, I'll setup the raw page to exactly reflect what rippled returned and raise a PR soon. Also let me know if anything new needs to be developed, I've some free time in hand and want to contribute to this repo :)

ckniffen:
I will try to find an push the draft I had created.  At first i thought it would be as simple as saving the data to raw earlier.  Unfortunately there were a lot of places that references `raw` to get values and there weren't sufficient tests in place to make me confident I did it all correctly.

ckniffen (approved):

shawnxie999 (approved):

ckniffen:
Tests are failing.  Part of this is much of the test data is mocks of what getTransaction produces.  Instead I would in the future like the tests to be raw output from rippled/clio that we then process.  This will require a fair amount of work with regards to the transaction specific tests as well as quite a few other places.

There are 17 places that reference properties on `raw`.  When making raw not be the result of `formatTransaction` only one assertion fails. :(

Luckily most of these these are contained to two components `Transaction` and `SimpleTab`.

ckniffen:
#479 will get us closer to having proper coverage of SimpleTab to start making more of these changes.

mvadari:
This fix creates a lot of repeated data - e.g. the metadata is now both under `tx.meta` and `meta`

ckniffen:
This will require a much more thorough update.

PR #469, opened 2022-11-22
refactor: convert `LedgersPage` to hooks and TypeScript

mvadari:

Title says it all.

### Context of Change

I was running into an issue with #470 that would be much easier to resolve if I could use `useQuery`

### Type of Change

### TypeScript/Hooks Update

## Before / After

No change

## Test Plan

CI passes.


src/containers/Ledgers/index.jsx
```diff
@@ -0,0 +1,99 @@
+import React, { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
```
ckniffen, line 2: Since this files is already being considered a new might as well make it typescript.

src/containers/Ledgers/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+  useQuery(['fetchValidatorData'], async () => fetchValidators(), {
+    refetchInterval: FETCH_INTERVAL_MILLIS,
+    refreshOnMount: true,
+  })
+
+  const updateSelected = (pubkey) => {
+    setSelected(selected === pubkey ? null : pubkey)
+  }
+
+  const updateLedgers = (newLedgers) => {
```
ckniffen, line 61: This could be removed and just directly use setLedgers.

ckniffen (approved):

shawnxie999 (approved):
looks good

PR #470, opened 2022-11-22
feat: add support for Networks tab for custom networks

mvadari:
## High Level Overview of Change

Title says it all - custom.xrpl.org will now support the Networks page.

### Context of Change

People will want to know the status of the network for custom networks.

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

### TypeScript/Hooks Update

- [ ] Updated files to React Hooks
- [ ] Updated files to TypeScript

## Test Plan

Modified tests as needed. I couldn't figure out how to test the NetworkContext generation, but it works locally.

PR #472, opened 2022-11-23
refactor: remove validator report API endpoint

mvadari:

This PR removes the `api/v1/validator_report` API endpoint and instead just directly calls the VHS. It also cleans up some non-crashing console errors.

### Context of Change

It's useful cleanup for #470 

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

No change.

## Test Plan

CI passes.


ckniffen (approved):

src/containers/Validators/index.js
```diff
@@ -67,9 +67,15 @@ class Validator extends Component {
   fetchData = () => {
     const { match } = this.props
     const { identifier = '' } = match.params
-    axios.get(`/api/v1/validator_report/${identifier}`).then((resp) => {
-      this.setState({ reports: resp.data })
-    })
+    axios
+      .get(`${process.env.REACT_APP_DATA_URL}/validator/${identifier}/reports`)
+      .then((resp) => resp.data.reports)
```
ckniffen, line 72: this is an extra then.

mvadari: Is there something wrong with doing it this way? I find it's a little cleaner to move some of the logic to a `then`.

pdp2121 (approved):

PR #473, opened 2022-11-23
refactor: convert `Validators` to hooks and TypeScript

mvadari:

Title says it all.

### Context of Change

I was running into an issue with #470 that would be much easier to resolve if I could use multiple contexts and didn't need to mess with redux actions

### Type of Change

### TypeScript/Hooks Update

## Before / After

No change

## Test Plan

CI passes.


src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+          analytics(ANALYTIC_TYPES.exception, {
+            exDescription: `${url} --- ${JSON.stringify(axiosError)}`,
+          })
+          setError(status)
+          setIsLoading(false)
+        })
+    }
+  }
+
+  useQuery(['fetchValidatorData'], async () => fetchValidatorData(), {
```
ckniffen, line 137: You need to add the validator key to the queryKey array so that it caches based on the master_key.

ckniffen, line 137: - `isLoading` is provided by `useQuery` so you don't have to manage the state of `loading` yourself.
- `data` is provided and is the resolved result of the promise.
- `error` is provided by `useQuery` as well by the promise returning a rejection

src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+}
+
+const Validator = (props: Props) => {
+  const [reports, setReports] = useState([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [error, setError] = useState(null)
+  const [data, setData] = useState<ValidatorData>({})
+  const [width, setWidth] = useState(0)
+  const ref = useRef<any>(null)
+  const { t, i18n } = useTranslation()
```
ckniffen, line 60: `useLanguage` hook will give you the language.  `i18n` is the i18n-next object and is not the language string.

src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+  SERVER_ERROR,
+} from '../shared/utils'
+import { getLedger } from '../../rippled'
+import SimpleTab from './SimpleTab'
+import { HistoryTab } from './HistoryTab'
+import './validator.scss'
+import SocketContext from '../shared/SocketContext'
+
+const ERROR_MESSAGES = {
+  [NOT_FOUND]: {
```
ckniffen, line 22: Where is `NOT_FOUND` set?

src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+    title: 'generic_error',
+    hints: ['not_your_fault'],
+  },
+}
+
+const getErrorMessage = (error: keyof typeof ERROR_MESSAGES | null) =>
+  (error && ERROR_MESSAGES[error]) || ERROR_MESSAGES.default
+
+interface Props {
+  match: {
```
ckniffen, line 36: You can use `useRouteMatch` and `useParams` to get these properties.

https://v5.reactrouter.com/web/api/Hooks/useroutematch
https://v5.reactrouter.com/web/api/Hooks/useparams

ckniffen (changes requested):

src/containers/Validators/index.tsx
mvadari: There's no `useLanguage` hook in `react-i18next`

ckniffen: Its in the explorer.  I wrote it to avoid having to access `i18n.resolvedLanguage`

src/containers/Validators/index.tsx
mvadari: It's in `shared/utils`

src/containers/Validators/index.tsx
mvadari: Wouldn't it just cache based on the identifier? I don't get how you'd cache based on data that it's receiving

mvadari: I couldn't get this to work for me - there were a bunch of errors

src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+            : SERVER_ERROR
+        analytics(ANALYTIC_TYPES.exception, {
+          exDescription: `${url} --- ${JSON.stringify(axiosError)}`,
+        })
+        setError(status)
+        setIsLoading(false)
+      })
+  }
+
+  useQuery(
```
ckniffen, line 139: Leverage `error`, `isFetching `, and `data` instead of managing state manually.

src/containers/Validators/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+        {renderSummary()}
+        {renderTabs()}
+        <div ref={ref} className="tab-body">
+          {body}
+        </div>
+      </>
+    )
+  }
+
+  const loader = isLoading ? <Loader className="show" /> : <Loader />
```
ckniffen, line 222: Don't render loader when no loading is happening.

pdp2121 (approved):

src/containers/Validators/types.ts
```diff
@@ -0,0 +1,10 @@
+export interface ValidatorReport {
```
ckniffen, line 1: nit: This can now extend ValidatorScore.

ckniffen (approved):

Issue #204, opened 2022-07-27
Refactor PaymentChannel transactions to use new transaction scheme


ckniffen:
Update PayChannelClaim, PayChannelCreate, and PayChannelFund transactions to use scheme defined in https://github.com/ripple/explorer/tree/staging/src/containers/shared/components/Transaction.

Also we need to update components to use hooks and tsx.

ckniffen:
Closed by #468, #471, and #475

PR #475, opened 2022-11-28
refactor(tx) Switch `PaymentChannelFund` to new transaction scheme

ckniffen:

Moves `PaymentChannelFund` to the new transaction schema, along with hooks, typescript, and more tests.

Now shows channel id even when the tx fails.

### Context of Change

Part of #204

### Type of Change

### TypeScript/Hooks Update

## Before
Screen Shot 2022-11-28 at 9 00 56 AM
Screen Shot 2022-11-28 at 9 04 50 AM
Screen Shot 2022-11-28 at 9 05 49 AM
Screen Shot 2022-11-28 at 9 06 25 AM

## After
Screen Shot 2022-11-28 at 9 02 26 AM
Screen Shot 2022-11-28 at 9 05 10 AM
Screen Shot 2022-11-28 at 9 06 02 AM
Screen Shot 2022-11-28 at 9 06 31 AM


src/containers/shared/components/Amount.tsx
```diff
@@ -22,7 +27,10 @@ export const Amount = ({ displayIssuer = true, value }: AmountProps) => {
 
   return (
     <span className="amount">
-      <span className="amount-localized">{localizedAmount}</span>{' '}
+      <span className="amount-localized">
+        {modifier && <span className="amount-modifier">{modifier}</span>}
```
shawnxie999, line 31: What kind of modifiers are there? just `+` and `-`?

src/containers/shared/components/Amount.tsx
```diff
@@ -8,9 +8,14 @@ import { ExplorerAmount } from '../types'
 export interface AmountProps {
   value: ExplorerAmount
   displayIssuer?: boolean
+  modifier?: string
```
shawnxie999, line 11: nit: should we add a comment about the usage of this prop? may not be obvious at first glance

src/containers/shared/components/Amount.tsx
ckniffen: So far I have only used `+`, but it could be `~` or `-`.  I suppose i could limit it to those 3.

src/containers/shared/components/Amount.tsx
ckniffen: Will do.

src/containers/shared/components/Amount.tsx
```diff
@@ -6,11 +6,16 @@ import Currency from './Currency'
 import { ExplorerAmount } from '../types'
 
 export interface AmountProps {
-  value: ExplorerAmount
   displayIssuer?: boolean
+  modifier?: `+` | '-' | '~' // value to put in front of the currency symbol and number
+  value: ExplorerAmount
```
JST5000, line 11: nit: I think it's often good practice to put the optional parameters at the end, rather than the middle.

JST5000 (approved):
LGTM

shawnxie999 (approved):

PR #479, opened 2022-11-28
refactor: Simplify tx and validator SimpleTab

ckniffen:

- Render the right content using SimpleRow components. This allows more code reuse between the mobile and tablet versions.
- Validator
	- Remove `Updated Date/Time (UTC)` from mobile view.  The value was removed from the desktop view April 2021
	- Change unl url to be white.
	- Remove commas from ledger index. Only place formatted that way
- Transaction
	- Increase account font-size to match other fields.

### Context of Change

This adds more assertions to help with #358 and #452. Was able to get rid of ~100 lines of code.

### Type of Change

### TypeScript/Hooks Update

## Before
Before Tx Desktop
Before TX Mobile
Before Validator Desktop
Before Validator Mobile

## After
After Tx Desktop
After Tx Mobile
After Validator Desktop
After Validator Mobile


JST5000:
For future PR's if there's a bunch of changes, could you pair them up? (Before + After for each page?) - I find it a bit hard to scroll to the right screen when the changes are stacked like this.

ckniffen:
> For future PR's if there's a bunch of changes, could you pair them up? (Before + After for each page?) - I find it a bit hard to scroll to the right screen when the changes are stacked like this.

Will do.

JST5000 (approved):
LGTM

src/containers/Transactions/SimpleTab.jsx
```diff
@@ -97,37 +72,24 @@ class SimpleTab extends Component {
...
       <div className="simple-body simple-body-tx">
         <div className="rows">
           <Simple type={raw.tx.TransactionType} data={data.summary} />
-          {rowIndex}
+          {width < BREAKPOINTS.landscape && rowIndex}
         </div>
-        {cartIndex}
+        {width >= BREAKPOINTS.landscape && (
+          <div className="index">{rowIndex}</div>
+        )}
```
mvadari, line 92: lol is that really the only difference?

mvadari:
Why is the UNL text changed to be white? Shouldn't it be a link to the UNL?

ckniffen:
> Why is the UNL text changed to be white? Shouldn't it be a link to the UNL?

It was not a link previously.  It just looked like one and used a color slightly different from the link color (`#292` vs `#80ffba`).  I can make it a link if you would like.

mvadari:
> It was not a link previously. It just looked like one and used a color slightly different from the link color (`#229922` vs `#80ffba`). I can make it a link if you would like.

I'm not sure it's necessary for it to be a link, given that a UNL manifest isn't exactly human-readable (hm, maybe we should add a UNL page to the Explorer). That's fine then.

mvadari (approved):

PR #489, opened 2022-12-01
refactor: remove `/api/v1/validators` endpoint

mvadari:

This PR removes the `/api/v1/validators` endpoint, and instead has the frontend directly call the VHS. All it was doing was caching, basically.

### Context of Change

Simplifying the Explorer codebase.

### Type of Change

### TypeScript/Hooks Update

## Before / After

N/A

## Test Plan

CI passes.

## Future Tasks

Update the VHS API to cache better/maybe even provide this exact API.


src/containers/Network/UpgradeStatus.tsx
```diff
@@ -13,49 +13,58 @@ import {
...
 import { useLanguage } from '../shared/hooks'
 import Log from '../shared/log'
+import { StreamValidator, ValidatorResponse } from './types'
 
-export const aggregateData = (validators: any[]) => {
+const ENV_NETWORK_MAP: Record<string, string> = {
+  mainnet: 'main',
+  testnet: 'test',
+  devnet: 'dev',
+  nft_sandbox: 'nft-dev',
```
pdp2121, line 22: Should add `amm-dev` as well?

src/containers/Network/Validators.tsx
```diff
@@ -23,41 +25,50 @@ export const Validators = () => {
...
+    const updated: Record<string, StreamValidator> = {}
     const keys = new Set(Object.keys(validators).concat(Object.keys(live)))
     keys.forEach((d: string) => {
-      updated[d] = validators[d] || live[d]
-      if (live[d] && live[d].ledger_index > updated[d].ledger_index) {
-        updated[d].ledger_index = live[d].ledger_index
-        updated[d].ledger_hash = live[d].ledger_hash
+      const newData: StreamValidator = validators[d] || live[d]
+      if (newData.ledger_index == null && live[d] && live[d].ledger_index) {
+        // VHS uses `current_index` instead of `ledger_index`
```
pdp2121, line 37: The VHS data is updated slower than the `Streams` data. So `ledger_index` might jump couple of index instead of increasing one by one as of now. Is that the behavior we want?

pdp2121:
Added couple comments. But the Validators and UpgradeStatus code looks much better with your changes 💯

src/containers/Network/Validators.tsx
mvadari: The VHS data should never override the Streams data if the Streams data exists - which I'm pretty sure is what happens here.

pdp2121: Yes you're right. I got confused between validators and live.

pdp2121 (approved):

src/containers/shared/components/Streams.jsx
```diff
@@ -68,9 +68,9 @@ const fetchLoadFee = (rippledSocket) =>
 
 const formatLedgers = (data) =>
   Object.entries(data)
-    .map((l) => {
-      const hashes = Object.entries(l[1].hashes || {}).map((h) => {
-        const validated = h[0] === l[1].ledger_hash
+    .map(([_index, ledger]) => {
```
ckniffen, line 71: Was this change due to _index not being used?

mvadari: Yes

src/containers/Network/types.ts
```diff
@@ -39,3 +39,52 @@ export interface NodeData extends NodeResponse {
...
+  revoked?: boolean
+  domain: string
+  chain: string
+  networks?: string
+  // eslint-disable-next-line camelcase -- from VHS
+  current_index: number
+  // eslint-disable-next-line camelcase -- from VHS
+  server_version?: string
+  // eslint-disable-next-line camelcase -- from VHS
+  agreement_1h: {
```
ckniffen, line 59: This could be its own type like 

```

export interface ValidatorScore {
    missed: number
    total: number
    score: string
    incomplete: boolean
}

ckniffen (approved):

PR #493, opened 2022-12-05
Fix accepter occasionally being incorrect 

will2022:

Occasionally the address sending the `NFTokenAcceptOffer` transaction won't be the one that buys/sells the NFT. This is fixed by using the `acceptedOfferNode.Destination` whenever it's not undefined.

### Context of Change

I was looking at https://livenet.xrpl.org/transactions/<some tx hash> and was wondering why the buyer wasn't the same as other explorers were showing.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Test Plan

Works locally.


ckniffen:
@will2022 Looks great.  Can you add a unit test for this scenario?

will2022:
I've looked a little more at the transaction that caused this bug. It doesn't look like this would be the final fix

PR #497, opened 2022-12-12
feat: AMM Accounts page

Amindv1:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## After

Screenshot 2022-12-14 at 10 38 28 AM


src/rippled/lib/rippled.js
```diff
@@ -426,6 +428,26 @@ const getOffers = (
...
+  const request = {
+    command: 'amm_info',
+    asset,
+    asset2,
+    ledger_index: 'validated',
+  }
+
+  return query(rippledSocket, request).then((resp) => {
+    if (resp.error_message || !resp.amm.validated) {
+      throw new Error('meep meep', 404)
```
ckniffen, line 444: Change the error message.

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -61,3 +68,7 @@ export const TransactionTableRow = ({ tx }: Props) => {
     </li>
   )
 }
+
+TransactionTableRow.defaultProps = {
```
ckniffen, line 72: If you export `Props` this isn't needed.

src/containers/shared/flags.jsx
```diff
@@ -0,0 +1,12 @@
+/*
+These are the supported flags for each api call. Please add them as necessary.
+ */
+
+const flags = {
```
ckniffen, line 5: `ACCOUNT_FLAGS` in `rippled/lib/utils.js` has the same info.

src/containers/shared/utils.js
```diff
@@ -438,3 +438,41 @@ export const fetchMetrics = () =>
...
+
+export const formatAsset = (asset) =>
+  typeof asset === 'string'
+    ? { currency: 'XRP' }
+    : {
+        currency: asset.currency,
+        issuer: asset.issuer,
+      }
+
+export const formatBalance = (asset) => {
```
ckniffen, line 450: This looks to be the same a `formatAmount in` `src/rippled/lib/txSummary/formatAmount.js`

src/containers/shared/utils.js
```diff
@@ -438,3 +438,41 @@ export const fetchMetrics = () =>
...
+  return typeof asset === 'string'
+    ? { currency: 'XRP', amount: Number(asset) / drop }
+    : {
+        currency: asset.currency,
+        amount: Number(asset.value),
+        issuer: asset.issuer,
+      }
+}
+
+export const localizeBalance = (balance, language) => {
```
ckniffen, line 461: Why are only those currencies to display the currency code?

src/containers/shared/components/Transaction/index.ts
```diff
@@ -1,3 +1,8 @@
+import { AMMCreate } from 'containers/shared/components/Transaction/AMM/AMMInstanceCreate'
```
ckniffen, line 1: This file does nothing in this PR.

ckniffen (commented):
First round of comments.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <Currency {...balance} showIssuer={false} />/
+          <Currency {...balance2} showIssuer={false} />
+        </div>
+      </div>
+      <div className="box-content">{renderHeaderContent()}</div>
+    </div>
+  )
+}
+
+export default AMMAccountHeader
```
ckniffen, line 87: Do not use default export.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+  ANALYTIC_TYPES,
+  formatAsset,
+  formatBalance,
+} from 'containers/shared/utils'
+import { Tabs } from 'containers/shared/components/Tabs'
+import { AccountAssetTab } from 'containers/Accounts/AccountAssetTab/AccountAssetTab'
+import { getAccountTransactions, getAMMInfo } from 'rippled/lib/rippled'
+import SocketContext from '../../../shared/SocketContext'
+import { ERROR_MESSAGES } from '../../Errors'
+import { getTokenPairData } from '../../../../apis/OnTheDex'
```
ckniffen, line 20: This file does not exist.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+const AMMAccounts = (props: any) => {
+  const { id: accountId, tab = 'transactions' } = useParams<{
+    id: string
+    tab: string
+  }>()
+  const { path = '/' } = useRouteMatch()
+  const { t } = useTranslation()
+  const mainPath = `${path.split('/:')[0]}/${accountId}`
+  const [currencySelected] = useState('XRP')
```
ckniffen, line 49: why is `currencySelected` using `useState` when it is never changed?

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { Tabs } from 'containers/shared/components/Tabs'
+import { AccountAssetTab } from 'containers/Accounts/AccountAssetTab/AccountAssetTab'
+import { getAccountTransactions, getAMMInfo } from 'rippled/lib/rippled'
+import SocketContext from '../../../shared/SocketContext'
+import { ERROR_MESSAGES } from '../../Errors'
+import { getTokenPairData } from '../../../../apis/OnTheDex'
+
+export interface AmmDataType {
+  balance: { currency: string; amount: number; issuer: string }
+  balance2: { currency: string; amount: number; issuer: string }
```
mvadari, line 24: Pretty sure these are already types somewhere (`IssuedCurrencyAmount` in `types.ts` I believe)

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+  useEffect(() => {
+    let asset1: { currency: string; issuer?: string }
+    let asset2: { currency: string; issuer?: string }
+    let volumeAsset1: number
+    let volumeAsset2: number
+    let assetXrpValue: number
+    let asset2XrpValue: number
+    let ammData: any
+    let balance: any
+    let balance2: any
```
mvadari, line 65: Why are these all `let` before? Seems unnecessary, they're never changed and only created/used once

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
Amindv1: You mean export the interface Props? Why would that remove the need for the default prop?

src/containers/shared/components/Transaction/index.ts
Amindv1: I'll remove it

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: It exists in the other PR: https://github.com/ripple/explorer/pull/498

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: I needed a way to pass the data down the promise chain and I didn't know how else to do it. We need to access the values from one network call/promise to another.

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
ckniffen: I don't know why but it makes eslint stop complaining about it.

src/containers/shared/flags.jsx
Amindv1: The account flags in utils dont work because theyre an int to string lookup whereas these flags are string to int so this file is still needed unless I'm missing something?

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
Amindv1: The purpose of the default prop here is actually to guarantee the default value of the field is false. Are you saying that even if this is removed the default value will still be false?

src/containers/shared/flags.jsx
ckniffen: I will take a look where it is used.

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
ckniffen: You can set the default in the parameters

src/containers/shared/utils.js
Amindv1: removed

src/containers/shared/utils.js
Amindv1: We actually dont display currency code for these currencies. The reasoning is that localizeNumber will add the icons for these main 3 currencies but not for others so in all other cases we include the standard shorthand for the currencies i.e. 'SOLO'. So for xrp it would look like'[xrp icon]400' instead of 'XRP [xrp icon]400' 

Let me know if you want me to remove it.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: The types differ slightly, IssuedCurrency doesnt have an amount field.

mvadari: https://github.com/ripple/explorer/blob/staging/src/containers/shared/types.ts#L12-L18
```
export type ExplorerAmount =
  | string
  | {
      issuer?: string
      currency: string
      amount: number
    }
```

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed it

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { useParams, useRouteMatch } from 'react-router'
+import { useTranslation } from 'react-i18next'
+import { connect } from 'react-redux'
+import { AMMAccountHeader } from 'containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader'
+import { AccountTransactionTable } from 'containers/Accounts/AccountTransactionTable/index'
+import NoMatch from 'containers/NoMatch'
+import 'containers/Accounts/styles.scss'
+import { analytics, ANALYTIC_TYPES, formatAsset } from 'containers/shared/utils'
+import { Tabs } from 'containers/shared/components/Tabs'
+import { AccountAssetTab } from 'containers/Accounts/AccountAssetTab/AccountAssetTab'
```
ckniffen, line 11: This line is causing the app to not run.

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -19,11 +20,12 @@ const DATE_OPTIONS = {
   timeZone: TIME_ZONE,
 }
 
-interface Props {
+export interface Props {
   tx: any
+  hasTokensColumn?: false
```
ckniffen, line 25: This should be `boolean` and the default value of `false` should be on line 28.

src/containers/Accounts/index.js
```diff
@@ -1,57 +1,24 @@
...
-ERROR_MESSAGES.default = {
-  title: 'generic_error',
-  hints: ['not_your_fault'],
-}
 
-const getErrorMessage = (error) =>
-  ERROR_MESSAGES[error] || ERROR_MESSAGES.default
+import { AccountAssetTab } from './AccountAssetTab/AccountAssetTab'
 
 const Accounts = (props) => {
```
ckniffen, line 15: `props` is no longer used

src/containers/Accounts/index.js
```diff
@@ -94,10 +58,6 @@ const Accounts = (props) => {
   )
 }
 
-Accounts.propTypes = {
-  error: PropTypes.number,
-}
-
 Accounts.defaultProps = {
   error: null,
 }
```
ckniffen, line 63: `connect` logic can be removed since error is no longer used.

src/containers/Accounts/AccountsRouter.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { id: accountId } = useParams<{ id: string }>()
+  const rippledSocket = useContext(SocketContext)
+  const [error, setError] = useState<any>()
+  const [comp, setComp] = useState<any>()
+  const flags: any = Object.entries(ACCOUNT_FLAGS).reduce(
+    (all, [key, value]) => ({ ...all, [value]: key }),
+    {},
+  )
+
+  useEffect(() => {
```
ckniffen, line 35: The way this is setup `getAccountInfo` is called twice.  Once here and then in the eventual component.

src/containers/shared/components/TxToken.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// We should consider moving this logic to each individual parser. This would give us more customizability.
+function getTokenPair(
+  type: string,
+  fee: number,
+  amount: { currency: string; amount: number },
+  amount2: { currency: string; amount: number },
+) {
+  if (type === 'AMMBid' || type === 'AMMVote') {
```
ckniffen, line 15: Could it be an LP token if it was part of a payment between two other parties and the tx is returned in `account_tx`?

ckniffen (changes requested):

package.json
```diff
@@ -51,6 +51,7 @@
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.19.1",
+    "@testing-library/react": "^12.1.2",
```
ckniffen, line 54: This is not used anywhere.

package.json
```diff
@@ -177,11 +180,24 @@
...
     "env": {
       "browser": true,
+      "node": true,
       "jest": true,
       "es6": true
     },
     "rules": {
-      "prettier/prettier": "error",
+      "import/no-unresolved": 0,
+      "import/no-extraneous-dependencies": [
```
ckniffen, line 189: Where was this erroring?

src/containers/Accounts/AMM/AMMAccounts/index.tsx
ckniffen: I think case the string version isn't allowed.  That portion could be cleaned up though so you could the type with issuer not required and then another type leveraging `Required<Type>`

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <AMMAccountHeader {...data!} />
+          <Tabs tabs={tabs} selected={tab} path={mainPath} />
+          {tab === 'transactions' && <AccountTransactionTable {...txProps} />}
+        </>
+      )}
+    </div>
+  )
+}
+
+export default connect(() => ({}))(AMMAccounts)
```
ckniffen, line 139: `connect` is not needed.

src/containers/shared/components/Currency.tsx
```diff
@@ -8,16 +8,18 @@ export interface Props {
   issuer?: string
   currency: string
   link?: boolean
+  showIssuer?: boolean
 }
 
 const Currency = (props: Props) => {
-  const { issuer, currency, link = true } = props
```
ckniffen, line 14: Can you add a test for this change?

tsconfig.json
```diff
@@ -15,7 +16,7 @@
     "checkJs": false,
     "forceConsistentCasingInFileNames": true,
     "esModuleInterop": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
ckniffen, line 19: Revert this change.  Running the app causes it to change anyway.  IntelliJ for some reason changes it when running tests.

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
```diff
@@ -0,0 +1,125 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import { I18nextProvider } from 'react-i18next'
+import thunk from 'redux-thunk'
+import { Provider } from 'react-redux'
+import { MemoryRouter, Route } from 'react-router'
+import AccountTransactionTable from 'containers/Accounts/AccountTransactionTable/AccountTransactionTable'
```
ckniffen, line 7: This needs to be an named import.

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { initialState } from 'rootReducer'
+import i18n from 'i18nTestConfig'
+import AMMAccountHeader from 'containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader'
```
ckniffen, line 10: This also needs to be an named import.

tsconfig.json
Amindv1: yea everytime i change it back it changes back to this, ill revert it

src/rippled/lib/rippled.js
```diff
@@ -426,6 +428,26 @@ const getOffers = (
...
+const getAMMInfo = (rippledSocket, asset, asset2) => {
+  const request = {
+    command: 'amm_info',
+    asset,
+    asset2,
+    ledger_index: 'validated',
+  }
+
+  return query(rippledSocket, request).then((resp) => {
+    if (resp.error_message || !resp.amm.validated) {
```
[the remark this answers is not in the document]
Amindv1: Yes it would. It should be the same regardless.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed

src/containers/Accounts/index.js
Amindv1: removed

src/containers/Accounts/index.js
Amindv1: removed

package.json
Amindv1: These are used to avoid import errors when importing directly from 'src/containers': 

Screenshot 2022-12-19 at 11 04 15 AM

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
Amindv1: changed it

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
Amindv1: changed it

src/containers/shared/components/TxToken.tsx
ckniffen: yeah i am talking about the potental for someone to create offers for LPTokens that dont interact with the AMM directly.  Those transactions should be returned in account_tx.  What should we show for those?

src/containers/shared/components/Currency.tsx
Amindv1: done

src/containers/Accounts/AccountsRouter.tsx
Amindv1: That is true.

src/containers/shared/components/TxToken.tsx
Amindv1: do you have an example transaction? what's the transaction type? 

this might be more of a product question.

package.json
ckniffen: Did you try using this? https://www.npmjs.com/package/eslint-import-resolver-typescript

package.json
```diff
@@ -204,6 +219,7 @@
       ],
       "no-bitwise": "off",
       "react/jsx-props-no-spreading": "warn",
+      "react/require-default-props": "off",
```
ckniffen, line 222: Why was this disabled?

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { useTranslation } from 'react-i18next'
+import 'containers/shared/css/nested-menu.scss'
+import 'containers/Accounts/AccountHeader/styles.scss'
+import 'containers/Accounts/AccountHeader/balance-selector.scss'
+import { AmmDataType } from '../index'
+import { localizeNumber, localizeBalance } from '../../../../shared/utils'
+import Currency from '../../../../shared/components/Currency'
+
+export const AMMAccountHeader = (props: AmmDataType) => {
+  const {
```
ckniffen, line 11: You can spread these in the parameter.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  useEffect(() => {
+    let asset1: { currency: string; issuer?: string }
+    let asset2: { currency: string; issuer?: string }
+    let ammData: any
+
+    /*
+    Get the first account transaction which in this case should be AMMInstanceCreate. From this we get
+    the two assets in the asset pool.
+    */
+    getAccountTransactions(rippledSocket, accountId, 1, undefined, true)
```
ckniffen, line 50: useQuery is the preferred way to do this now.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [error, setError] = useState<any>()
+  const { language } = props
+
+  useEffect(() => {
+    let asset1: { currency: string; issuer?: string }
+    let asset2: { currency: string; issuer?: string }
+    let ammData: any
+
+    /*
+    Get the first account transaction which in this case should be AMMInstanceCreate. From this we get
```
khancode, line 47:
```suggestion
    Get the first account transaction which in this case should be AMMCreate. From this we get
```
Transaction name has been renamed to `AMMCreate`

src/containers/Accounts/AMM/AMMAccounts/index.tsx
ckniffen: is this data not available in amm_info?

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+    the two assets in the asset pool.
+    */
+    getAccountTransactions(rippledSocket, accountId, 1, undefined, true)
+      .then((tData) => {
+        const { tx } = tData.transactions[0]
+        asset1 = formatAsset(tx.Amount)
+        asset2 = formatAsset(tx.Amount2)
+
+        // if one of the assets is XRP, make sure it's the second one
+        if (asset1.currency === 'XRP') {
```
ckniffen, line 57: Is this a bug?

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  }
+  return error ? (
+    renderError()
+  ) : (
+    <div className="accounts-page section">
+      {data && (
+        <>
+          <AMMAccountHeader {...data!} />
+          <Tabs tabs={tabs} selected={tab} path={mainPath} />
+          {tab === 'transactions' && <AccountTransactionTable {...txProps} />}
```
ckniffen, line 130: Spread operator is not desired here.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      window.scrollTo(0, 0)
+    }
+  })
+
+  document.title = `${t('xrpl_explorer')} | ${accountId.substr(0, 12)}...`
+
+  const tabs = ['transactions']
+  const txProps = {
+    accountId,
+    currencySelected: 'XRP',
```
ckniffen, line 119: Why is the selected currency XRP?  How does this work if neither assets are for XRP?

src/rippled/lib/rippled.js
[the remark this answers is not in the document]
ckniffen: Turns out the websocket client doesnt throw exceptions.

src/containers/shared/components/TxToken.tsx
Amindv1: return LP by default

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: no way around this right now

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: nope its intended - onthedex requires a certain format

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed it - it was a filler

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed it

package.json
Amindv1: removed it - it was being used for pathing but is unnecessary now.

package.json
mvadari: We use relative URLs, which doesn't have this issue, so for consistency it's probably better to do that.

Does `src/containers/...` work?

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { AmmDataType } from '../../index'
+
+const configureMockStore = require('redux-mock-store').default
+
+describe('AMM Account Header', () => {
+  const TEST_ACCOUNT_ID = 'rTEST_ACCOUNT'
+  const middlewares = [thunk]
+  const mockStore = configureMockStore(middlewares)
+
+  const creatWrapper = (state: AmmDataType) => {
```
mvadari, line 19: nit:
```suggestion
  const createWrapper = (state: AmmDataType) => {
```

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+    maximumFractionDigits: 2,
+  })
+  const tf = localizeNumber(tradingFee / tradingFeeTotal, language, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 3,
+  })
+
+  function renderHeaderContent() {
+    return (
+      <div className="section header-container">
```
mvadari, line 35: @ckniffen as a future task, maybe the page headers should all be unified.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return (
+    <div className="box account-header">
+      <div className="section box-header">
+        <div className="amm-title">AMMID</div>
+        <h2 className="amm">{ammId}</h2>
+        <div className="currency-pair">
+          <Currency {...balance} showIssuer={false} />/
+          <Currency {...balance2} showIssuer={false} />
```
mvadari, line 69: Why don't you want the issuer to be shown?

tsconfig.json
mvadari: You need to set the env var `DISABLE_NEW_JSX_TRANSFORM=true` - our npm scripts do this for you.

ckniffen: Ah.  I will try setting that in the IDE.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
ckniffen: That is in the plans.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: its long and clutters the screen

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
Amindv1: done

package.json
Amindv1: no src/containers doesn't work.

Amindv1: I disagree with this line of reasoning - i think this kind of importing is better since it allows us to move files around more easily but we need to get this merged so I'm going to make this change.

mvadari: My biggest worry is that since eslint doesn't support this, IDEs won't, and you can't use the Ctrl-Click feature to go to files. I find this _incredibly_ useful.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const b2 = localizeBalance(balance2, language)
+  const tradingFeeTotal = 1000
+  const lp = localizeNumber(lpBalance, language, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 2,
+  })
+  const tf = localizeNumber(tradingFee / tradingFeeTotal, language, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 3,
+  })
```
mvadari, line 32: Can these variables be named more descriptively?

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
mvadari: This seems like a potential safety issue - people might get confused about what AMM they're looking at, and could potentially be scammed as a result if they think they're looking at Gatehub's USD and are instead looking at Scammer1's USD.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { mount } from 'enzyme'
+import { I18nextProvider } from 'react-i18next'
+import thunk from 'redux-thunk'
+import { Provider } from 'react-redux'
+import { MemoryRouter } from 'react-router'
+import { initialState } from 'rootReducer'
+import i18n from 'i18nTestConfig'
+import { AMMAccountHeader, AmmDataType } from '../AMMAccountHeader'
+
+const configureMockStore = require('redux-mock-store').default
```
mvadari, line 11: Why is there a `require` used here? It looks like `configureMockStore` is usually directly imported from `'redux-mock-store'`

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+import {
+  AMMAccountHeader,
+  AmmDataType,
+} from './AMMAccountHeader/AMMAccountHeader'
+import { AccountTransactionTable } from '../../AccountTransactionTable'
+
+const getErrorMessage = (error: string) =>
+  ERROR_MESSAGES[error] || ERROR_MESSAGES.default
+
+export const AMMAccounts = (props: any) => {
```
mvadari, line 24: Why is this not typed?

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    id: string
+    tab: string
+  }>()
+  const { path = '/' } = useRouteMatch()
+  const { t } = useTranslation()
+  const mainPath = `${path.split('/:')[0]}/${accountId}`
+  const rippledSocket = useContext(SocketContext)
+  const [data, setData] = useState<AmmDataType>()
+  const [error, setError] = useState<any>()
+  const { language } = props
```
mvadari, line 35: This should use the `useLanguage` hook

src/containers/Accounts/AMM/AMMAccounts/index.tsx
mvadari: bumping the `useQuery` comment

src/containers/Accounts/AMM/AMMAccounts/index.tsx
mvadari: We're not using onthedex right now, though

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
mvadari: I also just think it could be confusing because someone might set up a Gatehub USD/Bitstamp USD AMM for arbitrage

src/containers/Accounts/AccountTransactionTable/test/AccountTransactionTable.test.js
```diff
@@ -28,6 +28,7 @@ describe('AccountTransactionsTable container', () => {
         () => {},
       ),
     currencySelected = null,
+    state = { hasToken: false },
```
mvadari, line 31: may as well put these both in some sort of `opts` param

src/containers/shared/MetaParser.tsx
```diff
@@ -0,0 +1,111 @@
+import { XRP_BASE } from './transactionUtils'
```
mvadari, line 1: nit: this file should be `metadataParser`(lowercase because it isn't a React component).

src/containers/Accounts/AMM/AMMAccounts/index.tsx
mvadari: also wow, this is horribly inconvenient and we should come up with a better way to do this.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  useEffect(() => {
+    let asset1: { currency: string; issuer?: string }
+    let asset2: { currency: string; issuer?: string }
+    let ammData: any
+
+    /*
+    Get the first account transaction which in this case should be AMMCreate. From this we get
+    the two assets in the asset pool.
+    */
+    getAccountTransactions(rippledSocket, accountId, 1, undefined, true)
```
mvadari, line 46: I'm not convinced that this will actually get you the first transaction for the account when there are multiple pages of transactions.

From the docs on the `forward` param:
```
If set to true, returns values indexed with the oldest ledger first. Otherwise, the results are indexed with the newest ledger first. (Each page of results may not be internally ordered, but the pages are overall ordered.)
```

mvadari: Also, this should really be added to `amm_info` - a way to search by account ID or object ID.

src/containers/shared/components/TransactionTable/TransactionTable.tsx
```diff
@@ -36,6 +42,7 @@ export const TransactionTable: TransactionTableComponent = ({
     <>
       <ol className="transaction-table">
         <li className="transaction-li transaction-li-header">
+          {hasTokensColumn && <div className="col-token"> {t('token')} </div>}
```
mvadari, line 45: I really don't like the idea of leaving an empty Tokens column. Can this part be moved to a different PR? You can just comment out this code if you don't want to remove the other parts, I just don't want it visible.

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+describe('AMM Account Page', () => {
+  const TEST_ACCOUNT_ID = 'rTEST_ACCOUNT'
+  const middlewares = [thunk]
+  const mockStore = configureMockStore(middlewares)
+  const accountTransactions: any = {
+    transactions: [
+      {
+        tx: {
+          Amount: '10000000000',
+          Amount2: { currency: 'USD', amount: '100000', issuer: 'your mom' },
```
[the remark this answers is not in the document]
Amindv1: wow so sorry i totally forgot to change this haha. Will do

src/containers/Accounts/AMM/AMMAccounts/index.tsx
ckniffen: Great point on the full history

ckniffen (approved):

src/containers/shared/components/TransactionTable/TransactionTable.tsx
Amindv1: The column has an option to be hidden - I just hid it and I'll unhide it in the next PR.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: yea i agree on the second point. it's actually something I brought up at the time but it didn't make sense to increase the ledger size for this specific use case. 

Good find on the forward flag - It seems like it's still guaranteed to be in the first page. I'll add a search function that iterates through all the transactions within the page results and finds AMMCreate.

src/containers/shared/MetaParser.tsx
Amindv1: will do

src/containers/Accounts/AMM/AMMAccounts/index.tsx
ckniffen: This method though will still be an issue if the node is not full history.  While this is not an issue for the fully supported devnets this will be an issue for custom and sidechains.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: was only used for language which has been removed. removing this.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
Amindv1: This is a remnant of a altered code. Removing it now.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: Sure

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: added first 4 chars.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: done

package.json
ckniffen: @Amindv1 lets remove the absolute paths for now.

src/containers/shared/components/TransactionTable/styles.scss
```diff
@@ -37,7 +38,15 @@
     @include semibold;
 
     @include for-size(desktop-up) {
-      width: 260px;
```
ckniffen, line 40: The widths should be based on whether or not there is the presence of the new column

src/containers/shared/components/Currency.tsx
```diff
@@ -8,16 +8,20 @@ export interface Props {
...
+  const { issuer, currency, link = true, showIssuer = true } = props
```
ckniffen, line 15: rename this to something like shorten or abbreviate.  Since we are still showing part of the issuer.

package.json
```diff
@@ -63,7 +63,7 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "canvas": "^2.11.0",
-    "eslint": "^7.31.0",
+    "eslint": "^7.32.0",
```
mvadari, line 66: Why is this being updated?

package.json
```diff
@@ -93,7 +94,10 @@
     "stylelint-config-standard": "^26.0.0",
     "stylelint-prettier": "^2.0.0",
     "stylelint-scss": "^4.3.0",
-    "typescript": "^4.3.5"
+    "typescript": "^4.9.4"
+  },
+  "resolutions": {
+    "react-error-overlay": "6.0.9"
```
mvadari, line 100: What does this do?

package.json
```diff
@@ -93,7 +94,10 @@
     "stylelint-config-standard": "^26.0.0",
     "stylelint-prettier": "^2.0.0",
     "stylelint-scss": "^4.3.0",
-    "typescript": "^4.3.5"
+    "typescript": "^4.9.4"
```
mvadari, line 97: Why is this being updated?

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
```
mvadari, line 1: This is in a nested folder `Accounts/AMM/AMMAccounts` - is there anything else in the `Accounts/AMM` folder? If not, then this could be moved to `Accounts/AMMAccounts`

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const wrapper = createWrapper(state)
+    await flushPromises()
+    wrapper.update()
+    expect(wrapper.find(AMMAccountHeader).length).toBe(1)
+    expect(wrapper.find('.amm-title').length).toBe(1)
+    expect(wrapper.find('.currency-pair').length).toBe(1)
+    expect(wrapper.text().includes('500')).toBe(true)
+    expect(wrapper.text().includes('%0.01')).toBe(true)
+    expect(wrapper.text().includes('XRP/USD')).toBe(true)
+    expect(wrapper.text().includes('1,000')).toBe(true)
```
mvadari, line 50: This should use the actual hex for the symbol instead of just ``

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        setData(ammInfo)
+      })
+      .catch((e) => {
+        analytics(ANALYTIC_TYPES.exception, {
+          exDescription: `Error setting up amm account --- ${JSON.stringify(
+            e,
+          )}`,
+        })
+
+        if (e.message !== 'onthedex failure') setError(e)
```
mvadari, line 97: There's no onthedex integration right now

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const tabs = ['transactions']
+
+  return error ? (
+    renderError()
+  ) : (
+    <div className="accounts-page section">
+      {data && (
+        <>
+          <AMMAccountHeader {...data!} />
```
mvadari, line 128: Why is this using a spread instead of `data={data}`? Our linter doesn't like spreads

src/containers/Accounts/AccountNFTTable/AccountNFTTable.tsx
```diff
@@ -28,6 +28,7 @@ export const AccountNFTTable = ({ accountId }: AccountNFTTableProps) => {
       getAccountNFTs(rippledSocket, accountId, pageParam).catch(
         (errorResponse) => {
           const errorLocation = `account NFTs ${accountId} at ${pageParam}`
+          // @ts-ignore
```
mvadari, line 31:
```suggestion
          // @ts-expect-error
```
So that it's easier to fix later, when we do have TS support for `analytics`

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
```diff
@@ -36,6 +38,7 @@ export const AccountTransactionTable = ({
         rippledSocket,
       ).catch((errorResponse) => {
         const errorLocation = `account transactions ${accountId} at ${pageParam}`
+        // @ts-ignore
```
mvadari, line 41: ditto

src/containers/Accounts/AccountTransactionTable/test/AccountTransactionTable.test.js
mvadari:
```suggestion
  const createWrapper = (
    getAccountTransactionsImpl = () =>
      new Promise(
        () => {},
        () => {},
      ),
      opts = {
        currencySelected=null,
        hasToken=false
      }
```

src/containers/Accounts/AccountsRouter.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <NoMatch title={message.title} hints={message.hints} />
+    </div>
+  )
+}
+
+const AccountsRouter = () => {
+  const { id: accountId } = useParams<{ id: string }>()
+  const rippledSocket = useContext(SocketContext)
+  const [error, setError] = useState<any>()
+  const [comp, setComp] = useState<any>()
```
mvadari, line 29: These should be typed better

src/containers/shared/components/Currency.tsx
mvadari: `truncate` perhaps?

src/rippled/lib/rippled.js
```diff
@@ -248,6 +248,7 @@ const getAccountTransactions = (
   account,
   limit = 20,
   marker = '',
+  reverseOrder = false,
```
mvadari, line 251:
```suggestion
  opts = {
    limit = 20,
    marker = '',
    reverseOrder = false
  }
```
I don't think that's quite legal JS but you get the idea - it'll simplify the number of random `undefined` params we need to add.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: fixed

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: removed

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
Amindv1: done

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: yea the folder has other contents.

src/containers/Accounts/AccountNFTTable/AccountNFTTable.tsx
Amindv1: i actually fixed the issue - must have forgotten to remove the ignore. removing this

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
Amindv1: removed

Amindv1: removed

src/containers/Accounts/AccountsRouter.tsx
Amindv1: It won't let me type these - results in an error: Argument of type 'Element' is not assignable to parameter of type 'SetStateAction '

src/rippled/lib/rippled.js
Amindv1: I see what youre saying. Right now it's just one parameter in one place so I don't think it matters. In the future we can simplify if we find that we're adding more parameters.

package.json
Amindv1: not sure why this updated but it should be fine. It's a minor version update which wont include breaking changes.

src/containers/shared/components/Currency.tsx
Amindv1: done

package.json
Amindv1: same as above

src/containers/shared/components/TransactionTable/styles.scss
Amindv1: I see what youre saying but 230px doesnt cut the types, and the code is pretty ugly with logic that assigns the css based off the existence of this column.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+        return getAMMInfo(rippledSocket, asset1, asset2)
+      })
+
+      /*
+      Use the assets to get the AMM Info.
+      */
+      .then((ammDataWrapper) => {
+        ammData = ammDataWrapper.amm
+        const balance = formatBalance(ammData.Amount)
```
ckniffen, line 76: The latest devnet update changed these properties.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useTranslation } from 'react-i18next'
+import '../../../../shared/css/nested-menu.scss'
+import '../../../AccountHeader/styles.scss'
+import '../../../AccountHeader/balance-selector.scss'
+import { localizeBalance, localizeNumber } from '../../../../shared/utils'
+import Currency from '../../../../shared/components/Currency'
+
+export interface AmmDataType {
+  balance: { currency: string; amount: number; issuer: string }
```
mvadari, line 10: Isn't this already a type somewhere? called Amount or something?

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return (
+    <div className="box account-header">
+      <div className="section box-header">
+        <div className="amm-title">Account ID</div>
+        <h2 className="amm">{accountId}</h2>
+        <div className="currency-pair">
+          <Currency {...balance} shortenIssuer />/
+          <Currency {...balance2} shortenIssuer />
```
mvadari, line 70: Don't use spreads if you can help it - our linter doesn't like it, and it's more descriptive to name each variable

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  it('renders AMM account page', async () => {
+    const state: AmmDataType = {
+      ...initialState,
+      balance: { currency: 'XRP', amount: 1000, issuer: 'hi' },
+      balance2: { currency: 'USD', amount: 9000, issuer: 'hi' },
+      lpBalance: 500,
+      tradingFee: 10,
+      accountId: 'the account ID',
```
mvadari, line 37: super nit: please use test strings for account IDs that are of the form `r[^ ]`

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <AMMAccountHeader data={state} />
+          </MemoryRouter>
+        </Provider>
+      </I18nextProvider>,
+    )
+  }
+
+  it('renders AMM account page', async () => {
+    const state: AmmDataType = {
+      ...initialState,
```
mvadari, line 32: This doesn't need to be there - it's already added in line 18

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }>()
+  const { path = '/' } = useRouteMatch()
+  const { t } = useTranslation()
+  const mainPath = `${path.split('/:')[0]}/${accountId}`
+  const rippledSocket = useContext(SocketContext)
+  const [data, setData] = useState<AmmDataType>()
+  const [error, setError] = useState<any>()
+  const language = useLanguage()
+
+  useEffect(() => {
```
mvadari, line 46: This should use `useQuery`

src/containers/Accounts/AMM/AMMAccounts/index.tsx
mvadari: Does it need to increase the ledger size if it's just an alternate form of request for `amm_info`?

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    await flushPromises()
+    wrapper.update()
+    expect(wrapper.find(AMMAccountHeader).length).toBe(0)
+    expect(wrapper.find(AccountTransactionTable).length).toBe(0)
+    expect(wrapper.find(NoMatch).length).toBe(1)
+    wrapper.unmount()
+  })
+})
+
+function flushPromises() {
```
mvadari, line 107: nit: these helper functions should go above the tests in this file

mvadari, line 107: Doesn't this function need to be async?

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
```diff
@@ -10,12 +10,14 @@ import { ANALYTIC_TYPES, analytics } from '../../shared/utils'
...
-  currencySelected: string
+  currencySelected?: string
+  hasTokensColumn: boolean
 }
 
 export const AccountTransactionTable = ({
   accountId,
-  currencySelected,
+  currencySelected = 'XRP',
+  hasTokensColumn,
```
mvadari, line 20: It's bad practice to have an optional param before a required param.

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
```diff
@@ -10,12 +10,14 @@ import { ANALYTIC_TYPES, analytics } from '../../shared/utils'
 
 export interface AccountTransactionsTableProps {
   accountId: string
-  currencySelected: string
+  currencySelected?: string
```
mvadari, line 13:
```suggestion
  currencySelected: string
```
The prop has a default value, so it's not optional

src/containers/Accounts/AccountsRouter.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const flags: any = Object.entries(ACCOUNT_FLAGS).reduce(
+    (all, [key, value]) => ({ ...all, [value]: key }),
+    {},
+  )
+
+  useEffect(() => {
```
mvadari, line 35: This should use `useQuery`

src/containers/Accounts/index.jsx
```diff
@@ -94,14 +54,6 @@ const Accounts = (props) => {
   )
 }
 
-Accounts.propTypes = {
-  error: PropTypes.number,
-}
-
 Accounts.defaultProps = {
   error: null,
 }
```
mvadari, line 59: Isn't this also no longer needed?

src/containers/shared/components/TxToken.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import '../css/txlabel.scss'
+
+interface Props {
+  tx: any
+}
+
+// We should consider moving this logic to each individual parser. This would give us more customizability.
```
mvadari, line 8:
```suggestion
// TODO: We should consider moving this logic to each individual parser. This would give us more customizability.
```

src/containers/shared/utils.js
```diff
@@ -330,8 +330,8 @@ export const formatLargeNumber = (d = 0, digits = 4) => {
 }
 
 // Document: https://developers.google.com/analytics/devguides/collection/analyticsjs/
-export const analytics = (type = null, fields = {}) => {
-  // Check if GoogleAnalytics is set, type and fields are not empty, type is valid
+export const analytics = (type, fields = {}) => {
```
mvadari, line 333: I believe this will not work properly without the `type = null`. You can just add `@ts-expect-error` where needed.

src/containers/shared/utils.js
mvadari: Doesn't look like this was removed

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
ckniffen: I believe it errored without it.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: no, we already had this discussion.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: currency would have to be changed to not use a spread here which would require me to go through and change every location we use currency which is out of scope for this project.

ckniffen: You can just pass `<Currency currency={balance2.currency} issuer={balance2.issuer}   shortenIssuer />`

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
Amindv1: done

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/test/AMMAccountHeader.test.tsx
Amindv1: removed

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
Amindv1: done

src/containers/Accounts/index.jsx
Amindv1: removed

src/containers/shared/components/TxToken.tsx
Amindv1: done

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: updated.

FYI: By using useQuery we lose dependency updating functionality. I.e. when language changes (one of the dependencies of this hook) we wont re-render the page with the new language. Updates happen when the cached data becomes stale (i think the default value is 1minute). 

Maybe im mistaken but I cant find documentation that says we can pass in a dependency list.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: You'd have to talk to Greg I can't comment on that.

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: Update: It seems like useQuery uses the query keys like useEffect dependencies: https://codemachine.dev/things-i-learned-while-using-react-query-part-2#treat-query-keys-like-a-dependency-array-useeffect-style

So the above FYI may not be relevant.

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
ckniffen: typescript infers it from the promise returned.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
Amindv1: updated.

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
Amindv1: done

src/containers/Accounts/AMM/AMMAccounts/index.tsx
Amindv1: updated

src/containers/Accounts/AccountsRouter.tsx
Amindv1: done

src/containers/shared/utils.js
Amindv1: everywhere we use this function we provide a type - there's no need for a default null value.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import '../../../../shared/css/nested-menu.scss'
+import '../../../AccountHeader/styles.scss'
+import '../../../AccountHeader/balance-selector.scss'
```
mvadari, line 5: @ckniffen do the CSS files need to be included here like this, or will inheritance take care of it?

Also, these should go below the JS imports

Amindv1: this is not a blocker - lets skip it

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const { path = '/' } = useRouteMatch()
+  const { t } = useTranslation()
+  const mainPath = `${path.split('/:')[0]}/${accountId}`
+  const rippledSocket = useContext(SocketContext)
+  const language = useLanguage()
+
+  const { data, error } = useQuery([accountId, language], () => {
+    let asset1: { currency: string; issuer?: string }
+    let asset2: { currency: string; issuer?: string }
+    let ammData: any
```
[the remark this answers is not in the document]
Amindv1: also not a blocker - lets skip

src/containers/Accounts/AMM/AMMAccounts/test/AMMAccounts.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  it('renders AMM account page when TVL not present', async () => {
+    const state = {
+      ...initialState,
+    }
+
+    setSpy(accountTransactions, ammInfo)
+
+    const wrapper = createWrapper(state)
+    await flushPromises()
```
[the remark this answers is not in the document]
Amindv1: nit + not a blocker - skipping

src/containers/Accounts/AccountTransactionTable/test/AccountTransactionTable.test.js
mvadari: bump

src/containers/App/test/App.test.js
```diff
@@ -114,15 +114,6 @@ describe('App container', () => {
     })
   })
 
-  it('renders account page', () => {
```
mvadari, line 117: Why was this test removed?

Amindv1: its a duplicate - already being tested in the accouts page tests

src/containers/shared/components/TransactionTable/TransactionTableRow.tsx
```diff
@@ -19,11 +20,12 @@ const DATE_OPTIONS = {
   timeZone: TIME_ZONE,
 }
 
-interface Props {
+export interface Props {
```
mvadari, line 23: Why is this exported?

Amindv1: it was erroring when it wasnt exported

src/containers/shared/components/test/Currency.test.tsx
```diff
@@ -16,6 +16,14 @@ describe('Currency', () => {
     wrapper.unmount()
   })
 
+  it('handles currency codes that are 4 characters and include issuer ', () => {
```
mvadari, line 19:
```suggestion
  it('handles currency codes that are shortened', () => {
```

Amindv1: nit - not a blocker lets skip

src/containers/shared/test/utils.test.js
```diff
@@ -205,3 +208,35 @@ describe('utils', () => {
...
+    expect(formatted).toEqual({ currency: 'USD', issuer: 'your mom' })
+  })
+})
+
+describe('AMM utils localize balance', () => {
+  it('formats XRP balance', () => {
+    const balance = { currency: 'XRP', amount: 9000000 }
+    const formatted = localizeBalance(balance, 'en-US')
+
+    expect(formatted).toEqual('9,000,000')
```
mvadari, line 233: The box should be replaced with the actual currency code unicode (`\uE900`)

src/containers/shared/utils.js
mvadari: bump

tsconfig.json
mvadari: This should be undone

public/locales/en-US/translations.json
```diff
@@ -394,6 +394,18 @@
...
   "fee_rate": "Fee Rate",
   "last_affecting_transaction": "Last affecting tx",
+  "Version": "Version",
   "increased_by": "increased by",
-  "Version": "Version"
+  "amm_lp_token_balance": "LP Token Balance",
+  "amm_token_balance": "Token Balance",
+  "trading_fee": "Trading Fee",
+  "amm_tvl": "TVL",
+  "amm_account_address": "Account Address",
```
mvadari, line 403: These keys don't need to be preceded with `amm_` - it makes it more general to not have that

Amindv1: nit & not a blocker - lets skip

mvadari: I'd rather handle these now, because we'll forget about it later and they're pretty quick to resolve

Amindv1: They are quick to resolve but theyve also existed since the first revision of the PR. This review has gone on for a long time and each iteration brings up more nits that have existed since the first revision. 

Theyre also not blocking - I think the best thing to do here is to get this change out for the user.  @ckniffen

src/containers/shared/test/utils.test.js
ckniffen: Lets still do this one.  We recently went through and corrected everywhere so dont want to take steps back.

src/containers/shared/components/test/Currency.test.tsx
mvadari: This one should be done - it'll be confusing later otherwise

src/containers/App/test/App.test.js
mvadari: Let's leave it in for now - this is testing routing.

src/containers/Accounts/AMM/AMMAccounts/AMMAccountHeader/AMMAccountHeader.tsx
ckniffen: All scss styles are included globally by the build importing them wont hurt anything especially good if the other place gets changed the styles wont disappears. 

We should look into an eslint rule to enforce.

tsconfig.json
ckniffen: Agreed

src/containers/shared/utils.js
ckniffen: Lets remove this

public/locales/en-US/translations.json
ckniffen: I dont think these need to be update at this time.

src/containers/App/test/App.test.js
Amindv1: the test also fails - but ill put it back in

src/containers/shared/test/utils.test.js
Amindv1: done

src/containers/shared/utils.js
Amindv1: weird, removed again

mvadari (approved):

PR #499, opened 2022-12-12
build: migrate to node 18.x

mvadari:

Title says it all.

webpack needs to be updated in order to remove the change to the `npm start` script and (I think) to fix the build issue.

### Context of Change

According to https://endoflife.date/nodejs, active support for Node v16 ended two months ago. Node v18 still has active support.

### Type of Change

- [x] Build

## Test Plan

Works locally. CI passes.


mvadari:
The build error is due to webpack/OpenSSL issues - if webpack is updated, it should be fixed.

mvadari:
I have no clue why this is failing. There is no difference in the diffs, and it passes locally.

ckniffen (approved):

pdp2121 (approved):

PR #500, opened 2022-12-13
feat: Block indexing by search engine bots

ckniffen:

We don't need every page indexed because they are all so dynamic

### Context of Change

Google was indexing every transaction and NFT.

### Type of Change

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


jonathanlei (approved):

public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
ledhed2222, line 2: don't you want:

```
User-agent: *
Allow: /
Disallow: /nft/
```

ledhed2222 (approved):

PR #509, opened 2022-12-14
refactor: migrate to vite from create-react-app

mvadari:

This PR migrates the build system from create-react-app and webpack to vite. 

Most of the file changes are switching env variables from `process.env.REACT_APP_...` to `process.env.VITE_...`, and most of the line changes come from `package-lock.json`.

### Context of Change

Vite is a better build system that requires less custom hacky code to get it to work properly. It also builds faster (down to 15-30 seconds from ~1.5 minutes in CI).

### Type of Change

## Before / After

It runs the same.

## Test Plan

Works locally and in CI.

## Future Steps

Test runtime can probably be improved.

[This](https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-scripts/scripts/test.js) is the script that CRA uses for its tests, so perhaps there is more mocking that we can take from there.


cssTransform.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
ckniffen, line 1: Does every transform need to be in a separate file?  Can they be moved to a folder called like `tooling`?  This would declutter the root.

src/rippled/lib/utils.js
```diff
@@ -96,8 +96,6 @@ function RippledError(message, code) {
   this.code = code
 }
 
-require('util').inherits(RippledError, Error)
```
ckniffen, line 99: Why was this removed?

src/rippled/lib/txSummary/formatAmount.js
```diff
@@ -1,4 +1,4 @@
-module.exports = (d) => {
+const formatAmount = (d) => {
```
ckniffen, line 1: could just export on this line.

src/rippled/lib/utils.js
mvadari: It was complaining when I ran `npm start`

mvadari: I also don't really understand what it's doing

ckniffen: It adds all properties of Error to RippledError.

mvadari: Wouldn't it be easier to do that with normal inheritance?

ckniffen: That works I just want to make sure the error properties/methods are there.

mvadari: I don't really know how to add this back in - some help would be appreciated.

tsconfig.json
```diff
@@ -10,8 +10,9 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
+    "noImplicitAny": false,
     "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "node",
+    "moduleResolution": "Node",
```
ckniffen, line 15: This should still be lowercase.

ckniffen (approved):

src/setupTests.ts
```diff
@@ -1,19 +1,25 @@
+import 'dotenv/config'
 import 'jest-enzyme'
 import { configure } from 'enzyme'
 import Adapter from 'enzyme-adapter-react-17-updated'
 
 const mockStorage = {}
 
+// @ts-ignore
```
shadymoses, line 8: Could also be `ts-expect-error`. Same for all of these.

src/index.html
```diff
@@ -36,6 +28,11 @@
     You need to enable JavaScript to run this app.
   </noscript>
   <div class="xrpl-explorer" id="xrpl-explorer"></div>
+  <script type="module">
+    import { Buffer } from "buffer";
+    window.Buffer = Buffer;
```
shadymoses, line 33: You shouldn't need this with the setup you have in `vite.config`. Was it causing problems without this script?

shadymoses (approved):

PR #512, opened 2022-12-16
ci: parallelize linting and tests

mvadari:

This PR splits linting, testing, and building into 3 separate Github Actions. Previously, building wasn't tested at all. This makes CI run faster, since it's now parallelized, saving up to about a minute of time.

### Context of Change

Faster CI

### Type of Change

### TypeScript/Hooks Update

## Test Plan

CI passes.

### DIFF
--- a/src/containers/Transactions/index.jsx
+++ b/src/containers/Transactions/index.jsx
@@ -44,23 +44,24 @@ const Transaction = (props) => {
   const { t } = useTranslation()
   const { actions, data, loading } = props
   const rippledSocket = useContext(SocketContext)
+  const hash = data.raw ? data.raw.hash : undefined
+  const short = identifier.substr(0, 8)
+  document.title = `${t('xrpl_explorer')} | ${t(
+    'transaction_short',
+  )} ${short}...`
 
   useEffect(() => {
-    const hash = data.raw ? data.raw.hash : undefined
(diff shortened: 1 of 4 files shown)


src/containers/Transactions/index.jsx
```diff
@@ -44,23 +44,24 @@ const Transaction = (props) => {
...
+  const short = identifier.substr(0, 8)
+  document.title = `${t('xrpl_explorer')} | ${t(
+    'transaction_short',
+  )} ${short}...`
 
   useEffect(() => {
-    const hash = data.raw ? data.raw.hash : undefined
-    const short = identifier.substr(0, 8)
-
-    document.title = `${t('xrpl_explorer')} | ${t(
```
ckniffen, line 52: We should still update title.

mvadari: It's updated outside of the `useEffect` now (line 49). Should it stay in the useEffect?

ckniffen: Sorry I missed that.  It is fine outside of useEffect.  Just check when switching to a different TX it still updates

mvadari: Confirmed that it does

pdp2121 (approved):

ckniffen (approved):

PR #513, opened 2022-12-16
refactor: clean up code from previous transaction scheme

ckniffen:

Removes the code that previously powered the old transaction scheme now that all transaction types have been refactored.

### Type of Change


src/containers/shared/components/TxDetails.tsx
```diff
@@ -1,148 +1,17 @@
-import React, { ReactElement } from 'react'
-import PropTypes from 'prop-types'
-import { useTranslation, withTranslation } from 'react-i18next'
-import { Amount } from './Amount'
+import React from 'react'
 import { transactionTypes } from './Transaction'
 
-interface Instructions {
```
ckniffen, line 7: Each tx defines what will be on the instructions.  This is redundant.

src/rippled/lib/txSummary/index.js
```diff
@@ -1,16 +1,8 @@
...
 const getInstructions = (tx, meta) => {
   const type = tx.TransactionType
-  // Locate the transaction parser which returns "instructions" to be used in transaction specific components
-  // TODO: Remove summarize[type] lookup once all transactions have been moved to the new definition style
-  const mappingFn = transactionTypes[type]?.parser
-    ? transactionTypes[type]?.parser
-    : summarize[type]
-
-  return mappingFn ? mappingFn(tx, meta) : {}
+  return transactionTypes[type]?.parser(tx, meta) || {}
```
mvadari, line 5: Thoughts on something like this for a better default case if there's a parser error?
```suggestion
  try {
    return transactionTypes[type]?.parser(tx, meta)
  catch {
    return tx
  }
```

Or should that be handled in a separate PR? I guess the parsers would all need to change as well.

mvadari (approved):

pdp2121 (approved):

src/rippled/lib/txSummary/index.js
ckniffen: Going with a separate PR.  I will create a ticket for this.

Issue #357, opened 2022-09-23
Add validator search


ckniffen:
Add ability to search by validator pubkey.  All validators start with "n" and are the same length so it should be easy to determine just by the string.

It should take the user to the validator detail page like https://livenet.xrpl.org/validators/nHDwHQGjKTz6R6pFigSSrNBrhNYyUGFPHA75HiTccTCQzuu9d7Za

PR #519, opened 2022-12-20
feat: add search by validators pubkey

pdp2121:

Allow users to search for validators by their pubkeys.
Linked issue: 
https://github.com/ripple/explorer/issues/357

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Users can search for validators by their pubkey:
Screenshot 2022-12-20 at 5 11 55 PM
Screenshot 2022-12-20 at 5 12 14 PM


JST5000 (approved):
LGTM

src/containers/shared/utils.js
```diff
@@ -33,6 +33,7 @@ export const CURRENCY_REGEX =
   /^[a-zA-Z0-9]{3,}\.r[rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz]{27,35}$/
 export const FULL_CURRENCY_REGEX =
   /^[0-9A-Fa-f]{40}\.r[rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz]{27,35}$/
+export const VALIDATORS_REGEX = /^n[9H][0-9A-Za-z]{50}$/
```
ckniffen, line 36: @mDuo13 is this a good regex for determining if a key is a validator?  Are there other rules?

src/containers/Header/Search.jsx
```diff
@@ -53,6 +54,9 @@ const getIdType = async (id, rippledContext) => {
   ) {
     return 'token'
   }
+  if (VALIDATORS_REGEX.test(id)) {
```
ckniffen, line 57: @pdp2121 can you add a unit test?

src/containers/shared/utils.js
intelliot: I'm not sure where this regex is from, but I believe all validation public keys start with `n9` or `nH` because of the way the [base58 encoding](https://xrpl.org/base58-encodings.html) works - using the predefined prefix byte and the XRPL alphabet ([implementation](https://github.com/XRPLF/xrpl.js/blob/76b73e16a97e1a371261b462ee1a24f1c01dbb0c/packages/ripple-address-codec/src/xrp-codec.ts#L239-L242)).

I think we can go with this regex, and update it in the future if needed.

intelliot (approved):

ckniffen (approved):

PR #531, opened 2023-01-09
lint: clean up un-linted js files

mvadari:

This PR lints `src/registerServiceWorker.js` and `src/setupTests.js`, which somehow aren't in the default eslint/prettier path.

### Context of Change

The files weren't linted, and it made some diffs larger than needed.

### Type of Change

## Before / After

N/A

## Test Plan

CI passes.


.eslintignore
```diff
@@ -1,3 +1,4 @@
 build/
 coverage/
 node_modules/
+src/registerServiceWorker.js
```
ckniffen, line 4: Why is this file being ignored?

package.json
```diff
@@ -224,6 +224,21 @@
...
+        "error",
+        {
+          "devDependencies": [
+            "**/*.test.js",
+            "**/*.test.jsx",
+            "**/*.test.ts",
+            "**/*.test.tsx",
+            "**/test/**",
+            "**/setupTests.ts",
+            "**/svgTransform.js",
```
ckniffen, line 238: Is this a file added in the vite work?

.eslintignore
mvadari: There are a bunch of errors and I didn't really know how to resolve them

ckniffen (approved):

jonathanlei (approved):
LGTM, should double check if it works with react-17

PR #533, opened 2023-01-09
fix: reduce frequency stable version is fetched

ckniffen:

Leverage useQuery to isolate the call to retrieve the stable version of rippled to only happen on page load and not when validators are retrieved from the VHS.

### Context of Change

Github was called every 5 seconds to get the current version of rippled. This was causing the Github rate limiter to be triggered.  If we still hit the limiter after this change the current version should be cached in the Explorer backend.

### Type of Change


src/containers/Network/UpgradeStatus.tsx
```diff
@@ -66,14 +65,22 @@ export const UpgradeStatus = () => {
     ['fetchUpgradeStatusData'],
     () => {
       fetchData()
-      fetchStableVersion()
     },
```
mvadari, line 68: This can be simplified to just `() => fetchStableVersion()` as well

mvadari (approved):

JST5000 (approved):
LGTM

PR #537, opened 2023-01-09
chore: remove language switching code

mvadari:

There is a bunch of code in `Footer` that isn't doing anything, regarding switching languages. This PR removes that code.

### Context of Change

The code isn't used anyways, since we don't have very many translations. The design also doesn't really fit the rest of our site, so it'd need to be redesigned anyways.

There are some Vite issues regarding this code, so it seems easier to just get rid of it.

### Type of Change

## Before / After

No visible changes.

## Test Plan

Works locally.


src/containers/Footer/index.jsx
```diff
@@ -229,7 +114,6 @@ const Footer = (props) => {
 }
 
 Footer.propTypes = {
-  language: PropTypes.string.isRequired,
   actions: PropTypes.shape({
     updateLanguage: PropTypes.func,
```
ckniffen, line 118: You can remove updateLanguage too which will allow the removal of redux and props from this component completely

src/containers/Footer/index.jsx
```diff
@@ -1,130 +1,16 @@
...
-import classnames from 'classnames'
-
-import { analytics, ANALYTIC_TYPES } from '../shared/utils'
 import { updateLanguage } from '../App/actions'
 
 import packageConfig from '../../../package.json'
 import { ReactComponent as Logo } from '../shared/images/XRPLedger.svg'
-import iconDownArrow from '../shared/images/down_arrow.svg'
-import iconCheck from '../shared/images/checkmark_small_green.png'
-import flagUSA from '../shared/images/flag_usa.png'
```
ckniffen, line 15: Should we just delete all these images too and maybe translation files that are never updated?

ckniffen (changes requested):
`updateLanguage` in /App/actions can be eliminated too.

src/containers/Footer/index.jsx
mvadari: I'm down to remove the images, but I think the translation files should stay - if/when we do decide to support multiple languages, it'll be a useful starting point.

ckniffen (approved):

jonathanlei (approved):

PR #547, opened 2023-01-17
feat: Update how transactions are differentiated

ckniffen:

Differentiate transactions by both type and action. Also adds a legend to the homepage to explain the new shapes and colors.

### Context of Change

Previously almost every transaction type was given its own color.  As more transactions were added this became overwhelming and there was no legend that showed what they all meant.

The new approach contains five action types each with their own shape slightly inspired by phases of the moon.

- Create: An open circle.
- Modify: A filled half circle.
- Finished: A filled full circle.
- Cancel: An open circle with a slash
- Send: A forward pointing arrow

Six colors were chosen to differentiate category of the transactions.

- Payment
- Dex (AMM, Offers, Trust sets, tokens)
- Account
- NFT
- XChain
- Pseudo

#### Round Up

Dex-Account
Payment-NFT
XChain-Psuedo

#### Additional Changes
- Transaction icons on the homepage animate with each growing from the center.
- Loading indicator will show on the homepage until the first ledger comes in.
- AMM color was changed to blue to match the DEX theme.
- The width of ledgers on the homepage has been increased.
- Adds `useLocalStorage` hook for leveraging local storage.

### Type of Change

- [x] Documentation Updates

##### Before
Homepage - Before

##### After
Homepage - After With Legend Expanded

Homepage - After

##### Before
Transaction Table Mobile - Before

Transaction Table Desktop - Before

##### After
Transaction Table Mobile - After

Transaction Table Desktop - After


src/containers/shared/components/Transaction/AccountDelete/index.ts
```diff
@@ -1,9 +1,15 @@
...
+  TransactionMapping,
+} from '../types'
 
 import { Simple } from './Simple'
 import { AccountDelete } from './types'
 
 export const AccountDeleteTransaction: TransactionMapping = {
   Simple,
   parser: (tx: AccountDelete): AccountDelete => tx,
+  action: TransactionAction.CANCEL,
```
mvadari, line 13: The name `action` feels a little confusing to me, especially given that we have redux actions-related pages still in the repo.

src/containers/shared/components/Transaction/types.ts
```diff
@@ -1,44 +1,23 @@
 import React from 'react'
 import { Memo } from '../../transactionUtils'
 
-export interface Instructions {
```
mvadari, line 4: Why was this type removed?

src/containers/shared/css/variables.scss
```diff
@@ -113,33 +113,10 @@ $testnet: $orange-60;
...
-$paymentchannelcreate: #7d8000;
-$paymentchannelfund: #004d23;
-$paymentchannelclaim: #4b4d00;
-$ticketcreate: #55805d;
-$enableamendment: #9ba2b0;
-$setfee: #9ba2b0;
-$unlmodify: #9ba2b0;
+
+// Feature Sets
+$amm: $blue;
```
mvadari, line 118: Is this changing for the network as well?

ckniffen: It is.

src/containers/shared/components/Transaction/types.ts
ckniffen: It was left over from the old transactions scheme.  It was never referenced.

src/containers/shared/components/Transaction/AccountDelete/index.ts
ckniffen: Open to other names. TransactionVerb?

mvadari: Sure

src/containers/shared/components/TransactionActionIcon/TransactionActionIcon.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface TransactionActionIconProps {
+  action?: TransactionAction
+  type?: string
+}
+
+export const TransactionActionIcon = ({
+  action,
+  type,
+}: TransactionActionIconProps) => {
+  const icons: Record<string, any> = {
```
mvadari, line 20: Is there an actual type to use instead of `any`? Isn't there a React component type?

intelliot (approved):
Makes the page look "busier" but the distinctions are actually informative, so it's an improvement overall

JST5000:
Two points of feedback:
1. Currently the legend hides the transactions from view on mobile when it's open, which is problematic. I'd recommend defaulting the legend to closed on mobile, especially since the on-hover tells you what kind of transaction is involved. 
2. This is exacerbated by the fact that the action item to "Hide Legend" is at the bottom, so it doesn't appear when the legend is overflowing off the screen. I'd recommend allowing any part of the legend to be clicked in order to collapse the section, instead of just the green text. 
image

src/containers/Ledgers/Legend.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ]
+
+  const categories = [
+    TransactionCategory.PAYMENT,
+    TransactionCategory.DEX,
+    TransactionCategory.NFT,
+    TransactionCategory.ACCOUNT,
+    TransactionCategory.PSEUDO,
+    TransactionCategory.UNKNOWN,
+  ]
```
mvadari, line 34: Can't you just use the enum itself instead of needing to list it out again?

ckniffen: This is a subset of all enums and a special order.

jonathanlei (approved):

ckniffen:
Added logic to hide legend on mobile before the user interacts with it the first time.

src/containers/Ledgers/Legend.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const LEGEND_STORAGE_KEY = 'explorer-legend-previous-interaction'
+
+export const Legend = () => {
+  const { t } = useTranslation()
+  const windowSize = useWindowSize()
+  const [previousInteraction, setPreviousInteraction] =
+    useLocalStorage<boolean>(LEGEND_STORAGE_KEY, false)
+  const [hidden, setHidden] = useState(previousInteraction)
+
+  // TODO: use global variables when we update places using width from redux.
```
ckniffen, line 21: Created #593 to track this.

package.json
```diff
@@ -295,7 +296,9 @@
             "**/svgTransform.js"
           ]
         }
-      ]
+      ],
+      "no-shadow": "off",
+      "@typescript-eslint/no-shadow": "error"
```
mvadari, line 301: What does this do?

src/containers/Ledger/ledger.scss
```diff
@@ -1,3 +1,4 @@
+@use 'sass:color';
```
mvadari, line 1: What does this do?

src/containers/shared/components/TransactionActionIcon/TransactionActionIcon.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface TransactionActionIconProps {
+  action?: TransactionAction
+  type?: string
+}
+
+export const TransactionActionIcon = ({
+  action,
+  type,
+}: TransactionActionIconProps) => {
+  const icons: Record<string, ReactElement> = {
```
mvadari, line 20:
```suggestion
  const icons: Record<TransactionAction, ReactElement> = {
```

src/containers/shared/components/Transaction/AccountDelete/index.ts
mvadari: bump

src/containers/shared/components/Transaction/AMMWithdraw/index.ts
```diff
@@ -1,8 +1,14 @@
...
+  TransactionCategory,
+  TransactionMapping,
+} from '../types'
 import { parser } from './parser'
 
 export const AMMWithdraw: TransactionMapping = {
   Simple,
+  action: TransactionAction.SEND,
   parser,
+  category: TransactionCategory.DEX,
```
mvadari, line 13: We should be consistent about the order of these elements (doesn't the linter do that?)

ckniffen: I can do that.  I am leery of a linter enforcing element order.  It can be difficult when you want to group for documentation purposes.

src/containers/Ledger/ledger.scss
ckniffen: Allows you to use built in sass color methods.  I think it is no longer needed in this file.  I was using color.adjust earlier.

src/containers/shared/components/Transaction/AMMWithdraw/index.ts
mvadari: That's fair.

package.json
ckniffen: This fixes a parsing bug in eslint for how typescript enums are treated.  It would think you were using the same variable name twice.  This disables the no-shadow rule in core eslint but then enables an alternative one in `@typescript-eslint`

src/containers/shared/components/Transaction/AccountDelete/index.ts
ckniffen: I was joking about calling it TransactionVerb, I should have made that clearer :) .  I could not come up with a better name.  I don't think it will be cause too much confusion with redux actions as they are typed.

JST5000: Best I can think of is maybe `TransactionCategory`?

ckniffen: I just added `TransactionCategory`.  The tricky part is I needed to add names for two new taxonomies.

mvadari: I used `Category` and `Type` in my initial breakdown

mvadari: `Action` is fine for now, we can always change it later if we decide it's confusing

src/containers/shared/components/TransactionActionIcon/TransactionActionIcon.tsx
```diff
@@ -0,0 +1,36 @@
...
+import TransactionCancelIcon from './TransactionCancelIcon.svg'
+import TransactionCreateIcon from './TransactionCreateIcon.svg'
+import TransactionFinishIcon from './TransactionFinishIcon.svg'
+import TransactionModifyIcon from './TransactionModifyIcon.svg'
+import TransactionSendIcon from './TransactionSendIcon.svg'
+import TransactionUnknownIcon from './TransactionUnknownIcon.svg'
+
+export interface TransactionActionIconProps {
+  action?: TransactionAction
+  type?: string
```
mvadari, line 13: Why are these both optional? Shouldn't they be mandatory?

src/containers/shared/components/Transaction/AccountDelete/index.ts
ckniffen: > I used `Category` and `Type` in my initial breakdown

I was worried `Type` would get confused with `TransactionType`.

src/containers/shared/components/TransactionActionIcon/TransactionActionIcon.tsx
ckniffen: Because you can provide the action or the `TransactionType` which will derive the action for you.  One is required.

mvadari: Wouldn't it be better to do something like `type TransactionActionIconProps = { action: TransactionAction } | { type: string }`

ckniffen: I get `TS2339: Property 'action' does not exist on type 'TransactionActionIconProps'.` on line 14 and 15 if I change it to that.

Had to update to

```
type TransactionActionIconProps =
  | { action: TransactionAction; type: never }
  | { action: never; type: string }
```

mvadari (approved):

PR #551, opened 2023-01-18
chore: enable new jsx transform

ckniffen:

This was added in react@17 and is more performant and removes the need to import `React` in every file containing JSX.

It also removes the need to set DISABLE_NEW_JSX_TRANSFORM which sometimes caused the tsconfig.json to flap a change.

### Type of Change


src/containers/NFT/test/NFT.test.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
mvadari, line 1: Why is this needed?

src/containers/shared/EmptyMessageTableRow.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
mvadari, line 1: ditto

src/containers/shared/components/Transaction/SimpleRow.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
mvadari, line 1: ditto

src/containers/shared/components/Transaction/test/createWrapperFactory.tsx
```diff
@@ -1,5 +1,5 @@
 import { mount, ReactWrapper } from 'enzyme'
-import React from 'react'
+import * as React from 'react'
```
mvadari, line 2: ditto

src/containers/shared/components/Transaction/types.ts
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
mvadari, line 1: ditto

src/containers/NFT/test/NFT.test.js
ckniffen: Due to the way the test is mocking `useState`.  I was able to update all the other places though.  Good catch there were 12 places.  I used a script that was supposed to rename all of those but I guess it missed a few.

src/containers/CustomNetworkHome/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { KeyboardEvent, MouseEvent, useState } from 'react'
+import { KeyboardEvent, MouseEvent as ReactMouseEvent, useState } from 'react'
```
mvadari, line 1: Does importing it as `MouseEvent` not work?

ckniffen: No because we also need the native MouseEvent type in the file.

mvadari (approved):

khancode (approved):

PR #558, opened 2023-01-20
feat: remove ZenDesk widget

ckniffen:

I don't know when this was used but it is never activated on any of our deployed instances.


intelliot (approved):

intelliot:
This should probably be `refactor:` instead of `feat:` according to [this](https://stackoverflow.com/questions/48075169/semantic-commit-type-when-remove-something)

PR #559, opened 2023-01-24
fix: EscrowCancel table detail and test warnings

ckniffen:

- `<Trans>` warning turned out to be the component was used incorrectly and the test was inadequate
- `toHaveTagName` -> `toHaveDisplayName`
- use `createWrapper` in `<Amount>` test

### Type of Change

## Before
EscrowCancel Before

## After
EscrowCancel After


JST5000:
What is that "- 9" on the end of the green text in the screenshot? (I can't figure out what it's supposed to represent from looking at it)

JST5000 (approved):
LGTM, just a few questions for my understanding.

ckniffen:
> What is that "- 9" on the end of the green text in the screenshot? (I can't figure out what it's supposed to represent from looking at it)
> 
> Also nit: The space between the green text and the "- 9" seems to be absorbed / disappeared for some reason, even though it is in the code. (It looks like "...xt56- 9" instead of "...xt56 - 9")

The 9 is the sequence number of the escrow that was cancelled.  I will look into the absorbed space.

ckniffen:
@JST5000 fixed the spacing issue.

intelliot (approved):

PR #587, opened 2023-02-07
feat: support for issued currencies on Escrows and PaymentChannels

dangell7:

Updates to the explorer to reflect the changes made by amendment xls34d.

### Context of Change

Updated the Tx Description to reflect both XPR and currencies
Updated the account to update the escrow when the user changes the primary (current) currency.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Test Plan

Please review. I didn't do paychannels as I wanted to get your thoughts on the escrow. I dont like how its only the currency and not the issuer but the explorer only uses the currency so I think it might suffice unless you add that additional functionality. 

If you do want to add that functionality in rippled we use `USD/rIssuerAddress` for the `Amount` string.

### DIFF
--- a/src/containers/shared/utils.js
+++ b/src/containers/shared/utils.js
@@ -231,10 +231,14 @@ export function formatPrice(number, options = {}) {
 // Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat
 export const localizeDate = (date, lang = 'en-US', options = {}) => {
   // TODO: default config
-  if (!date) {
+  try {
+    if (!date) {
+      return null
+    }
+    return new Intl.DateTimeFormat(lang, options).format(date)
+  } catch {
     return null
   }
-  return new Intl.DateTimeFormat(lang, options).format(date)
 }
 
 /**
--- a/src/containers/Accounts/AccountHeader/index.jsx
+++ b/src/containers/Accounts/AccountHeader/index.jsx
@@ -68,6 +68,7 @@ const AccountHeader = (props) => {
     )
   }
 
+  const options = { ...CURRENCY_OPTIONS, currency: currencySelected }
   function renderPaymentChannels() {
     const { paychannels } = data
     return (
@@ -95,6 +96,8 @@ const AccountHeader = (props) => {
 
   function renderEscrows() {
     const { escrows } = data
+    const fIn = escrows && escrows.totalIn[currencySelected]
+    const fOut = escrows && escrows.totalOut[currencySelected]
     return (
       escrows && (
         <div className="escrows secondary">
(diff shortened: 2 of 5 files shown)


dangell7:
A small video showing the functionality..

https://user-images.githubusercontent.com/25502424/217374332-b7873885-2eb9-479a-ab96-aada337e7af4.mov

mvadari:
Please add testcases.

ckniffen:
I think the PR should be broken out into transaction updates and then one for account page changes.

For the account page I agree with you that grouping by currency is odd.  That is one reason we created the assets tab.  

@bugsbunnies do you think it makes sense to add an Escrows and Paychannel tabs so that we can show the breakdowns there?  This would be similar to the account page for XRPScan.  The amount column would show what currency and issuer it is using the <Amount> component we use else where.

Monosnap rLSn6Z3T8uCxbcd1oxwfGQN1Fdn5CyGujK - XRP Account | XRPSCAN 2023-02-08 14-43-47

bugsbunnies:
Concept after discussion with @ckniffen 
image

dangell7:
> Concept after discussion with @ckniffen
> 
> Transactions image
> 
> Escrow as a separate tab image

This looks really amazing!

mvadari:
> Concept after discussion with @ckniffen
> 
> Transactions image
> 
> Escrow as a separate tab image

A thought: could the tabs not appear when a user doesn't have escrows or payment channels? Since most users won't have either.

ckniffen:
> A thought: could the tabs not appear when a user doesn't have escrows or payment channels? Since most users won't have either.

I like the idea of them always being there.  That way we don't have to make requests to get them everytime the user visits an account page like we currently do.  This will speed up load time.

mvadari:
> > A thought: could the tabs not appear when a user doesn't have escrows or payment channels? Since most users won't have either.
> 
> I like the idea of them always being there. That way we don't have to make requests to get them everytime the user visits an account page like we currently do. This will speed up load time.

We already make the requests every time the user visits an account page.

ckniffen:
> We already make the requests every time the user visits an account page.

Only because we currently show the escrow and paychannel information outside of the tabs.  I am suggesting this change could move it out.

src/containers/shared/utils.js
```diff
@@ -231,10 +231,14 @@ export function formatPrice(number, options = {}) {
 // Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat
 export const localizeDate = (date, lang = 'en-US', options = {}) => {
   // TODO: default config
-  if (!date) {
+  try {
+    if (!date) {
+      return null
+    }
+    return new Intl.DateTimeFormat(lang, options).format(date)
+  } catch {
```
mvadari, line 239: What does this try-catch block catch that wasn't caught before? Seems better to solve the root cause instead of using this blanket fix.

mvadari:
Looks like tests are currently failing on this PR.

intelliot:
For reference: `PaychanAndEscrowForTokens` amendment in rippled - https://github.com/XRPLF/rippled/pull/4396

PR #618, opened 2023-03-08
refactor: NoMatch to functional component

ckniffen:

- Add back default landing pages for `/accounts` and `/transactions`
- Add ability to customize warning
- Fix missing space between icon and warning
- Update default i18n key because their name was causing notFound to be false so it would not show `"Not Found"` warning by default.

### Type of Change

- [x] Documentation Updates

### TypeScript/Hooks Update

#### Before
Account No Id - Before

#### After 
Account No Id - After

#### Before
Transaction No Hash - Before

#### After 
Transaction No Hash - After

#### Before
Transaction Invalid Hash - Before

#### After 
Transaction Invalid Hash - After

## Future Work

Switch to Typescript.


pdp2121 (approved):

pdp2121:
Can this handle ledger not found when clicking on a validator on outdated ledger as well?

ckniffen:
> Can this handle ledger not found when clicking on a validator on outdated ledger as well?

I think that should be handled differently because we have all the data except for the hash of the ledger this validator last reported.  We should display everything normally and just make the ledger index show as white.

What this should be used for is when a validator cant be found at all.

pdp2121 (approved):

src/containers/NoMatch/index.jsx
```diff
@@ -1,54 +1,58 @@
...
 
-  render() {
-    const { t, title, hints, isError = true } = this.props
-    const notFound = title.includes('not_found')
-    const hintMsg = hints.map((hint) => (
-      <div className="hint" key={hint}>
-        {t(hint)}
-      </div>
-    ))
+  const notFound = title.includes('not_found')
```
JST5000, line 20: Should this be documented somewhere so that future people adding `not_found`-like pages know to do this?

src/containers/NoMatch/index.jsx
```diff
@@ -1,54 +1,58 @@
...
+  const hintMsg = hints.map((hint) => (
+    <div className="hint" key={hint}>
+      {t(hint)}
+    </div>
+  ))
+  const derivedWarning = warning ?? (notFound && t('not_found'))
```
JST5000, line 26: If `notFound` is false, would this produce an empty warning, or would it print `false`? (I'm not quite clear on how this would resolve)

JST5000 (approved):
LGTM, just two small comments

PR #619, opened 2023-03-08
feat: Try all nodes and allow backup nodes

ckniffen:

s1 and s2 stopped responding. This will allow the explorer to try multiple entry points at once and choose the winner.

### Type of Change


ckniffen:
Still need to add some tests.

src/containers/shared/SocketContext.tsx
```diff
@@ -23,7 +23,19 @@ function getSocket(rippledUrl?: string): XrplClient {
...
+    process.env.VITE_RIPPLED_SECONDARY.split(',').forEach((network) => {
+      wsUrls.push.apply(wsUrls, [
+        `${prefix}://${network}:${process.env.VITE_RIPPLED_WS_PORT}`,
+        `${prefix}://${network}:443`,
+      ])
+    })
+  }
+
+  const socket = new XrplClient(wsUrls, {
+    tryAllNodes: true,
```
mvadari, line 37: Wasn't this causing problems? Were those fixed?

mvadari (approved):

src/containers/shared/SocketContext.tsx
ckniffen: They were fixed in 2.0.3

src/containers/shared/SocketContext.tsx
```diff
@@ -12,25 +12,37 @@ function isInsecureWs(rippledHost: string | undefined): boolean {
 }
 
 function getSocket(rippledUrl?: string): XrplClient {
```
JST5000, line 14: Should `getSocket` accept an optional string array of urls instead of being converted to an array immediately?

src/containers/shared/test/SocketContext.test.ts
```diff
@@ -19,18 +19,47 @@ describe('getSocket', () => {
 
   describe('server defined entrypoint', () => {
     beforeEach(() => {
+      delete process.env.VITE_RIPPLED_SECONDARY
```
JST5000, line 22: Why is this necessary? (Maybe worth a comment?)

JST5000: Does this have side effects if run on production? (i.e. is the process.env value permanently removed?)

ckniffen: I will remove this.  This was originally there because I was going to have two environment variables.  Since all entry points are tried at once it made no sense for them to be called secondary.

src/containers/shared/SocketContext.tsx
ckniffen: I kept it this way to make the handling of the url as an environment variable or user specified work the same.

src/containers/shared/test/SocketContext.test.ts
mvadari: First off, this is a test, so it's not run in production. Second off, there's a `beforeEach` in the outside `describe` that saves and replaces `process.env`.

JST5000 (approved):
LGTM

PR #628, opened 2023-03-14
refactor: tx page react-query and typescript

ckniffen:

- Use react-query instead of redux.
- Update Transaction component to typescript
- Remove ~200 lines of code

### Type of Change

### TypeScript/Hooks Update


src/containers/Header/Banner.jsx
```diff
@@ -22,7 +22,6 @@ Banner.propTypes = {
 export default connect((state) => {
   const messages = [
     ['ledgerError', state.ledger.error],
-    ['transactionError', state.transaction.error],
```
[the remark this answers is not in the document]
ckniffen: Correct.  The previous only showed the same message everytime and would not always disappear.

src/containers/test/utils.ts
ckniffen, line 1: Will have a future PR to leverage this in other tests.

justinr1234 (approved):

src/containers/Transactions/index.tsx
```diff
@@ -85,17 +108,18 @@ const Transaction = (props) => {
   }
 
   function renderTransaction() {
-    const { language, width } = props
     let body
 
+    if (!data) return <></>
```
khancode, line 113: nit: move above `let body`

khancode (approved):
LGTM with one nit comment

PR #630, opened 2023-03-15
fix: handle multiple rippled hosts in backend

ckniffen:

This also gets rid of a redundant node connection that was used only for getting the streams of ledgerClosed.

### Context of Change

#619 added the feature to the frontend but using it breaks the backend.

### Type of Change


JST5000 (approved):
LGTM

PR #631, opened 2023-03-20
fix: show parsed currency code in balance

ckniffen:

Update account balance selection to use parsed currency code.

### Type of Change

## Before
Monosnap XRPL Explorer | rsoLo2S1kiGe  2023-03-20 09-55-41

## After
Monosnap XRPL Explorer | rsoLo2S1kiGe  2023-03-20 09-54-57

### DIFF
--- a/src/containers/Accounts/AccountHeader/BalanceSelector.jsx
+++ b/src/containers/Accounts/AccountHeader/BalanceSelector.jsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types'
 import { localizeNumber } from '../../shared/utils'
 import IconDownArrow from '../../shared/images/down_arrow.svg'
 import iconClose from '../../shared/images/close.png'
+import Currency from '../../shared/components/Currency'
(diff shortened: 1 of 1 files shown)


mvadari:
IMO this should be of the form `SOLO.rABC...` given the many different USDs and such

lathanbritz:
> IMO this should be of the form `SOLO.rABC...` given the many different USDs and such

I like the idea, but i x characters after the token still leaves it open to overlap. One already has the full break down when selecting the assets tab. Maybe when one has selected the asset (you can click and select it) then it shows the issuer.

ckniffen:
> IMO this should be of the form `SOLO.rABC...` given the many different USDs and such

The way it works now is these totals are a roll up of a currency regardless of issuer.  If the user wants to see the totals by issuer they can go to the assets tab.

JST5000 (approved):
LGTM

justinr1234 (approved):

PR #633, opened 2023-03-20
refactor: token page hooks

ckniffen:

Future update will remove redux and switch to typescript.

### Type of Change

### TypeScript/Hooks Update


src/containers/Token/index.jsx
```diff
@@ -35,16 +36,21 @@ ERROR_MESSAGES.default = {
...
-    return {
-      accountId: match.params.id,
-      prevId: prevState && prevState.accountId,
+const Token = ({ error }) => {
+  const { currency, id: accountId } = useParams()
+  const { t } = useTranslation()
+  const showError = error
+
+  useEffect(() => {
+    document.title = `${t('xrpl_explorer')} | ${accountId.substr(0, 12)}...`
```
justinr1234, line 45: Why not use helmet?

src/containers/Token/index.jsx
```diff
@@ -53,57 +59,32 @@ class Token extends Component {
...
-        {!accountId && (
-          <div style={{ textAlign: 'center', fontSize: '14px' }}>
-            <h2>Enter an account ID in the search box</h2>
-          </div>
-        )}
-      </div>
-    )
-  }
+  return showError ? (
+    renderError(error)
```
justinr1234, line 63: Why not refactor to a component? <Error

if(error)  {
  return <Error
}

return <div

justinr1234 (approved):

src/containers/Token/index.jsx
ckniffen: I will create a ticket to do this across the board.  These were existing methods and it is how it is done in many places throughout the app.

src/containers/Token/index.jsx
ckniffen: This is how it is done throughout the app.  I will create a ticket to update this to helmet everywhere.

ckniffen: #650

src/containers/Token/index.jsx
ckniffen: Created #651

PR #638, opened 2023-03-21
refactor: paystring using hooks, ts, and fixes

ckniffen:
Updates the PayString page to use hooks, typescript in addition to the following fixes and enhancments

- Fixed PayString logo being scrunched
- Fixed Hovering over info icon crashed the page.
- Render tooltip text as html. Saves 38kb gzipped. App now 7% smaller.
- Use i18n for text values
- Fix error messaging

### Context of Change

The error messaging code in <PayString> never ran because the redux variable being monitored was incorrectly set to `accountHeader.status`.  The correct variable is `payStringData.error` and all that it ever emits is `resolve_paystring_failed`.  I updated to always show that message at the page level if the PayString retrieval fails.

### Type of Change

### TypeScript/Hooks Update

#### Before
Table - Before

#### After
Table - After

#### Before
Table Loading - Before

#### After
Table Loading - After

#### Before (Once crash is fixed)
Tooltip - Before

#### After
Tooltip - After

#### Before
Empty  - Before

#### After
Empty  - After

#### Before

Error - Before

#### After
Error - After


ckniffen:
> > @bugsbunnies this might be of interest to you.
> 
> Thanks! For the tooltip bg color, can we use black 60 just like the "before" screenshot? The dark tooltip really blends in with the bg and it's hard to tell what's what.

I can try that.  Is it okay if I make that a global tooltip change and put it in a new PR?  The previous design was a total one off.

JST5000 (approved):
LGTM

bugsbunnies:
> > > @bugsbunnies this might be of interest to you.
> > 
> > 
> > Thanks! For the tooltip bg color, can we use black 60 just like the "before" screenshot? The dark tooltip really blends in with the bg and it's hard to tell what's what.
> 
> I can try that. Is it okay if I make that a global tooltip change and put it in a new PR? The previous design was a total one off.

Yeah that sounds good thanks!

pdp2121 (approved):

PR #643, opened 2023-03-24
refactor: retrieve country data on page load

ckniffen:

Moves country data out of the js bundle and instead load it when the user visits the node page. This reduces the bundle by 39kb (gzipped).

### Type of Change


pdp2121 (approved):

JST5000 (approved):
LGTM

PR #644, opened 2023-03-27
fix: custom network node server crash

ckniffen:

The custom network version of the explorer would crash because no rippled host is supplied.

### Type of Change


JST5000 (approved):
LGTM

PR #648, opened 2023-03-29
fix: custom mode double initialized xrpl-client

### LINKED ISSUE #645: custom mode has trouble connecting when no port is specified.
The App component re-renders twice and causes the `xrpl-client` to be double instantiated.

ckniffen:

Dynamically determining the VHS network id was causing the sockets to be reinitialized when the response was received. This moves the logic into their own components with their own scope that doesn't trigger everything to be called again.  This would only happen in custom mode which is when network id is unknown upon load.

Fixes #645

### Type of Change


src/containers/App/test/App.test.js
```diff
@@ -192,4 +199,14 @@ describe('App container', () => {
...
     })
   })
+
+  it('renders custom mode', async () => {
+    process.env.VITE_ENVIRONMENT = 'custom'
+    const wrapper = createWrapper('/s2.ripple.com/')
+    await flushPromises()
+    wrapper.update()
+    // Make sure the sockets aren't double initialized. Called twice. Once for main host and p2p
+    expect(XrplClient).toHaveBeenCalledTimes(2)
```
JST5000, line 209: Can this ever just be 1 if there is no P2P host specified?

JST5000 (approved):
LGTM - one question, but it's about a test

src/containers/App/App.tsx
```diff
@@ -46,28 +40,9 @@ const App = ({ actions }: AppProps) => {
     return function cleanup() {
       window.removeEventListener('resize', actions.updateViewportDimensions)
       window.removeEventListener('scroll', actions.onScroll)
-
-      socket.close()
-      if (socket.p2pSocket !== undefined) {
-        socket.p2pSocket.close()
-      }
```
mvadari, line 53: Why was this part removed?

ckniffen: It was moved inside the new `<SocketProvider>` component which handles all setup and teardown of the SocketContext.

mvadari (approved):

PR #639, opened 2023-03-21
refactor: paystring remove redux and payid-lib

ckniffen:
## High Level Overview of Change

Switch from redux to react-query and payid-lib to @paystring/utils.

This eliminates ~230 lines.

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)

### TypeScript/Hooks Update

- [x] Updated files to React Hooks
- [x] Updated files to TypeScript

PR #649, opened 2023-03-31
refactor: paystring remove redux and payid-lib

ckniffen:

Switch from redux to react-query and payid-lib to @paystring/utils.

This eliminates ~200 lines.

This is a recreation of #639.

### Type of Change

### TypeScript/Hooks Update


pdp2121 (approved):

src/containers/PayStrings/PayStringMappingsTable/test/PayStringMappingsTable.test.js
```diff
@@ -1,86 +1,44 @@
...
 
   it('renders without crashing', () => {
-    const wrapper = creatWrapper()
+    const wrapper = createWrapper()
     wrapper.unmount()
   })
 
   it('renders static parts', () => {
-    const wrapper = creatWrapper()
+    const wrapper = createWrapper()
```
JST5000, line 21: nice catch :) (Took me a couple reads to see the difference)

src/rippled/payString.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface PayStringResponse {
+  payString: string
+  addresses: {
+    paymentNetwork: string
+    environment: string
+    addressDetailsType: string
+    addressDetails: {
+      address: string
+      tag: string // optional
```
JST5000, line 18: Should this be written as `tag?: string` instead of this comment? or is the implication that it can be empty?

JST5000 (approved):
LGTM after a small question

src/rippled/payString.ts
ckniffen: Just confirmed the output from the server and it will be undefined.

PR #662, opened 2023-04-06
fix: only use captureStackTrack when available

### LINKED ISSUE #661: Indicating when an account does not exist
It should be possible to clearly indicate that an account does not exist when it does not exist.

The current error message makes me think that there is some error on the server side.

https://livenet.xrpl.org/accounts/rndpDXFdkgR4EmDXnPgC36VxRBbjqu8rHP

image

ckniffen:

This was broken when code was ported to the browser about a year ago.

Fixes #661

### Type of Change

### DIFF
--- a/src/rippled/lib/utils.js
+++ b/src/rippled/lib/utils.js
@@ -95,7 +95,9 @@ const formatTransaction = (tx) => {
 }
 
 function RippledError(message, code) {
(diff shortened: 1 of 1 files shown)


JST5000 (approved):
LGTM

mvadari (approved):

Issue #554, opened 2023-01-20
Add language picker


ckniffen:
Add language picker back. @develoQ has put in a PR (#553) to update the Japanese translations.

Recently the language picking code was removed as it had been hidden from the user for over 3 years.  We need to add back the capability to change to Japanese and also add fallbacks for any values that are not yet translated so that keys arent shown when there is no value for them.

PR #667, opened 2023-04-12
refactor: account and token header tsx and hooks

### LINKED ISSUE #476: Email hash may not be ever displayed in account header
Email hash may not be ever shown in account header.
Props declares email hash as snake case in [AccountHeader](https://github.com/ripple/explorer/blob/8ac4e416097d03b6ab2faad44390d0a28f6259e6/src/containers/Accounts/AccountHeader/index.js#L355). But inside [formatAccountInfo](https://github.com/ripple/explorer/blob/8ac4e416097d03b6ab2faad44390d0a28f6259e6/src/rippled/lib/utils.js#L62), email hash is returned in camelcase.

Worth looking in having test cases

ckniffen:

- Make `<AccountHeader>` and `<TokenHeader>` use typescript and hooks
- Remove incorrect types
- Remove last usages of language property on redux state
- Fix EmailHash not showing up (types were incorrect)

Fixes #476 and related to #554

### Context of Change

With the new language picker all updates to language happen directly through i18n-next.  All references to language now come through the `useLanguage` hook instead of redux.

In order to use the hook `<TokenHeader>` had to be refactored to a functional component. `<AccountHeader>` is a very similar component (`<TokenHeader>` was based on it) so it make sense to complete the Typescript refactor of that component.

While setting up the component's types I fixed #476 since I did not want incorrect types.  Additionally `<TokenHeader>` had many props defined that were copied from `<AccountHeader>` but are never returned by `getToken`.

### Type of Change

### TypeScript/Hooks Update


src/containers/Accounts/AccountHeader/index.tsx
```diff
@@ -21,16 +21,63 @@ const CURRENCY_OPTIONS = {
...
+  loading: boolean
+  accountId: string
+  data: {
+    balances: {
+      XRP: number
+    }
+    paychannels: {
+      // eslint-disable-next-line camelcase
+      total_available: string
+      channels: any[]
```
justinr1234, line 36: Why any

src/containers/Accounts/AccountHeader/index.tsx
```diff
@@ -317,60 +364,8 @@ const AccountHeader = (props) => {
...
-  }).isRequired,
-  actions: PropTypes.shape({
-    loadAccountState: PropTypes.func.isRequired,
-  }).isRequired,
-}
-
 export default connect(
-  (state) => ({
-    language: state.app.language,
+  (state: any) => ({
```
justinr1234, line 368: Why any

src/containers/Token/TokenHeader/index.tsx
```diff
@@ -158,85 +174,33 @@ class TokenHeader extends Component {
...
+        {loading ? <Loader /> : renderHeaderContent()}
+      </div>
+    </div>
+  )
 }
 
 export default connect(
-  (state) => ({
-    language: state.app.language,
+  (state: any) => ({
```
justinr1234, line 203: Why any

justinr1234 (approved):

src/containers/Accounts/AccountHeader/index.tsx
ckniffen: The previous definition was `PropTypes.shape({ length: PropTypes.number })`.  No other properties are referenced so I figured any[] array matches that.  This might be something to address when following up on #587.

src/containers/Accounts/AccountHeader/index.tsx
ckniffen: It is just a mapping and will soon go away when redux is fully removed.  The account and token pages are the last ones using redux for data.  The other redux usages will be removed with the new header.

src/containers/Token/TokenHeader/index.tsx
ckniffen: Same answer as `<AccountHeader>`.

jonathanlei (approved):

Issue #681, opened 2023-04-19
Transaction page crashes for EscrowCreate with no FinishAfter or CancelAfter


ckniffen:
Example transaction: https://testnet.xrpl.org/transactions/90376FAC0AF53B8729414703F1CC68EFA37ADB2EA0D5477549D103C37036C8A4

```
{
    "tx": {
        "Account": "r4bqzg7iZFBGmLyYRa3o6vgtCMDzvVoRCh",
        "Amount": "1241",
        "CancelAfter": 735241162,
        "Condition": "A02580207B6FB711478A7FB792E4BD6BCB111C1389751569468578CCD49261DA824095CE810120",
        "Destination": "rhDEt27CCSbdA8hcnvyuVniSuQxww3NAs3",
        "Fee": "12",
        "Flags": 0,
        "LastLedgerSequence": 37127595,
        "Memos": [
            {
                "Memo": {
                    "MemoData": "6469643A6B75646F733A75726C3A68747470733A2F2F6769746875622E636F6D2F686F6F3239",
                    "MemoFormat": "746578742F706C61696E",
                    "MemoType": "72656C6179746F2E6964656E746966696572"
                }
            }
        ],
        "Sequence": 36125103,
        "SigningPubKey": "027B8F2C5F8345BFB049E4FDCFA533B53BD958EE783200233D7A2EC95F0587DE59",
        "TransactionType": "EscrowCreate",
        "TxnSignature": "304402204364BC4645BDF5D32AC9F99E925799C89499294E2AF752771DCDE655AB2B449A02201D620453AE71764F112B1597261312E1F9F481AFAC0C31772DB0CED9204B0812",
        "date": 1681925662000
    },
    "meta": {
        "AffectedNodes": [
            {
                "CreatedNode": {
                    "LedgerEntryType": "Escrow",
                    "LedgerIndex": "51D2E46947434797380C522094A6B8EDB5D38B3401928D514E3537F58A82FFF6",
                    "NewFields": {
                        "Account": "r4bqzg7iZFBGmLyYRa3o6vgtCMDzvVoRCh",
                        "Amount": "1241",
                        "CancelAfter": 735241162,
                        "Condition": "A02580207B6FB711478A7FB792E4BD6BCB111C1389751569468578CCD49261DA824095CE810120",
                        "Destination": "rhDEt27CCSbdA8hcnvyuVniSuQxww3NAs3",
                        "DestinationNode": "a",
                        "OwnerNode": "9"
                    }
                }
[...]

PR #682, opened 2023-04-19
fix: Transaction crash with no cancel or finish

justinr1234:
Cloees #681


khancode (approved):

ckniffen:
TableDetail might suffer from the same issue.

justinr1234:
> TableDetail might suffer from the same issue.

Follow up issue

justinr1234:
> Can you add a unit test for these cases?

Follow up issue, I think getting fix out is priority

mvadari:
> TableDetail might suffer from the same issue.

Doesn't appear to be an issue with TableDetail: https://testnet.xrpl.org/accounts/r4bqzg7iZFBGmLyYRa3o6vgtCMDzvVoRCh

ckniffen (approved):

PR #703, opened 2023-05-08
feat(i18n): bump i18next and add typings

ckniffen:

Newer versions of i18next necessitated improving the typings of translations. As a result the build now fail if a translation is referenced that is not in the `en-US` file.  This already caught several missing and/or misspelled keys.

The `context` option of `t` is now being leveraged for enum based translations.  Other dynamic places had to use `as any` to get around typing errors.  Luckily there are only a few and in the future they could also leverage context.

Replaces #690

### Type of Change


public/locales/en-US/translations.json
```diff
@@ -141,47 +146,47 @@
...
-  "transaction_XChainClaim": "XChain Claim",
-  "transaction_XChainCommit": "XChain Commit",
-  "transaction_XChainCreateBridge": "XChain Create Bridge",
-  "transaction_XChainCreateClaimID": "XChain Create Claim ID",
-  "transaction_XChainModifyBridge": "XChain Modify Bridge",
-  "transaction_EnableAmendment": "Enable Amendment",
-  "transaction_SetFee": "Set Fee",
-  "transaction_UNLModify": "UNL Modify",
-  "transaction_AccountDelete": "Account Delete",
+  "transaction_type_name_AMMCreate": "AMM Create",
```
mvadari, line 149: Why were these changed?

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
```diff
@@ -79,7 +79,7 @@ export const AccountTransactionTable = ({
       transactions={transactions}
       loading={loading}
       hasTokensColumn={hasTokensColumn}
-      emptyMessage={emptyMessage && t(emptyMessage)}
+      emptyMessage={emptyMessage && t(emptyMessage as any)}
```
mvadari, line 82: Should there be `TODO`s on all these casts to come back and fix later?

src/i18n/baseConfig.ts
```diff
@@ -1,9 +1,12 @@
-import i18n, { InitOptions } from 'i18next'
+// @ts-check
```
mvadari, line 1: Why is this necessary in a TS file?

ckniffen: Good catch.  This was from me running around.

src/containers/Accounts/AccountTransactionTable/AccountTransactionTable.tsx
ckniffen: I dont think so in this case.

mvadari (approved):

Issue #621, opened 2023-03-10
Add a node info section to the header


ckniffen:
Add a place in the header where the user can see the following:

- Entrypoint used
- public key
- rippled/clio version
- complete ledgers
- uptime

bugsbunnies:
Proposed design (Work in Progress)

<img width="1038" alt="image" src="https://user-images.githubusercontent.com/39572746/224452914-68af3c6c-c8b0-4dc6-9178-3b5f6162dd60.png">

<img width="1032" alt="image" src="https://user-images.githubusercontent.com/39572746/224452944-52519f28-3f37-4040-a990-691a5f09c596.png">

rippleitinnz:
looks good

ckniffen:
@bugsbunnies @mvadari due to this PR there will be more room in the footer. #623

bugsbunnies:
Makes sense

> @bugsbunnies @mvadari due to this PR there will be more room in the footer. #623

PR #705, opened 2023-05-12
feat: network picker redesign and language picker

### LINKED ISSUE #554: Add language picker

ckniffen:

- Redesigned network picker
- Add more network picker tests
- Add language picker with support for Japanese.
- Cleanup i18next configuration
- Set foundation to add localStorage powered custom networks

Closes #554 and sets up #621

### Context of Change

In relation to #621 this makes the top part of the header use flex box to make it easier to add additional items such as connected node info.

### Type of Change

### TypeScript/Hooks Update

##### Before
Mobile - Before

##### After
Mobile - After

##### Before
Network Picker Open - Before
Custom Network Picker - Before

##### After
Network Picker Open - After
Custom Network Picker - After

##### Before
Mobile Network Picker Open - Before
Mobile Custom Network Picker Open - Before

##### After
Mobile Network Picker Open - After
Mobile Custom Network Picker Open - After

###  Language Picker
Mobile Language Picker Open - After
Language Picker Open - After


src/containers/Header/NetworkPicker/NetworkPicker.tsx
```diff
@@ -0,0 +1,131 @@
...
+      .toLowerCase()}`
+
+  const trackNetworkSwitch = (network: Network) => {
+    analytics(ANALYTIC_TYPES.event, {
+      eventCategory: 'mode switch',
+      eventAction: network.url,
+    })
+  }
+
+  const networks = [
```
justinr1234, line 47: Any reason not to make these things global?

justinr1234 (approved):

src/containers/Header/NetworkPicker/NetworkPicker.tsx
ckniffen: No but these are the only place these are used.

justinr1234: I meant outside the render path. Global to the file.

ckniffen: I moved it inside because it was messing with unit tests.  Having it inside the render made it happen later.

Is there a better way to mock environment variables for a test so that the file is evaluated during ever `it`?

justinr1234 (approved):

jonathanlei (approved):
LGTM, much better design/info now

src/containers/Header/NetworkPicker/NetworkPicker.tsx
ckniffen: Used jest.isolateModules to test the env variables

PR #719, opened 2023-05-26
feat: remove old translation files

ckniffen:

These files were written by google translate, haven't been updated since June 2021, and are missing 250+ entries.

### Type of Change


JST5000 (approved):
LGTM (Assuming that if you try to view those translations it doesn't let you - e.g. by setting your default language to "ko-KP" as an example)

justinr1234 (approved):

ckniffen:
> LGTM (Assuming that if you try to view those translations it doesn't let you - e.g. by setting your default language to "ko-KP" as an example)

That is correct.  It will revert to English.

PR #716, opened 2023-05-26
feat: add `XChainBridge` info on door accounts' Account pages

mvadari:
## High Level Overview of Change

This PR adds a section to the Account Header of a page (next to the Signer List) with bridge details, if the account is a door account, and adds a little banner at the top. It also redoes the CSS a bit for the signer list.

### Context of Change

Better presentation of bridge info

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

### TypeScript/Hooks Update

N/A

### Door Account
Before:
image

After:
Monosnap XRPL Explorer | rMAXACCrp3Y8  2023-06-01 12-14-13

### Non-door account
Before:
image

After:
Monosnap XRPL Explorer | r481fwr9G6Gq  2023-06-01 12-22-00

## Test Plan

CI passes. Works locally.

PR #721, opened 2023-05-30
feat: add `XRP` symbol when XRP is the currency

mvadari:

Title says it all.

### Context of Change

Pulled out from #716 

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

CI passes.

### DIFF
--- a/src/containers/shared/types.ts
+++ b/src/containers/shared/types.ts
@@ -1,6 +1,6 @@
 export interface IssuedCurrency {
   currency: string
-  issuer: string
+  issuer?: string
 }
 
 export interface IssuedCurrencyAmount extends IssuedCurrency {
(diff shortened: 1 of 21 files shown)


src/containers/shared/types.ts
```diff
@@ -14,3 +20,43 @@ export type ExplorerAmount = {
   currency: string
   amount: number
 }
+
+export interface Node {
+  DeletedNode?: any
```
justinr1234, line 25: Why any?

justinr1234 (approved):

src/containers/shared/types.ts
mvadari: I didn't write these types, I just moved them from another file. I assume laziness/simplicity.

ckniffen:
Please add a test to `/shared/components/test/Currency.test.tsx`.  I know the code changes probably already has coverage due to it being used in so many places it would set my mind at ease.

ckniffen (approved):

PR #733, opened 2023-06-02
feat: store previously-visited custom networks

mvadari:

Title says it all.

### Context of Change

You don't need to keep entering the same URLs over and over again, there's a handy link.

### Type of Change

### TypeScript/Hooks Update

N/A

### Custom Network Page
Before:
image

After:
image

### Network Picker
Before:
image

After:
image

## Test Plan

Works locally.


src/containers/Header/NetworkPicker/NetworkPicker.tsx
```diff
@@ -116,12 +140,12 @@ export const NetworkPicker = () => {
       <>
         {networks.map(({ network, title, url = '' }) => {
           if (
-            network === currentMode ||
+            (currentMode !== 'custom' && network === currentMode) ||
```
ckniffen, line 143: Can you add a comment explaining what is going on here.  I forgot what the reasoning is.

src/containers/Header/NetworkPicker/NetworkPicker.tsx
```diff
@@ -69,6 +76,23 @@ export const NetworkPicker = () => {
         handler={handleNetworkClick(network)}
       >
         {text}
+        {network === 'custom' && (
```
ckniffen, line 79: Use `isCustom` instead.

ckniffen:
Should you be able to remove the custom networks on the homepage?

ckniffen (approved):

mvadari:
> Should you be able to remove the custom networks on the homepage?

That wasn't part of the initial design, but I could add that if you want (replace the arrows with an X)

ckniffen:
> > Should you be able to remove the custom networks on the homepage?
> 
> That wasn't part of the initial design, but I could add that if you want (replace the arrows with an X)
I think it is fine for now just something to consider.

pdp2121 (approved):

PR #757, opened 2023-06-23
feat: add support for `SetHook` transaction

mvadari:

Title says it all.

### Context of Change

Better hooks support

### Type of Change

### TypeScript/Hooks Update

Files added used TS/hooks

## Before / After

`SetHook` now has a custom `Simple` page.

image

image

## Test Plan

Added tests. CI passes. Works locally.


src/containers/shared/components/Transaction/SetHook/utils.ts
```diff
@@ -0,0 +1,133 @@
+import { zeroPad } from '../../../transactionUtils'
+
+const TRANSACTION_TYPES: Record<string, number> = {
```
dangell7, line 3: This might get annoying to update every time a new transaction is added. I exported from the xrpl.js the TRANSACTION_TYPE_MAP. Might think about that. Should also add the URIToken transactions.

dangell7 (approved):

src/containers/shared/components/Transaction/SetHook/utils.ts
mvadari: We're working on some tree-shaking from ripple-binary-codec, given the size of the package I think we'll need to wait on that first.

src/containers/shared/components/Transaction/SetHook/test/SetHookSimple.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { createSimpleWrapperFactory } from '../../test/createWrapperFactory'
+import { Simple } from '../Simple'
+import mockSetHook from './mock_data/SetHook.json'
+import mockSetHook2 from './mock_data/SetHook2.json'
+import { expectSimpleRowText } from '../../test/expectations'
+
+const createWrapper = createSimpleWrapperFactory(Simple)
+
+describe('SetHookSimple', () => {
```
ckniffen, line 9: Please add a test with a failing transaction especially with the metadata being parsed (a common issue).

src/containers/shared/components/Transaction/SetHook/Simple.tsx
```diff
@@ -0,0 +1,72 @@
+import { useTranslation } from 'react-i18next'
+import { buildHookFlags } from '../../../transactionUtils'
+import { Account } from '../../Account'
+import { SimpleGroup } from '../SimpleGroup'
+import { SimpleRow } from '../SimpleRow'
+import { TransactionSimpleProps } from '../types'
+import { HookData, SetHookInstructions } from './types'
+import { hookOnToTxList } from './utils'
+
+// interface TxPillProps {
```
ckniffen, line 10: Commented code.

src/containers/shared/components/Transaction/SetHook/utils.ts
mvadari: ripple-binary-codec needs to be cleaned up a bit before it can be used in the Explorer without introducing a lot of bloat

src/containers/shared/components/Transaction/SetHook/Simple.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { t } = useTranslation()
+
+  const renderHook = (hook: HookData) => (
+    <SimpleGroup title={t('hook')} key={hook.HookHash}>
+      <SimpleRow label={t('hash')} data-test="hook-hash">
+        {hook.HookHash ?? 'undefined'}
+      </SimpleRow>
+      {hook.HookOn && (
+        <SimpleRow label={t('triggered_on')} data-test="hook-on">
+          {/* {hookOnToTxList(hook.HookOn).map((tx) => (
```
jonathanlei, line 34: commented code here too

ckniffen (approved):

src/containers/shared/components/Transaction/SetHook/parser.ts
```diff
@@ -0,0 +1,23 @@
...
+      (node.ModifiedNode?.LedgerEntryType === 'Hook' &&
+        !!node.ModifiedNode?.PreviousFields?.Hooks),
+  )
+  const hashes = affectedNodes.flatMap((node: any) =>
+    (node.ModifiedNode?.FinalFields ?? node.CreatedNode?.NewFields)?.Hooks?.map(
+      (hook: any) => hook.Hook.HookHash,
+    ),
+  )
+  // TODO: there may be bugs here when a `HookHash` is already specified in a hook
+  // It's difficult to understand what situation that would be in, so this is left here for now
```
jonathanlei, line 17: is there a test for this?

src/containers/shared/components/Transaction/SetHook/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+    // eslint-disable-next-line no-param-reassign -- fine for a reduce
+    flipped[value] = key
+    return flipped
+  },
+  {} as Record<number, string>,
+)
+
+const maxTransactionValue: number = 103
+
+function hex2bin(input) {
```
jonathanlei, line 52: is this function from somewhere already? 
there could be a simplification using `parseInt` and `padStart`
something like: 
`
function hex2bin(input) {
    let hex = input.replace('0x', '').toLowerCase();
    let bin = '';
    for (let i = 0; i < hex.length; i++) {
        let binFragment = parseInt(hex[i], 16).toString(2);
        bin += binFragment.padStart(4, '0');
    }
    return bin;
}`

src/containers/shared/components/Transaction/SetHook/parser.ts
mvadari: No, because I don't know what situations this can arise in (the docs aren't specific enough)
That's why I left a TODO

src/containers/shared/components/Transaction/SetHook/utils.ts
mvadari: The other `hex2bin` didn't work, because the number in decimal would be too large. I'll try this, though

pdp2121 (approved):

Issue #392, opened 2022-10-07
Strip leading and trailing quotes from search values


ckniffen:
Per @legleux 

> “copy to clipboard” buttons on the explorer raw page copies the quotes too (which is fine bc you’d need them for pasting sometimes) but the search input should at least strip quotes instead of saying “page not found”

Simply removing any leading `"` or `'` from the search string before applying parsing logic will fix this annoyance.

ckniffen:
Fixed by #759

PR #759, opened 2023-06-26
strip leading and trailing quotes from search

jonathanlei:

referring to issue https://github.com/ripple/explorer/issues/392
programmatically strip the quotes  
Before: 
Screenshot 2023-06-26 at 3 00 43 PM

after:
Screenshot 2023-06-26 at 3 00 06 PM

### DIFF
--- a/src/containers/Header/Search.tsx
+++ b/src/containers/Header/Search.tsx
@@ -104,15 +104,17 @@ export const Search = ({ callback = () => {} }: SearchProps) => {
   const history = useHistory()
 
   const handleSearch = async (id: string) => {
-    const type = await getIdType(id, socket)
-
+    const strippedId = id.replace(/^["']|["']$/g, '')
+    const type = await getIdType(strippedId, socket)
     track('search', {
-      search_term: id,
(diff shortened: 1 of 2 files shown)


justinr1234:
Do we also strip leading and trailing space?

ckniffen:
Please add a test.

jonathanlei:
> Do we also strip leading and trailing space?

no - this is mainly for people copying from raw that always contain the quotes, but if we think it's useful to strip trailing spaces then I can add it

ckniffen:
> Do we also strip leading and trailing space?

stripping leading and spaces would be good too.

jonathanlei:
> > Do we also strip leading and trailing space?
> 
> stripping leading and spaces would be good too.

actually we do strip it already - https://github.com/ripple/explorer/blob/3182ebac65b0ff2e38cd474d74c8b532898df041/src/containers/Header/Search.tsx#L122

justinr1234 (approved):

ckniffen (approved):

PR #760, opened 2023-06-26
fix: clean up `Transaction` page borders

mvadari:

This PR shortens the line between the two columns in the Transaction Simple page (which extended well past the text) and removes the bottom line in the last element of the Simple page.

### Context of Change

The edges were messy. Noticed while working on #757 

### Type of Change

### TypeScript/Hooks Update

N/A

### Payment
Before:
image

After:
image

### SetHook (a longer transaction)
Before:
image

After:
image

## Test Plan

Works locally.


ckniffen (approved):

src/containers/Transactions/simpleTab.scss
```diff
@@ -94,6 +94,14 @@ $subdued-color: $black-40;
       background: rgba($black-80, 0.7);
       gap: 15px;
 
+      &:first-child {
```
ckniffen, line 97: I would think gap would be accomplishing this.

jonathanlei (approved):

src/containers/Transactions/simpleTab.scss
mvadari: Apparently not /shrug

PR #764, opened 2023-06-29
Clawback transaction

shawnxie999:
If the tx is a success, the _actual_ amount that was clawed back is returned and shown in all components.
If the tx failed, the _attempted_ amount is shown in all components.

### Context of Change
Clawback transaction. See https://github.com/XRPLF/XRPL-Standards/pull/104/files?short_path=cb82c33#diff-cb82c337d579659136224b35dbbaac97d0ffaa3cfaa79981f1bebf4f4888feb5

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After
image

Successful Simple
image

Failed Simple
image

Success table detail
image

Failed table detail
image

Description
image


src/containers/shared/components/Transaction/Clawback/parser.ts
```diff
@@ -0,0 +1,18 @@
+import { Clawback, ClawbackInstructions } from './types'
+import { TransactionParser } from '../types'
+import { formatAmount } from '../../../../../rippled/lib/txSummary/formatAmount'
+
+export const parser: TransactionParser<Clawback, ClawbackInstructions> = (
+  tx,
+) => {
+  const account = tx.Account
+  let amount = formatAmount(tx.Amount)
+  const holder = amount.issuer
```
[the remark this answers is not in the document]
shawnxie999: image

It's designed to avoid an extra parameter. Otherwise, the account that submits the tx is the same as the `issuer` field, which would be duplicated. So we utilize the `issuer` for the holder's account instead. 

it's a bit confusing but i believe there has been a precursor somwhere, written in doc
image

mvadari (approved):

ckniffen:
Can you add a `<TableDetail>` and a `<Description>` component for the type?  They should be really straight forward.

I think you can just put amount and holder for `<TableDetail>` and the `<Description>` should be a nice sentence describing the transaction in sentence form.

src/containers/shared/components/Transaction/Clawback/Simple.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const Simple: TransactionSimpleComponent = ({
+  data,
+}: TransactionSimpleProps<ClawbackInstructions>) => {
+  const { account: issuer, amount, holder } = data.instructions
+  const { t } = useTranslation()
+
+  return (
+    <>
+      <SimpleRow label={t('issuer')} data-test="issuer">
```
ckniffen, line 16: This seems redundant since it is already in the amount.

shawnxie999: I don’t think it would hurt tho. It’s really minimal, and I put there to just make it dummy-proof

mvadari: Isn't the issuer also the sender of the transaction? Seems extra unnecessary

shawnxie999: fixed

mvadari (commented):
Given the difference in the amount shown (total actually clawed back vs total attempted to be clawed back), should there be a difference in the label? e.g. Amount vs Attempted Amount or something

ckniffen:
> Given the difference in the amount shown (total actually clawed back vs total attempted to be clawed back), should there be a difference in the label? e.g. Amount vs Attempted Amount or something

Or treat it like partial payments with a callout?

shawnxie999:
> Given the difference in the amount shown (total actually clawed back vs total attempted to be clawed back), should there be a difference in the label? e.g. Amount vs Attempted Amount or something

How much would it matter? The only thing that's relevant in a successful clawback transaction is how much token that was actually clawed back. The only time when the actual amount is different from the specified is if the holder owns less token than the specified amount, in which case, the issuer claws back the entirety of the holder's supply. 

The actual amount clawed back will always _less than or equal to_ the specified amount. There is not much information lost here. And if the person wants to know what the request looked like, they could always look at the raw JSON.

public/locales/en-US/translations.json
```diff
@@ -452,5 +452,9 @@
   "language_ja-JP": "Japanese",
   "xchain_account_claim_count": "XChain Account Claim Count",
   "xchain_account_create_count": "XChain Account Create Count",
-  "min_signer_quorum": "Minimum weight <0>{{quorum}}</0> required"
+  "min_signer_quorum": "Minimum weight <0>{{quorum}}</0> required",
+  "holder": "Holder",
+  "claws_back": " Claws back",
+  "claws_back_from": "<source/> claws back from <destination/>",
+  "instruct_to_claw": "It was instructed to claw back at most "
```
mvadari, line 459:
```suggestion
  "instruct_to_claw": "The max clawback amount is ____"
```

mvadari: This should be a translation that takes a param, because other languages might have this in a different order

src/containers/shared/components/Transaction/Clawback/Description.tsx
```diff
@@ -0,0 +1,30 @@
...
+            source: <Account account={issuer} />,
+            destination: <Account account={holder} />,
+          }}
+        />
+      </div>
+      <div data-test="amount-line">
+        {t('instruct_to_claw')}
+        <Amount value={formatAmount(amount)} />
+      </div>
+    </>
```
mvadari, line 28: IMO this should all be one translation

src/containers/shared/components/Transaction/Clawback/TableDetail.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <div>
+      {amount && holder && (
+        <div className="clawback">
+          <span className="label">{t('claws_back')}</span>
+          <Amount value={amount} displayIssuer />
+          <span>{t('from')}</span>
+          <Account account={holder} />
+        </div>
```
mvadari, line 21: IMO this should all be one translation

src/containers/Transactions/Meta/RippleState.jsx
```diff
@@ -3,7 +3,7 @@ import { CURRENCY_OPTIONS } from '../../shared/transactionUtils'
 import { localizeNumber } from '../../shared/utils'
 import { Account } from '../../shared/components/Account'
 
-const render = (t, language, action, node, index) => {
+export const computeBalanceChange = (node) => {
```
mvadari, line 6: This function should move to the utils file

src/containers/shared/components/Transaction/Clawback/types.ts
```diff
@@ -0,0 +1,12 @@
+import { TransactionCommonFields } from '../types'
+import { Amount } from '../../../types'
+
+export interface Clawback extends TransactionCommonFields {
+  Amount: Amount
+}
+
+export interface ClawbackInstructions {
+  account: string
+  amount?: { currency: string; amount: number; issuer?: string }
```
mvadari, line 10: Shouldn't this be `ExplorerAmount`?

src/containers/shared/components/Transaction/Clawback/TableDetail.tsx
shawnxie999: i dont think so. The `claws_back` translation needs the `className="label"` in order to appear in bold

ckniffen: You can use the `<Trans>` component to accomplish this as one translation.

src/containers/shared/components/Transaction/Clawback/test/ClawbackSimple.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { createSimpleWrapperFactory, expectSimpleRowText } from '../../test'
+import { Simple } from '../Simple'
+import transaction from './mock_data/Clawback.json'
+import transactionFailure from './mock_data/Clawback_Failure.json'
+
+const createWrapper = createSimpleWrapperFactory(Simple)
+
+describe('Clawback', () => {
+  it('handles Clawback simple view ', () => {
```
ckniffen, line 9: Can you add a test where the clawback amount is less than the `Amount` field?

src/containers/shared/utils.js
```diff
@@ -287,3 +287,42 @@ export const formatTradingFee = (tradingFee) =>
         maximumFractionDigits: 3,
       })
     : undefined
+
+export const CURRENCY_OPTIONS = {
```
ckniffen, line 291: Please put this at the top with the other default formatting values.

src/containers/shared/components/Transaction/Clawback/TableDetail.tsx
shawnxie999: fixed

ckniffen (approved):

PR #774, opened 2023-07-06
fix: support higher versions of node 18

mvadari:

Title says it all.

### Context of Change

My laptop uses 18.16 now and wasn't able to run `npm i`.

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

Works locally.


jonathanlei:
> Might as well bump to 20 and fix the tests.

would 18 still work if we bump to 20?

ckniffen:
> > Might as well bump to 20 and fix the tests.
> 
> would 18 still work if we bump to 20?

No but this is an app so we can set whatever version we want.

mvadari:
Tests will be fixed by #775

ckniffen (approved):

PR #787, opened 2023-07-24
Fix Clawback Description account swap

shawnxie999:

Accounts fields are mutated because of swapping values on an object reference. We now construct a new  object using spread operator.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### TypeScript/Hooks Update

### DIFF
--- a/src/containers/shared/components/Transaction/Clawback/Description.tsx
+++ b/src/containers/shared/components/Transaction/Clawback/Description.tsx
@@ -7,7 +7,10 @@ import { formatAmount } from '../../../../../rippled/lib/txSummary/formatAmount'
 export const Description = ({ data }: TransactionDescriptionProps) => {
   const issuer = data.tx.Account
   const holder = data.tx.Amount.issuer
-  const amount = data.tx.Amount
+
+  // construct a new amount object by spreading to avoid mutating
+  // the original object reference
+  const amount = { ...data.tx.Amount }
   amount.issuer = issuer
   return (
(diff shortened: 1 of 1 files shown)


src/containers/shared/components/Transaction/Clawback/Description.tsx
```diff
@@ -7,7 +7,10 @@ import { formatAmount } from '../../../../../rippled/lib/txSummary/formatAmount'
 export const Description = ({ data }: TransactionDescriptionProps) => {
   const issuer = data.tx.Account
   const holder = data.tx.Amount.issuer
-  const amount = data.tx.Amount
+
+  // construct a new amount object by spreading to avoid mutating
+  // the original object reference
+  const amount = { ...data.tx.Amount }
```
mvadari, line 13: Is this an issue in the parser too?

shawnxie999: i dont believe so. In the parser, it calls `formatAmount`, which returns a new object rather an reference to the raw data

mvadari: Why not call `formatAmount` in this too?

shawnxie999: that works too. will make change

ckniffen (approved):

mvadari (approved):

PR #788, opened 2023-07-31
feat: add French translation

lucbocahut:

Translation into French, using the methodology outlined in the original readme. 

### Context of Change

We provided the relevant language field translation using the built in locale folder structure. No changes were made to the main code to add missing internationalisation fields that might have been missing. 

### Type of Change

- [x ] New feature (non-breaking change which adds functionality)

## Test Plan

I ran the repo locally and tested by manually going through all the transaction types I could to check that the detailed descriptions worked. All other UI areas were also checked for clarity.

## Future Tasks
We will work on adding a few missing fields which prevent the whole interface from being properly translated (for instance action verbs in the detailed transaction descriptions) or the legend headers on the main view with legends.


public/locales/fr-FR/translations.json
```diff
@@ -0,0 +1,477 @@
+{
+  "action": "action",
+  "assets": "Actifs",
+  "assets.nft_tab_title": "Jetons non fongibles",
```
JST5000, line 4: Should each of these words be capitalized since this is a title? 

(More general pattern - looking at the english translation, some of these phrases use Title Case, while others use all lowercase, and others use special capitalization. I think it'd be worth doing a check through to make sure the translated match the casing - I think most do from the look of it, but since I don't speak French I can't tell what words shouldn't be capitalized)

https://github.com/ripple/explorer/blob/staging/public/locales/en-US/translations.json

ckniffen:
@lucbocahut Can you take another look at this PR?

PR #808, opened 2023-08-25
fix: Display AMM Account properly after `lsfAMM` became `AMMID`

JST5000:

Updates the view for AMM Accounts to properly display the AMM view instead of the normal account view.

### Context of Change

`lsfAMM` was removed, but `AMMID` is now returned when doing an `account_info` request. Also refactors the logic to look up the AMM object based on the new AMM changes to use `AMMID` instead of looking up the originating `AMMCreate` transaction.

### Type of Change

- [X] Refactor (non-breaking change that only restructures code)

## Before / After

Before:
image

After:
image

## Test Plan

CI Passes


ckniffen (approved):

src/containers/Accounts/AMM/AMMAccounts/index.tsx
```diff
@@ -70,50 +63,47 @@ export const AMMAccounts = () => {
     Get the first account transaction which in this case should be AMMCreate. From this we get
     the two assets in the asset pool.
     */
-    return (
-      getAccountTransactions(rippledSocket, accountId, 1, undefined, true)
```
mvadari, line 74: The `getAccountTransactions` method can be simplified a bit now to remove the last param

src/rippled/lib/rippled.js
```diff
@@ -74,6 +74,70 @@ const getLedger = (rippledSocket, parameters) => {
   })
 }
 
+// get ledger_entry
+const getLedgerEntry = (rippledSocket, parameters) => {
```
mvadari, line 78: This (the `parameters` field) seems unnecessarily complex for this - we can add more parameters later.

mvadari: All of the other functions in this file really only use parameters as-needed, which is why I'm suggesting we do the same here. Nothing here needs to be backwards-compatible, so we can add that later if needed.

mvadari: All of the other functions in this file really only use parameters as-needed, which is why I'm suggesting we do the same here. Nothing here needs to be backwards-compatible, so we can add that later if needed, since it'll be clearer for now.

jonathanlei (approved):

Issue #807, opened 2023-08-21
Don't translate language names


develoQ:
`language_en-US`, `language_ja-JP` and `language_fr-FR` should be provided in English without conversion to other languages.

If users change to a language they are not familiar with, they may not be able to return to the original language.

mdazad1508:
could you give more insight of the issue?

develoQ:
For example, if an English speaker mistakenly selects Japanese, he or she will want to revert to the original English.

In such a case, the English speaker does not know which option to choose.

mdazad1508:
Got it . could you please review my PR.
* #809

PR #809, opened 2023-08-27
Fixes issue : Don't translate language names #807

mdazad1508:
## This Pull Request makes following changes
* fixes issue #807 

### Type of Change

- [ ✔] Bug fix (non-breaking change which fixes an issue)

## Before 

image

## After
Screenshot (2)


src/containers/Header/LanguagePicker/LanguagePicker.tsx
```diff
@@ -5,8 +5,16 @@ import './LanguagePicker.scss'
 import { supportedLanguages } from '../../../i18n/baseConfig'
 
 export const LanguagePicker = () => {
-  const { t, i18n } = useTranslation()
-  const currentLanguage = useLanguage()
+  const { i18n } = useTranslation()
+  const currentLanguage = useLanguage();
+  let lang:String;
+  if(currentLanguage=='en-US'){
+    lang  = "English";
```
ckniffen, line 12: I would prefer this use a map that could be defined here https://github.com/ripple/explorer/blob/staging/src/i18n/baseConfig.ts#L3.

A map will eliminate the need for all these if statements.

src/containers/Header/LanguagePicker/LanguagePicker.tsx
```diff
@@ -15,21 +23,19 @@ export const LanguagePicker = () => {
...
             <DropdownItem
               key={language}
               className={`language-picker-${language}`}
               handler={handleLanguageClick(language)}
             >
-              {t(`language`, {
-                context: language,
-                defaultValue: '',
-              })}
+             {language=='en-US'?"English":language=='ja-JP'?"Japaneese":"Korean"}
```
ckniffen, line 38: Same here.  This will also make it so less files need to be updated when adding a language.

ckniffen:
Please remove the language entries from the translation json values.

src/containers/Header/LanguagePicker/LanguagePicker.tsx
mdazad1508: got it what you saying , lemme try again , thanks for feedback

mdazad1508:
i've removed the language entries from translations json of korean , because korean was the one which was not working properly , now its working fine locally . take a look
Screenshot (3)

mdazad1508:
types and eslint error has been fixed , waiting for approval

JST5000:
Hey @mdazad1508, I think the expected output for this would be that each language name is outputted in it's own language.  

So English would look like "English", Japanese would look like 日本語 and Korean would look like 한국어. 

That way if someone accidentally changed everything to a language they didn't speak, they'd still be able to tell which language was the langauge they spoke (as they may not know the english spelling for their language)

ckniffen:
Fixed with #858

PR #810, opened 2023-08-29
feat: Add AMMDelete Transaction

JST5000:

Adds the `AMMDelete` transaction and corresponding pages / views

### Context of Change

Added as part of the AMM changes for the explorer.

### Type of Change

- [X] New feature (non-breaking change which adds functionality)

## Before / After

Simple:
image

Detailed:
image

TableView: (You can get here by clicking the account for the `AMMDelete` transaction)
image

## Test Plan

Added SimplePage CI 

You can also inspect the changes in the preview by going to this extension on the AMM Devnet to see an AMMDelete transaction in action: accounts/rm5c42Crqpdch5fbuCdHmSMV1wrL9arV9


ckniffen (approved):

src/containers/shared/components/Transaction/AMMDelete/Description.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <div data-test="amm-delete-description">
+      <Trans
+        i18nKey="amm_delete_description"
+        components={{
+          Asset: (
+            <Currency
+              currency={Asset.currency}
+              issuer={(Asset as any).issuer}
```
mvadari, line 19: If you want to have this error when this is fixed in xrpl.js, you can just use `ts-expect-error` on this line

src/containers/shared/components/Transaction/AMMDelete/test/AMMDeleteDescription.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Description } from '../Description'
+import { createDescriptionWrapperFactory } from '../../test'
+
+const createWrapper = createDescriptionWrapperFactory(Description, i18n)
+
+describe('AMMDelete: Description', () => {
+  it('renders description for AMMDelete transaction', () => {
+    const wrapper = createWrapper(mockAMMDelete)
+
+    expect(wrapper.html()).toBe(
```
mvadari, line 12: This is not a good way to test - it's difficult to edit the tests when things change minorly.

JST5000: Do you happen to know what might be better?

I think it's good to test that the translations work, and that the hyperlinks are added, since those are things that can't easily be checked via code review. 

And in this case, as far as I can tell, the only tangential change which would accidentally fail this test is if we changed how we hyperlink currencies (`<span class="currency">...</span>` and `<a class="" href="/token/FOO.rm5c42Crqpdch5fbuCdHmSMV1wrL9arV9">...</a>`)

The rest of the object is specific to the `AMMDelete` logic. 

(Or is the problem that I'm using `wrapper.html()` instead of a more specific lookup?)

ckniffen (approved):

src/containers/shared/components/Transaction/AMMDelete/test/AMMDeleteDescription.test.tsx
JST5000: Nevermind - updated to have individual checks instead of a one-liner

pdp2121 (approved):

PR #828, opened 2023-09-15
Add transaction_type_name `AMMDelete`, `Invoke`

develoQ:

Add `AMMDelete`, `Invoke`

### Type of Change

- [x ] Documentation Updates


ckniffen (approved):

mvadari (approved):

mvadari:
Do they need to be added as `null` to the Korean translation?

develoQ:
> Do they need to be added as `null` to the Korean translation?

For processing purposes, it is the same either way, with or without the addition.

Setting it to null makes it clear that it is untranslated, so there is no need to compare it with original version.

PR #842, opened 2023-09-28
fix: use v2 api for `nft_info` and `nft_history`

ckniffen:

With clio specifying version 1 is forwarding to rippled regardless of if the command exists on rippled.

### Type of Change


shawnxie999 (approved):

mvadari (commented):
Why not edit it at the `queryP2P` method?

ckniffen:
> Why not edit it at the `queryP2P` method?

`queryP2P` method is used for `gateway_balance` so I dont want to expose the explorer to any changes to that api in v2.

mvadari (approved):

PR #835, opened 2023-09-25
fix: always specify `api_version: 1`

ckniffen:
## High Level Overview of Change

`clio` defaults the `api_version` to `2` and rippled` to `1`.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #843, opened 2023-09-28
fix: show info for old transactions

ckniffen:

Viewing old transactions was broken by #835. When sending `api_version: 1` to clio it will currently forward to a rippled node which is not full history. That will be addresses soon but until then wont send now that for every request.

Explicitly send `api_version` when getting `account_info` because we only want the result from the most recently validated ledger.

### Type of Change


pdp2121 (approved):
LGTM

mvadari (approved):

PR #836, opened 2023-09-25
feat: add amendments list page

pdp2121:
## High Level Overview of Change

The Explorer will roll out new features to include voting data (amendments + fee) soon. This page will be one of the new feature added.

Add a amendments page to display all amendments in voting and enabled.

Note that the links to specific amendment summary page (name, id) is not yet enabled (still white text). It will be updated once the amendment summary page has been created.

**Please do not merge this PR until the validator-history-service has been updated with voting data.**

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Desktop View

Screenshot 2023-10-12 at 2 53 18 PM

### Mobile View

Screenshot 2023-09-25 at 5 24 08 PM

Add Amendment Summary page for each individual amendment

Issue #831, opened 2023-09-19
Unable to look up ledger by ledger hash in the search box


intelliot:
Example: DFBD364C8A518FC936AF4AC68DCBF3282BF40340AA839F8E08AAE74047B55A07

Expected result: show ledger [DFBD364C8A518FC936AF4AC68DCBF3282BF40340AA839F8E08AAE74047B55A07](https://livenet.xrpl.org/ledgers/DFBD364C8A518FC936AF4AC68DCBF3282BF40340AA839F8E08AAE74047B55A07)

Actual result: search box incorrectly redirects to https://livenet.xrpl.org/nft/DFBD364C8A518FC936AF4AC68DCBF3282BF40340AA839F8E08AAE74047B55A07 as if the hash is being interpreted as an NFT.

Note: some other explorers like XRPSCAN handle this correctly

intelliot:
(of course, the page says "No NFTs found.")

mvadari:
I believe hashes only work right now for transactions and NFTs. https://github.com/ripple/explorer/issues/544 addresses the error message.

Issue #544, opened 2023-01-12
When searching for a transaction that doesn't exist you are show "NFT Not Found"


ckniffen:
Instead of completely assuming that if a transaction is not found for a hash we should tell the user we tried to look up by NFT and Transaction but found nothing.

intelliot:
Not sure what the error currently says, but it might be helpful to mention that the tx (if not found) might be on some other parallel network

PR #855, opened 2023-10-03
feat: add amendment summary page

pdp2121:

The page would include:

- Information about the amendment
- List of voted Yes and No validators (for amendments in voting)
- A graph showing distribution of votes broken down by UNL and non-UNL validators (for amendments in voting)

This PR would also include search by Amendment Id/Name

### Type of Change

### Amendment in voting (Desktop)

localhost_3001_amendment_Clawback (1)

### Amendment in voting (Mobile)

localhost_3001_amendment_Clawback(iPhone 12 Pro) (1)

### Amendment enabled
Screenshot 2023-10-03 at 7 09 56 PM

### Chart legend

Screenshot 2023-10-04 at 11 49 39 AM

### Footnote

Screenshot 2023-10-04 at 4 36 09 PM


bugsbunnies:
@pdp2121 The `Yeas` and `Nays` colors are reversed, should be Yeas - Green, Nays - Magenta

pdp2121:
> @pdp2121 The `Yeas` and `Nays` colors are reversed, should be Yeas - Green, Nays - Magenta. In addition to the color fix, let's have the `Yeas` bar on the left and `Nays` on the right

@bugsbunnies Nice catch! I just updated the color and screenshots

mvadari:
The "UNLs"/"non-UNLs" labels look like they're supposed to be labeling each of the bars. Can those titles be moved to the top of each graph?

bugsbunnies:
> The "UNL"/"non-UNL" labels look like they're supposed to be labeling each of the bars. Can those titles be moved to the top of each graph?

I think the problem is that they're at an angle, can we just make it so they center align with each bar group and not at an angle?

mvadari:
> > The "UNL"/"non-UNL" labels look like they're supposed to be labeling each of the bars. Can those titles be moved to the top of each graph?
> 
> I think the problem is that they're at an angle, can we just make it so they center align with each bar group and not at an angle?

Yes, agreed - I think it'd look better with a top label, but I'm not opposed to a bottom axis label.

bugsbunnies:
> > > The "UNL"/"non-UNL" labels look like they're supposed to be labeling each of the bars. Can those titles be moved to the top of each graph?
> > 
> > 
> > I think the problem is that they're at an angle, can we just make it so they center align with each bar group and not at an angle?
> 
> Yes, agreed - I think it'd look better with a top label, but I'm not opposed to a bottom axis label.

I suggested bottom label just to keep it consistent with the basics of a graph with y-axis label and x-axis label in their expected places. The tricky thing with a top label (if im imagining what you're saying correctly) is that if we give them a fixed position on the graph, they might not always be close to the data bars, and if we keep them close to the data, the label could move around depending on the bars height.

mvadari:
Shouldn't the table only show UNL - or have tabs to show UNL vs non-UNL?

pdp2121:
I have fixed the axis accordingly and updated the screenshots. The Simple of this page shows both Yeas(All), Nays(All), Yeas(UNL), and Nays(UNL), while the amendments list page in https://github.com/ripple/explorer/pull/836 only count UNL for voters.

intelliot:
Agree that it's useful to show all validators (instead of just UNL) and it makes sense to show all. 

dUNL validators should be marked with the green check. 

optional - It would also be useful to have a footnote to indicate what the green check means: "this validator is on the vl.ripple.com UNL" (or vl.xrplf.org, whichever is accurate).

pdp2121:
Makes sense! I have moved the UNL validators to the top and added a footnote for clarification. I think this is clear enough to explain those columns.

intelliot (approved):

src/containers/Header/Search.tsx
```diff
@@ -36,9 +38,24 @@ const determineHashType = async (id: string, rippledContext: XrplClient) => {
     await getTransaction(rippledContext, id)
     return 'transactions'
   } catch (e) {
+    const isAmendment = await isAmendmentType(id)
```
ckniffen, line 41: Maybe this should happen in parallel since there are 3 options.  This is even further complicated by #544.  Doing in parallel makes sense to allow for #831. I think this should be moved out into a new PR so these other issues can maybe be addressed there too.

src/containers/Amendment/amendment.scss
```diff
@@ -0,0 +1,138 @@
...
+      margin: 0 10px;
+      color: $green-40;
+      vertical-align: middle;
+    }
+  }
+
+  .votes {
+    margin: 48px 0;
+
+    .barchart {
```
ckniffen, line 79: Consider putting the `max-width` on `.votes`

src/containers/Amendment/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+    },
+  }
+
+  const getErrorMessage = (error: keyof typeof ERROR_MESSAGES | null) =>
+    (error && ERROR_MESSAGES[error]) || ERROR_MESSAGES.default
+
+  const { data, error, isLoading } = useQuery<
+    AmendmentVote,
+    keyof typeof ERROR_MESSAGES | null
+  >(['fetchAmendmentData', identifier], async () => fetchAmendmentData(), {
```
ckniffen, line 61: Should `network` be a part of the key too?

src/containers/Amendment/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const fetchAmendmentData = async (): Promise<AmendmentVote> => {
+    const url = `${process.env.VITE_DATA_URL}/amendment/vote/${network}/${identifier}`
+    return axios
+      .get(url)
+      .then((resp) => resp.data)
+      .then((response) => {
+        if (response.voting_status === 'voting') {
+          return {
+            voting_status: 'voting',
```
ckniffen, line 88: Make voting_status options a constant.

public/locales/en-US/translations.json
```diff
@@ -488,6 +488,30 @@
...
+  "voting": "Voting",
+  "yeas": "yeas",
+  "nays": "nays",
+  "eta": "eta",
+  "consensus": "consensus",
+  "amendment_summary": "Amendment Summary",
+  "not_enabled": "Not Enabled",
+  "enabled": "Enabled",
+  "enabled_on": "enabled on",
+  "yeas_all": "yeas (all)",
```
ckniffen, line 504: these three could be a combination of the two values so you dont need all 4.

jonathanlei (approved):

Issue #714, opened 2023-05-26
Handle protocols inside domain fields


ckniffen:
Some accounts such as issuers of NFTs put fully formed urls in the domain field.  Since we prepend "https://" to the value of that field always this creates an invalid URI.  We should only add "https://" if we do not detect a protocol.

Example: https://livenet.xrpl.org/accounts/r481fwr9G6GqJsnDTRanb9wicGjFf3rfgp/transactions
Component to be updated: https://github.com/ripple/explorer/blob/staging/src/containers/shared/components/DomainLink.tsx

sohammk08:
Hi there! I'm interested in working on this issue. Would it be possible to assign this to me?

ckniffen:
We only assign to main contributors.  You can go ahead and submit a PR though.

sohammk08:
Thanks for commenting in such a short time. I'll start working on it :)

sohammk08:
I tried solving the issue. I require your feedback on the same.
Check it out here:

[https://github.com/ripple/explorer/pull/856](url)

PR #856, opened 2023-10-09
changed the href to fix the problem

sohammk08:

Slightly changed the _href_ in the **DomainLink.tsx** file. Addresses issue #714 

### Context of Change
Some accounts such as the issuers of NFTs put fully formatted urls in the domain field. Since _"https://"_ is prepended to the value of that field, an invalid URI is created. The change adds _"https://"_ only if a protocol is not detected.

### Type of Change

### TypeScript/Hooks Update

## Before / After
**Before:** _"https://"_ was prepended to the value of that field, creating a invalid URL
**After:** Now, _"https://"_ is added only if a protocol is not detected.
## Test Plan

No test plan in practice for the current change.

## Note
This is the example given in the issue: https://livenet.xrpl.org/accounts/r481fwr9G6GqJsnDTRanb9wicGjFf3rfgp/transactions

For some reason, whenever I click the **DOMAIN** on the accounts page(of the example), it leads to:
```
http://https//marketplace-api.onxrp.com/api/metadata/
```

When it should lead to this, instead:
```
https://marketplace-api.onxrp.com/api/metadata/
```

I suspect that there might be a problem with the data being passed to the **DomainLink** component. I have not been able to pinpoint it though.

Further feedback is much appreciated.

Thanks you :)

### DIFF
--- a/src/containers/shared/components/DomainLink.tsx
+++ b/src/containers/shared/components/DomainLink.tsx
@@ -7,16 +7,19 @@ export interface Props {
   domain: string
 }
 
+// Matches a protocol (e.g. 'http://' or 'https://') at the start of a string.
+const protocolRegex = /^([a-z][a-z0-9+\-.]*):\/\//
+
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
 
   // If decode is true, decode the domain
   const decodedDomain = decode ? decodeHex(domain) : domain
 
-  // Check if the decodedDomain contains a protocol (http:// or https://)
-  const domainHasProtocol = decodedDomain.match(/^([a-z][a-z0-9+\-.]*):\/\//)
+  // Use the test method to check for the protocol
+  const domainHasProtocol = protocolRegex.test(decodedDomain)
 
-  // if the decodedDomain already has a protocol, use it; oitherwise, add 'https://'
+  // If decoded domain does not have a protocol, add one ; otherwise, don't
   const href = domainHasProtocol ? decodedDomain : `https://${decodedDomain}`
 
   return (
--- a/src/containers/shared/components/DomainLink.tsx
+++ b/src/containers/shared/components/DomainLink.tsx
@@ -14,7 +14,7 @@ const DomainLink = (props: Props) => {
       className={classnames(`domain`, className)}
       rel="noopener noreferrer"
       target="_blank"
-      href={`https://${decode ? decodeHex(domain) : domain}`}
+      href={decode ? domain : `https://${domain}`}
     >
       {decode ? decodeHex(domain) : domain}
     </a>
--- a/src/containers/shared/components/DomainLink.tsx
+++ b/src/containers/shared/components/DomainLink.tsx
@@ -8,7 +8,7 @@ export interface Props {
 }
 
 // Matches a protocol (e.g. 'http://' or 'https://') at the start of a string.
-const protocolRegex = /^([a-z][a-z0-9+\-.]*):\/\//
+const PROTOCOL_REGEX = /^([a-z][a-z0-9+\-.]*):\/\//
 
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
@@ -17,7 +17,7 @@ const DomainLink = (props: Props) => {
   const decodedDomain = decode ? decodeHex(domain) : domain
 
   // Use the test method to check for the protocol
-  const domainHasProtocol = protocolRegex.test(decodedDomain)
+  const domainHasProtocol = PROTOCOL_REGEX.test(decodedDomain)
 
   // If decoded domain does not have a protocol, add one ; otherwise, don't
   const href = domainHasProtocol ? decodedDomain : `https://${decodedDomain}`
--- a/src/containers/shared/components/DomainLink.tsx
+++ b/src/containers/shared/components/DomainLink.tsx
@@ -9,12 +9,27 @@ export interface Props {
 
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
+
+  // Ensure that "https://" is not added twice
+  const domainWithoutProtocol = domain.replace(/^https?:\/\//, '')
+
+  // If decode is true, decode the domain
+  const decodedDomain = decode
+    ? decodeHex(domainWithoutProtocol)
+    : domainWithoutProtocol
+
+  // Construct the href based on the decoded or original domain
+  const href =
+    decodedDomain.startsWith('http://') || decodedDomain.startsWith('https://')
+      ? decodedDomain
+      : `https://${decodedDomain}`
+
   return (
     <a
-      className={classnames(`domain`, className)}
+      className={classnames('domain', className)}
       rel="noopener noreferrer"
       target="_blank"
-      href={decode ? domain : `https://${domain}`}
+      href={href}
     >
       {decode ? decodeHex(domain) : domain}
     </a>
--- a/src/containers/shared/components/DomainLink.tsx
+++ b/src/containers/shared/components/DomainLink.tsx
@@ -10,19 +10,14 @@ export interface Props {
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
 
-  // Ensure that "https://" is not added twice
-  const domainWithoutProtocol = domain.replace(/^https?:\/\//, '')
-
   // If decode is true, decode the domain
-  const decodedDomain = decode
-    ? decodeHex(domainWithoutProtocol)
-    : domainWithoutProtocol
+  const decodedDomain = decode ? decodeHex(domain) : domain
 
-  // Construct the href based on the decoded or original domain
-  const href =
-    decodedDomain.startsWith('http://') || decodedDomain.startsWith('https://')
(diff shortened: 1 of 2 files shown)


src/containers/shared/components/DomainLink.tsx
```diff
@@ -14,7 +14,7 @@ const DomainLink = (props: Props) => {
       className={classnames(`domain`, className)}
       rel="noopener noreferrer"
       target="_blank"
-      href={`https://${decode ? decodeHex(domain) : domain}`}
+      href={decode ? domain : `https://${domain}`}
```
ckniffen, line 17: If the decode property is set then we need to decode the domain which is always in hexadecimal.  The issue is that the decoded value sometimes starts with the protocol like `https://`.  If that is case we need not add http `https://`

ckniffen: The reason why the tests are failing is that after your change when you specify decode it is outputting the domain in hex.

You also need to add a new test that when the domain contains the protocol it needs to not double the protocol.

sohammk08: Yes. Thanks for the follow up. I'll revise it.

sohammk08: I added the test as such:
```typescript
  it('handles domain link with domain containing protocol', () => {
    const wrapper = mount(<DomainLink domain="https://example.com" />)
    expect(wrapper.find('a').props().className).toEqual('domain')
    expect(wrapper.find('a').text()).toEqual('https://example.com')
    expect(wrapper.find('a').props().href).toEqual('https://example.com')
    wrapper.unmount()
  })
```
Would the above be viable?

ckniffen: You actually need to test the input as coming in encoded as HEX.

```
const wrapper = mount(<DomainLink domain=" 68747470733A2F2F6578616D706C652E636F6D" />)
```

sohammk08: I've changed it as suggested.
Take a look:
```typescript
  it('handles domain link with domain provided in HEX-encoded format', () => {
    const wrapper = mount(
      <DomainLink domain="68747470733A2F2F6578616D706C652E636F6D" decode />,
    )
    // The HEX-encoded input should be decoded to 'example.com'
    expect(wrapper.find('a').props().className).toEqual('domain')
    expect(wrapper.find('a').text()).toEqual('https://example.com')
    expect(wrapper.find('a').props().href).toEqual('https://example.com')
    wrapper.unmount()
  })

```

sohammk08:
I ran it on my machine and it passes all the tests without raising any additional issues. Please take a look.

src/containers/shared/components/DomainLink.tsx
```diff
@@ -9,12 +9,27 @@ export interface Props {
 
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
+
+  // Ensure that "https://" is not added twice
+  const domainWithoutProtocol = domain.replace(/^https?:\/\//, '')
```
mvadari, line 14: Wouldn't it be easier to just check if the domain has `"http"` and avoid adding it if it does?

sohammk08: ### Yes, it would be easier to _just check if the domain has "https://" and avoid adding it if it does._

- ### However, if we go by that method, the **5th test** keeps failing:
The test:
```typescript
  it('handles domain link with domain provided in HEX-encoded format', () => {
    const wrapper = mount(
      <DomainLink decode domain="68747470733A2F2F6578616D706C652E636F6D" />,
    )
    expect(wrapper.find('a').props().className).toEqual('domain')
    expect(wrapper.find('a').text()).toEqual('https://example.com')
    expect(wrapper.find('a').props().href).toEqual('https://example.com')
    wrapper.unmount()
  })
```

- ### Feedback / test output after running the test (after implementing - _check if it already has domain and avoid adding one if it does_):
The output from the tests/console:
```
● DomainLink › handles domain link with domain provided in HEX-encoded format

    expect(received).toEqual(expected) // deep equality

    Expected: "https://example.com"
    Received: "https://https://example.com"
```

I attempted to implement your suggested approach, but encountered a persisting issue with the **5th test** failing. I found that only when I include the code for _Ensure that "https://" is not added twice_ does the **5th test** pass consistently, without causing any further complications or errors.

- ### _Ensure that "https://" is not added twice:_
The code:
```typescript
  // Ensure that "https://" is not added twice
  const domainWithoutProtocol = domain.replace(/^https?:\/\//, '')
```

mvadari: I was imagining something like this:
```js
const decodedDomain = decode ? decodeHex(domain) : domain
const domainHasProtocol = decodedDomain.match((/^https?:\/\//)
const href = domainHasProtocol ? decodedDomain : `https://${decodedDomain}`
```

ckniffen: One that would include all protocols would be `/^([a-z][a-z0-9+\-.]*):\/\//`

src/containers/shared/components/DomainLink.tsx
```diff
@@ -9,12 +9,22 @@ export interface Props {
 
 const DomainLink = (props: Props) => {
   const { className, decode = false, domain } = props
+
+  // If decode is true, decode the domain
+  const decodedDomain = decode ? decodeHex(domain) : domain
+
+  // Check if the decodedDomain contains a protocol (http:// or https://)
+  const domainHasProtocol = decodedDomain.match(/^([a-z][a-z0-9+\-.]*):\/\//)
```
ckniffen, line 17: nit: I think test might be faster and you should define the regex outside the component so it doesn't have to be compiled at every usage.

ckniffen (approved):

src/containers/shared/components/DomainLink.tsx
```diff
@@ -7,14 +7,27 @@ export interface Props {
   domain: string
 }
 
+// Matches a protocol (e.g. 'http://' or 'https://') at the start of a string.
+const protocolRegex = /^([a-z][a-z0-9+\-.]*):\/\//
```
mvadari, line 11:
```suggestion
const PROTOCOL_REGEX = /^([a-z][a-z0-9+\-.]*):\/\//
```
Since it's a global constant

src/containers/shared/components/DomainLink.tsx
```diff
@@ -7,14 +7,27 @@ export interface Props {
...
   const { className, decode = false, domain } = props
+
+  // If decode is true, decode the domain
+  const decodedDomain = decode ? decodeHex(domain) : domain
+
+  // Use the test method to check for the protocol
+  const domainHasProtocol = protocolRegex.test(decodedDomain)
+
+  // If decoded domain does not have a protocol, add one ; otherwise, don't
+  const href = domainHasProtocol ? decodedDomain : `https://${decodedDomain}`
```
mvadari, line 23: @ckniffen should this be `http://` or `https://`? What if a domain doesn't have HTTPS enabled?

src/containers/shared/components/test/DomainLink.test.tsx
```diff
@@ -38,4 +38,13 @@ describe('DomainLink', () => {
...
   })
+  it('handles domain link with domain provided in HEX-encoded format', () => {
+    const wrapper = mount(
+      <DomainLink decode domain="68747470733A2F2F6578616D706C652E636F6D" />,
+    )
+    expect(wrapper.find('a').props().className).toEqual('domain')
+    expect(wrapper.find('a').text()).toEqual('https://example.com')
+    expect(wrapper.find('a').props().href).toEqual('https://example.com')
+    wrapper.unmount()
+  })
```
mvadari, line 49: This is a bit more understandable:
```suggestion
    const url = 'https://example.com'
    const urlHex = encodeHex(url)
    const wrapper = mount(
      <DomainLink decode domain={urlHex} />,
    )
    expect(wrapper.find('a').props().className).toEqual('domain')
    expect(wrapper.find('a').text()).toEqual(url)
    expect(wrapper.find('a').props().href).toEqual(url)
    wrapper.unmount()
  })
```

sohammk08: As for the additional test that is created in the _DomainLink.test.tsx_, I just followed the format of the previous tests in the same test file. If only this test is changed, the file will be non-uniform. If you still want to change it the way you mentioned above, let me know.

src/containers/shared/components/DomainLink.tsx
sohammk08: Implemented it in the new commit _Refactor: Improved code readability and consistency_ (https://github.com/ripple/explorer/pull/856/commits/53132ca2cc21119414b489a8fd65fc175e3c46ba)

src/containers/shared/components/test/DomainLink.test.tsx
sohammk08: Changed it for better readability: https://github.com/ripple/explorer/pull/856/commits/b8fc786640d384fc3d893dca5b4410f6f4ad74a6

mvadari (approved):

PR #858, opened 2023-10-17
fix: keep language names consistent

ckniffen:

The display text for language in the language picker will remain consistent regardless of which language is currently selected.

- Switch to using a map of language codes to language names
- Remove language specific entries for language names
- Update guide on defining languages

Fixes: #807
Alternative To: #809

### Type of Change

- [x] Translation Updates

## Before
Language Picker - Before

## After
Language Picker - After


ckniffen:
@develoQ Is this what you envisioned.

ckniffen:
> @develoQ Is this what you envisioned?

Confirmed directly that this is what is desired.

khancode (approved):

pdp2121 (approved):

PR #876, opened 2023-11-17
docs: add process for updating translations

### LINKED ISSUE #873: Document process for updating translation files when there is a change
Document a policy in the adding language docs.

ckniffen:

Make translations documentation more generic and add process for adding/updating entries.

Closes #873

### Type of Change

- [x] Documentation Updates


docs/translating.md
```diff
@@ -0,0 +1,24 @@
...
+1. Add the new language code and language name to the map `supportedLanguages` in [/src/i18n/baseConfig.ts](../src/i18n/baseConfig.ts)
+2. Create a folder in [/public/locales/](../public/locales) with the language code as its name.
+3. Add a new file, `translations.json`, to the new folder.
+4. Translate all the entries.  If you prefer to use the English version just set the value to `null`.
+
+   Example file: [/public/locales/ja-JP/translations.json](../public/locales/ja-JP/translations.json)
+
+## Updating Existing Translations
+
+When making changes to the base (English) language file follow the ?
```
JST5000, line 14: What does the "?" mean?

docs/translating.md
```diff
@@ -0,0 +1,24 @@
...
+
+### New Entry
+1. Create entry in `en-US/translations.json`
+2. Add an entry to all other `translation.json` files with a value of `null`
```
JST5000, line 18: This phrasing is a bit ambiguous - it could mean only add an entry if there is a value of null in the other translation.json files. (Was a bit confused when I read it, even though with an understanding of what's going on that other interpretation doesn't make sense. Hopefully this makes it more clear though)
```suggestion
2. Add an entry to all other `translation.json` files. Set the value to be `null` (which means it will default to using the English version)
```

docs/translating.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Existing Entry
+1. Update the entry in `en-US/translations.json`
+2. Update other `translation.json` files
+   - Leave entry as is if the string does not change materially. ex. "Please check your transaction hash" => "Please check your transaction hash or CTID."
+   - Set the value to `null` in all other language files to fallback to new English string
```
JST5000, line 24: Since similar content doesn't require any further action, I recommend updating the steps to "lead with action / decisions" (and then provide the additional context afterwards). That way key steps aren't lost in the sub-bullet section.
```suggestion
2. If the entry has a new meaning than before, set the value to `null` in all other `translation.json` files. 
  - If the meaning does NOT change materially, you can leave them as is. Ex. "Please check your transaction hash" => "Please check your transaction hash or CTID."

*Note that changing a translation to `null` causes it to fallback to the English version, and is a signal to language contributors that they may want to provide a fresh localization.*
```

ckniffen: Addressed in 3adc028777c037d5e972b557ebd0f75e63fffbec.

docs/translating.md
ckniffen: Addressed in 3adc028777c037d5e972b557ebd0f75e63fffbec.

docs/translating.md
ckniffen: Addressed in 3adc028777c037d5e972b557ebd0f75e63fffbec.

JST5000 (approved):
LGTM

christianlantin:
Master

christianlantin (commented):
Master

PR #899, opened 2023-12-21
fix: display uppercase token when uppercase token is searched

mvadari:

Title says it all.

### Context of Change

It was a mild annoyance when looking up a token.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

When an uppercase token is searched in the search bar, it goes to the uppercase token page instead of the lowercase token page.

## Test Plan

CI passes. Test was added.


ckniffen (approved):

christianlantin:
Master

JST5000 (approved):
LGTM

PR #908, opened 2024-01-02
feat: add loader to transactions section of homepage

mvadari:

This PR adds a loader to the ledger details displayed on the homepage, for when the transaction info takes a bit longer to load. It also adds a Σ to the total fees in the ledger, so it's clearer that it's not the average or base fee.

### Context of Change

With the heavy load happening right now, some ledgers take a second or two to load all of the transactions. This makes it clearer that there isn't anything that's broken.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

image
image

## Test Plan

Works locally.


ckniffen (approved):

JST5000 (approved):
LGTM

Issue #684, opened 2023-04-24
Validator page stuck in infinite loop when ledger not found


ckniffen:
When the last ledger_index of a validator is missing from the connected node the validator page enters into an infinite loop.

https://testnet.xrpl.org/validators/n9Krc7SYztNiDvGLdXYzyPLwUCkWKPaptxX9bgCYbmKVFzL33JkH

PR #917, opened 2024-01-11
fix: validator page validation vs ledger_closed race condition

jonathanlei:

See issue https://github.com/ripple/explorer/issues/684

### Context of Change
the validator page is stuck when retrieving info on a validator that has validated a ledger which doesn't have a state of validated or cannot be found, resulting in a "Something bad happened" page, see image:
image

Solution:
add error catching and log the error, return response anyway without last updated time and ledger index 

Screenshot 2024-01-10 at 3 08 18 PM

### Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)

## Test Plan

added test in Validator.test.js


src/containers/Validators/test/Validator.test.js
```diff
@@ -159,4 +159,42 @@ describe('Validator container', () => {
...
+
+    expect(getLedger).toBeCalledWith('12345', undefined)
+    expect(document.title).toBe('Validator test.example.com')
+    // test ledger-time isn't updated
+    const lastLedgerDateTime = wrapper.find(`[data-test="ledger-time"]`)
+    expect(lastLedgerDateTime).not.toExist()
+    // test ledger-index stays the same
+    const lastLedgerIndx = wrapper.find(`[data-test="ledger-index"]`)
+    expect(lastLedgerIndx).toExist()
+    const linkText = lastLedgerIndx.find('.value').text()
```
ckniffen, line 196: You can do `expect(lastLedgerIndx.find('.value')).toHaveText()`

src/containers/Validators/index.tsx
```diff
@@ -24,6 +24,9 @@ import NetworkContext from '../shared/NetworkContext'
 import { VALIDATOR_ROUTE } from '../App/routes'
 import { buildPath, useRouteParams } from '../shared/routing'
 import { VotingTab } from './VotingTab'
+import logger from '../../rippled/lib/logger'
+
+const log = logger({ name: 'ledgers' })
```
ckniffen, line 29: I would update the name to be more specific to this page. Maybe just `validator`.

ckniffen (approved):

src/containers/Validators/test/Validator.test.js
```diff
@@ -159,4 +159,41 @@ describe('Validator container', () => {
...
+
+    expect(getLedger).toBeCalledWith('12345', undefined)
+    expect(document.title).toBe('Validator test.example.com')
+    // test ledger-time isn't updated
+    const lastLedgerDateTime = wrapper.find(`[data-test="ledger-time"]`)
+    expect(lastLedgerDateTime).not.toExist()
+    // test ledger-index stays the same
+    const lastLedgerIndx = wrapper.find(`[data-test="ledger-index"]`)
+    expect(lastLedgerIndx).toExist()
+    expect(lastLedgerIndx.find('.value')).toHaveText('12345')
```
JST5000, line 196: nit: Typo
```suggestion
    const lastLedgerIndex = wrapper.find(`[data-test="ledger-index"]`)
    expect(lastLedgerIndex).toExist()
    expect(lastLedgerIndex.find('.value')).toHaveText('12345')
```

JST5000 (approved):
LGTM - just one typo

jonathanlei:
> LGTM - just one typo

oh I thought it's okay to use abbreviation haha i've used indx in other code before

JST5000:
> oh I thought it's okay to use abbreviation haha i've used indx in other code before

It's not the end of the world, but it only saves a letter & can trip someone up when typing, so I figure writing it all out is better :)

PR #919, opened 2024-01-18
refactor: Tooltip to hooks and Typescript

ckniffen:

Update `<Tooltip>` component to use hooks and Typescript.

### Type of Change

### TypeScript/Hooks Update


jonathanlei (approved):
code logic LGTM, could you add a screenshot on what it would look like?

ckniffen:
> code logic LGTM, could you add a screenshot on what it would look like?

I will, but no visuals changed

jonathanlei:
oh nevermind it's a refactor...

src/containers/shared/components/Tooltip.tsx
```diff
@@ -0,0 +1,115 @@
...
+    return <></>
+  }
+
+  const { data } = tooltip
+
+  const renderNegativeUnlTooltip = () =>
+    data.nUnl.map((key) => {
+      const short = key.substr(0, 8)
+      return <div key={key} className={`nUnl: ${key}`}>{`${short}...`}</div>
+    })
```
mvadari, line 41: Should the tooltip be refactored to contain this data as a child element as well?

ckniffen: Yeah it could be done further.  Wanted to have this impact the least amount of spots.  This is because I have a future PR that will provide a tooltip hook so that I can properly separate the homepage into components.

ckniffen: Making it a sub component is tricky too because it means the tooltip will be rendered ahead of time.  Another solution is to have a property called content which takes a React component.

JST5000 (approved):
LGTM

PR #920, opened 2024-01-18
refactor: Notification to Typescript

ckniffen:

One of the last remaining places requiring a refactor

### Type of Change

### TypeScript/Hooks Update


src/containers/shared/components/Notification/index.tsx
```diff
@@ -1,27 +1,35 @@
...
-const VALID_USAGES = [
-  'default',
-  'success',
-  'warning',
-  'danger',
-  'dark',
-  'light',
-  'dark50',
-]
+type NotificationLevel = 'primary'
```
JST5000, line 4: If this is just a single string, why is this not a constant? (Or just included in the component as a hard-coded value?)

ckniffen: That is a good question.  The previous prop type was `level: PropTypes.oneOf([PRIMARY])` so this is equivalent.  Currently it is the only valid option but could be expanded later via like `'primary' | 'secondary'`.

jonathanlei (approved):

JST5000 (approved):

PR #921, opened 2024-01-20
refactor: update homepage to typescript and hooks

ckniffen:

Update homepage to typescript and hooks and break out aspects of the homepage's scrolling ledgers into smaller components. This also adds a new hook `useTooltip` that cleans up tooltip logic when many places are updating tooltips.

A small change to move the transaction iteration into its own component reduced render times of the homepage due by between 30ms to 60ms each re-render.

### Type of Change

### TypeScript/Hooks Update

## Future Work

Optimize the <Streams> component into a hook that causes way fewer re-renders.


src/containers/Ledgers/LedgerEntryTransaction.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { getAction, getCategory } from '../shared/components/Transaction'
+import { TRANSACTION_ROUTE } from '../App/routes'
+import { TransactionActionIcon } from '../shared/components/TransactionActionIcon/TransactionActionIcon'
+import { RouteLink } from '../shared/routing'
+import { useTooltip } from '../shared/components/Tooltip'
+
+export const LedgerEntryTransaction = ({
+  transaction,
+}: {
+  transaction: any
```
mvadari, line 11: This type should be improved

src/containers/Ledgers/LedgerListEntry.tsx
```diff
@@ -0,0 +1,74 @@
...
+        {/* Render Total Fees (can be 0) */}
+        {ledger.total_fees !== undefined && (
+          <div className="fees">
+            {SIGMA} {t('fees')}:
+            <b>
+              <Amount value={{ currency: 'XRP', amount: ledger.total_fees }} />
+            </b>
+          </div>
+        )}
+        <LedgerEntryTransactions transactions={ledger.transactions} />
```
mvadari, line 60:
```suggestion
        <LedgerEntryTransactions transactions={ledger.transactions || []} />
```

ckniffen: I choose not to always make it an array to differentiate between 0 transactions and it not loaded in.

src/containers/Ledgers/LedgerEntryTransaction.tsx
ckniffen: Agreed.

khancode (approved):
LGTM 👍

src/containers/Ledgers/LedgerEntryHashTrustedCount.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { t } = useTranslation()
+  const { hideTooltip, showTooltip } = useTooltip()
+  const className = hash.trusted_count < (unlCount || 0) ? 'missed' : ''
+
+  const getMissingValidators = () => {
+    const unl = {}
+
+    Object.keys(validators).forEach((pubkey) => {
+      if (validators[pubkey].unl) {
+        unl[pubkey] = false
```
JST5000, line 23: Why do we set this to false instead of the unl value? It seems similar to `undefined` and I took a second to re-read to confirm that `false` just seems to be a sentinel value that's main trait is that it is NOT undefined?

ckniffen: I honestly dont know this code was written this way before I just broke it out into its own component.   I did rewrite it in the streams refactor though to this 

```
const status = useMemo(() => {
    const missing = [...(unl || [])]
    validations.forEach((v) => {
      if (!validators?.[v.validation_public_key]) {
        return
      }

      const missingIndex = missing.findIndex(
        (assumedMissing) => assumedMissing === v.validation_public_key,
      )
      if (missingIndex !== -1) {
        missing.splice(missingIndex, 1)
      }
    })

    return {
      missing: missing.map((v) => validators?.[v]),
      trustedCount: (unl?.length || 0) - missing.length,
    }
  }, [unl, validations])
  ```

File: src/containers/Ledgers/LedgerMetrics.tsx [hunk 4]
```diff
@@ -0,0 +1,120 @@
...
+  txn_sec: '--',
+  txn_ledger: '--',
+  ledger_interval: '--',
+  avg_fee: '--',
+  quorum: '--',
+  nUnl: [],
+}
+
+export const LedgerMetrics = ({
+  data: suppliedData,
```
JST5000, line 21: I don't see where `suppliedData` is defined

src/containers/Ledgers/LedgerMetrics.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onPause,
+  paused,
+}: {
+  data: any
+  onPause: any
+  paused: boolean
+}) => {
+  const data = { ...DEFAULTS, ...suppliedData }
```
JST5000, line 29: Doesn't this cause a name collision with the input parameter `data` for LedgerMetrics? I'm a bit confused on the naming here.

src/containers/Ledgers/LedgerMetrics.tsx
```diff
@@ -0,0 +1,120 @@
...
+  if (data.load_fee === '--') {
+    data.load_fee = data.base_fee || '--'
+  }
+  delete data.base_fee
+  const items = Object.keys(data)
+    .map((key) => {
+      let content: any = null
+
+      let className = 'label'
+      if (data[key] === undefined && key !== 'nUnl') {
```
JST5000, line 62: Can you add a comment explaining why `nUnl` is special here? It's getting a lot of special treatment in this if/else - maybe there's a better way to separate out that logic? (ie. similar to how you deal with `base_fee` beforehand then delete it from the object?)

src/containers/Ledgers/LedgerMetrics.tsx
```diff
@@ -0,0 +1,120 @@
...
+            onFocus={() => {}}
+            onBlur={() => {}}
+            onMouseOver={(e) => showTooltip('nUnl', e, { nUnl: data.nUnl })}
+            onMouseOut={() => hideTooltip()}
+            tabIndex={0}
+            key={key}
+          >
+            <a
+              key={`link ${key}`}
+              href="https://xrpl.org/negative-unl.html"
```
JST5000, line 86: Should this link be "localized" to point to the proper language if the explorer is in Japanese for example?

JST5000 (approved):
One open question, but otherwise looks good to me :)

src/containers/Ledgers/LedgerMetrics.tsx
ckniffen: `data` is the props name but the linter prevents redefining a prop so i name it differently locally scoped to the function to make it shorter instead of having a longer name ever where i reference all the metrics

src/containers/Ledgers/LedgerMetrics.tsx
ckniffen: See previous comment

src/containers/Ledgers/LedgerMetrics.tsx
ckniffen: We don't link anywhere else based on the language.  I wouldn't know where to start since every site might have different rules like localstorage based or path based.

PR #922, opened 2024-01-23
Convert Validator's History tab dates to UTC with note

pdp2121:

Convert the date in History tab into UTC to ensure consistency for users in different timezones.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

## Before / After

Date column header will change to inform users:

Screenshot 2024-01-24 at 11 40 49 AM

### DIFF
--- a/src/containers/Validators/historyTab.scss
+++ b/src/containers/Validators/historyTab.scss
@@ -92,14 +92,3 @@
     color: $orange-50;
   }
 }
-
-.history-note {
-  overflow: hidden;
-  color: $black-40;
-  font-size: 12px;
-  letter-spacing: 0px;
-  line-height: 20px;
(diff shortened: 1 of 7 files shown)


src/containers/Validators/historyTab.scss
```diff
@@ -92,3 +92,14 @@
     color: $orange-50;
   }
 }
+
+.history-note {
```
ckniffen, line 96: Are there existing styles that could apply? 

Try using the truncate mixin to remove some of these properties

ckniffen:
Instead of adding the not you could just add UTC next to the date like we do elsewhere in the app.

ckniffen (approved):

jonathanlei (approved):

PR #939, opened 2024-02-14
fix: add better error handling for transactions' Simple pages

mvadari:

This PR has better error handling if there is a bug in a transaction's `Simple` page. This results in better user experience when something is broken.

This does add a package, `react-error-boundary`, but the package is very tiny and is quite useful.

### Context of Change

I encountered this while using the Explorer.

### Type of Change

### TypeScript/Hooks Update

N/A

## Before / After

Before: a black screen

After:
image


JST5000 (approved):
LGTM

pdp2121:
@bugsbunnies Should the message be something similar to the `<NoMatch />` component for cohesiveness?

Screenshot 2024-02-15 at 3 59 16 PM

mvadari:
> @bugsbunnies Should the message be something similar to the `<NoMatch />` component for cohesiveness?

I copied the old error message/CSS that was displayed in the `Simple` page when it wasn't supported for an unknown transaction, before we added support for a default `Simple` page.

pdp2121:
> > @bugsbunnies Should the message be something similar to the `<NoMatch />` component for cohesiveness?
> 
> I copied the old error message/CSS that was displayed in the `Simple` page when it wasn't supported for an unknown transaction, before we added support for a default `Simple` page.

Make sense!

bugsbunnies:
> @bugsbunnies Should the message be something similar to the `<NoMatch />` component for cohesiveness?
> 
> Screenshot 2024-02-15 at 3 59 16 PM

Yeah that makes sense

mvadari:
> > @bugsbunnies Should the message be something similar to the `<NoMatch />` component for cohesiveness?
> 
> Yeah that makes sense

IMO the current error message is better, because this would exclusively be due to an Explorer error, rather than user error. But I'm happy to edit the CSS if you'd prefer that.

mvadari:
image

PR #945, opened 2024-02-22
test: begin the switch to RTL

mvadari:
### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### TypeScript/Hooks Update

- [ ] Updated files to React Hooks
- [ ] Updated files to TypeScript

PR #972, opened 2024-04-09
fix: fix LedgersPage tests

mvadari:

This PR fixes the `LedgersPage` tests so they stop failing intermittently/in some PRs (like #945).

### Context of Change

The `LedgersPage` tests fail for me locally.
image

### Type of Change

### TypeScript/Hooks Update

N/A - too complicated, not worth doing here.

## Test Plan

Tests pass locally now.

### DIFF
--- a/src/containers/shared/components/Streams.jsx
+++ b/src/containers/shared/components/Streams.jsx
@@ -143,8 +143,8 @@ class Streams extends Component {
     }, 100)
   }
 
-  componentDidMount() {
-    this.connect()
+  async componentDidMount() {
+    await this.connect()
     this.updateNegativeUNL()
     if (process.env.VITE_ENVIRONMENT !== 'custom') {
       this.updateMetricsFromServer()
@@ -179,8 +179,8 @@ class Streams extends Component {
   }
 
   // handle ledger messages
-  handleLedger(data) {
-    const ledger = this.addLedger(data)
+  async handleLedger(data) {
+    const ledger = await this.addLedger(data)
     const {
       ledger_hash: ledgerHash,
       ledger_index: ledgerIndex,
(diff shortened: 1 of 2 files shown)


src/containers/shared/components/Streams.jsx
```diff
@@ -330,6 +330,12 @@ class Streams extends Component {
     })
   }
 
+  async setStateAsync(state) {
```
justinr1234, line 333: Is there a reason to not use other lifecycle methods to manage this? If the component unmounts you could end up with react errors

mvadari: What do you mean by other lifecycle methods? I didn't really want to get into a full refactor right now.
Honestly, this file is already sort of a mess (and probably has a ton of little bugs) and @ckniffen has a WIP refactor of all of this to use TypeScript/hooks.

justinr1234: Why didn't the functional version of setState in the existing code not already work?

mvadari: It did, but it resulted in errors in local tests. I'm not really sure why it worked in CI but not locally, but I googled around and this seemed to fix it.

mvadari: Yes.
https://github.com/ripple/explorer/blob/staging/src/containers/shared/components/Streams.jsx#L401-L415
Lines 414-415 would execute before the state was finished updating. According to some StackOverflow research, this is the best way to resolve this.

justinr1234 (approved):

src/containers/shared/components/Streams.jsx
justinr1234: Ok makes sense, thanks

anissa-ripple (approved):

pdp2121 (approved):

PR #640, opened 2023-03-22
refactor: network pages ts, hooks, and no redux

ckniffen:
## High Level Overview of Change

- `<Hexagons>` remove redux and switch to hooks
- `<Map>` remove redux and switch to hooks and ts
- `<Nodes>` use `isEarlierVersion` instead of one off compare

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)

### TypeScript/Hooks Update

- [x] Updated files to React Hooks
- [x] Updated files to TypeScript

## Future Tasks

Remove countries.json from the javascript bundle and load it separately only on the nodes page.  Will save 39kb gzipped.

PR #634, opened 2023-03-20
refactor: remove redux from validator page

ckniffen:
## High Level Overview of Change

Leverage `useWindowSize` instead.

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)

PR #984, opened 2024-05-20
refactor: remove the use of redux

mvadari:

This PR removes the last of redux from the codebase. The only two pieces left were Account and Token.

It also removes all mentions of gravatars.

This also reduces the size of the build by about 35kb.

There were some build issues that required regenerating the package-lock, which is almost all of the diff.

### Context of Change

* https://github.com/ripple/explorer/pull/649
* https://github.com/ripple/explorer/pull/640
* https://github.com/ripple/explorer/pull/635
* https://github.com/ripple/explorer/pull/634
* https://github.com/ripple/explorer/pull/1100

### Type of Change

### TypeScript/Hooks Update

N/A

## Test Plan

Tests were fixed to handle the refactor.


src/containers/App/test/App.test.jsx
```diff
@@ -270,6 +262,12 @@ describe('App container', () => {
           event: 'screen_view',
           network: 'mainnet',
         },
+        {
```
mvadari, line 265: I'm not sure why these test outputs changed. This should be investigated before this PR is merged.

ckniffen (approved):

achowdhry-ripple (approved):

PR #1000, opened 2024-06-12
fix: Upgrade Status renders on data error

pdp2121:

The VHS is experiencing an intermittent issue with CORS header that causes occasional disruption in data fetching for the Upgrade Status chart. This PR will add checks so to ensure the chart renders even when there's a disruption.  

Screenshot 2024-06-12 at 4 13 54 PM

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### DIFF
--- a/src/containers/Network/UpgradeStatus.tsx
+++ b/src/containers/Network/UpgradeStatus.tsx
@@ -22,6 +22,7 @@ import {
 } from '../shared/vhsTypes'
 import NetworkContext from '../shared/NetworkContext'
 import { ledgerCompare } from './Nodes'
+import { Loader } from '../shared/components/Loader'
 
 interface DataAggregation {
   label: string
@@ -154,11 +155,15 @@ export const UpgradeStatus = () => {
           newValidatorList[validator.signing_key] = validator
         })
 
-        setVList(newValidatorList)
-        setUnlCount(
-          validators.filter((validator) => Boolean(validator.unl)).length,
-        )
-        return Object.values(newValidatorList)
+        if (validators != null) {
+          setVList(newValidatorList)
+          setUnlCount(
+            validators.filter((validator) => Boolean(validator.unl)).length,
+          )
+          return Object.values(newValidatorList)
+        }
+
+        return null
       })
       .catch((e) => Log.error(e))
 
@@ -192,7 +197,9 @@ export const UpgradeStatus = () => {
       })
       .catch((e) => Log.error(e))
     Promise.all([validatorsReq, nodesReq]).then(([validators, nodes]) => {
-      setAggregated(aggregateData(validators, nodes))
+      if (validators != null && nodes != null) {
+        setAggregated(aggregateData(validators, nodes))
+      }
     })
   }
 
@@ -250,9 +257,13 @@ export const UpgradeStatus = () => {
       </div>
       <div className="wrap">
         <NetworkTabs selected="upgrade-status" />
-        <div className="upgrade_status">
-          <BarChartVersion data={aggregated} stableVersion={stableVersion} />
-        </div>
+        {aggregated ? (
+          <div className="upgrade_status">
+            <BarChartVersion data={aggregated} stableVersion={stableVersion} />
+          </div>
+        ) : (
+          <Loader />
+        )}
       </div>
     </div>
   )


mvadari (commented):
Should we add a testcase too?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -154,11 +155,15 @@ export const UpgradeStatus = () => {
           newValidatorList[validator.signing_key] = validator
         })
 
-        setVList(newValidatorList)
-        setUnlCount(
-          validators.filter((validator) => Boolean(validator.unl)).length,
-        )
-        return Object.values(newValidatorList)
+        if (validators != null) {
```
mvadari, line 158: Shouldn't this always be update even if `validators` is null?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -192,7 +197,9 @@ export const UpgradeStatus = () => {
       })
       .catch((e) => Log.error(e))
     Promise.all([validatorsReq, nodesReq]).then(([validators, nodes]) => {
-      setAggregated(aggregateData(validators, nodes))
+      if (validators != null && nodes != null) {
```
mvadari, line 200: ditto here

src/containers/Network/UpgradeStatus.tsx
pdp2121: `validators` can only be `null` if the VHS link is broken, otherwise it would be an array (can be empty). Not sure what you meant by handles updates

src/containers/Network/UpgradeStatus.tsx
pdp2121: Same here, this will only update if the links work. Otherwise it will use old data

pdp2121:
I will add test cases

pdp2121:
@mvadari I have updated the logic. Aggregation for nodes and validators will be handled separately (in case one fails, it would still update the other).

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,119 @@ import {
...
     if (version) {
       if (!aggregation[version]) {
-        aggregation[version] = { validatorsCount: 0, nodesCount: 0 }
+        aggregation[version] = { validatorsCount: 0, validatorsPercent: 0 }
       }
       aggregation[version].validatorsCount += 1
     }
   })
+  for (const label in aggregation) {
+    if (Object.prototype.hasOwnProperty.call(aggregation, label)) {
```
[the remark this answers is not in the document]
pdp2121: It's an eslint complaint:
The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype.eslint[guard-for-in](https://eslint.org/docs/latest/rules/guard-for-in)
I will look for a cleaner fix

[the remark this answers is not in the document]
pdp2121: Used for of instead

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,115 @@ import {
...
     if (version) {
       if (!aggregation[version]) {
-        aggregation[version] = { validatorsCount: 0, nodesCount: 0 }
+        aggregation[version] = { nodesCount: 0, nodesPercent: 0 }
       }
       aggregation[version].nodesCount += 1
     }
   })
+  for (const label in aggregation) {
+    if (Object.prototype.hasOwnProperty.call(aggregation, label)) {
```
mvadari, line 86: Can you fix it in the couple of other places as well?

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,115 @@ import {
 } from '../shared/vhsTypes'
 import NetworkContext from '../shared/NetworkContext'
 import { ledgerCompare } from './Nodes'
+import { Loader } from '../shared/components/Loader'
 
-interface DataAggregation {
-  label: string
-  validatorsPercent: number
-  validatorsCount: number
+interface NodesStats {
```
mvadari, line 27: These types are a bit harder to read with the first word pluralized, and I don't think the pluralization adds anything - can you remove it? i.e. `NodeStats` instead of `NodesStats`.

pdp2121:
Not sure why `npm run test:ci` succeed locally but fails on GitHub

mvadari:
> Not sure why `npm run test:ci` succeed locally but fails on GitHub

Usually due to flakiness or something, or missing an `async/await` somewhere.

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,108 @@ import {
...
+  validatorsCount: number
+}
 
-  let totalVals = 0
-  let totalNodes = 0
-  interface aggregationTypes {
-    validatorsCount: number
-    nodesCount: number
-  }
+interface ValidatorsAggregation {
```
mvadari, line 37:
```suggestion
interface ValidatorAggregation {
```

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,108 @@ import {
...
+  [label: string]: ValidatorStats
+}
+
+interface NodesAggregation {
```
mvadari, line 41: Easier to read when singular
```suggestion
interface NodeAggregation {
```

src/containers/Network/UpgradeStatus.tsx
```diff
@@ -22,63 +22,108 @@ import {
...
+  return aggregation
+}
+
+export const aggregateData = (
+  validatorsAggregation: ValidatorsAggregation,
+  nodesAggregation: NodesAggregation,
+): DataAggregation[] => {
+  const combinedAggregation: { [label: string]: ValidatorStats & NodeStats } =
+    {}
+  for (const label of Object.keys(validatorsAggregation)) {
```
mvadari, line 99: ditto with these variable names

mvadari (approved):

anissa-ripple (approved):

PR #1057, opened 2024-10-10
Add metadata styling for Multi Purpose Token

pdp2121:

MPTokenMetadata is an essential part of a MPT token as it allows issuers to define their own metadata standards, tailored to their specific requirements, encompassing details like asset codes, common names, symbols, or even visual representations.

This PR formats the display of this metadata field in the Explorer by adding json formatter for readability.

### Context of Change

Spec: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033d-multi-purpose-tokens

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

## Before / After

MPTokenIssuanceCreate Page:

Screenshot 2024-10-10 at 3 53 01 PM

MPT Page:
Screenshot 2024-10-10 at 3 53 20 PM

### DIFF
--- a/src/containers/MPT/MPTHeader/Details.tsx
+++ b/src/containers/MPT/MPTHeader/Details.tsx
@@ -1,9 +1,10 @@
 import { useTranslation } from 'react-i18next'
 import './styles.scss'
 import { useLanguage } from '../../shared/hooks'
-import { localizeNumber } from '../../shared/utils'
+import { isValidJsonString, localizeNumber } from '../../shared/utils'
 import { MPTIssuanceFormattedInfo } from '../../shared/Interfaces'
 import { TokenTableRow } from '../../shared/components/TokenTableRow'
+import { JsonView } from '../../shared/components/JsonView'
 
 interface Props {
   data: MPTIssuanceFormattedInfo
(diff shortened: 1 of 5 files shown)


mvadari:
Why not use `SimpleGroup` instead?

image

shawnxie999 (approved):
LGTM

achowdhry-ripple (approved):
LGTM

pdp2121:
> Why not use `SimpleGroup` instead?
> 
> image
> 
> Or how the hook parameters are done here: image

I was thinking of this at first but since we don't have a standardized structure for metadata, the SimpleGroup might get messy (when there's nested json or array for example). Julian and I both think we can just use the Json formatter for now and revisit once the metadata structure has been defined, probably in the next couple of months. Please let me know if you think differently.

mvadari:
> I was thinking of this at first but since we don't have a standardized structure for metadata, the SimpleGroup might get messy (when there's nested json or array for example). Julian and I both think we can just use the Json formatter for now and revisit once the metadata structure has been defined, probably in the next couple of months. Please let me know if you think differently.

That's fair - if you don't know how much it'll be nested that can get complex. Makes sense for now.

PR #1059, opened 2024-10-18
fix count on validator page, esp important around activation time.

shortthefomo:

Fix bug on count of validators, the current implementation uses length and not length -1
### Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### TypeScript/Hooks Update


mvadari:
Why should it use `length - 1`?

shortthefomo:
> Why should it use `length - 1`?

length of 
`const array = ['0'] ` 
`array.length = 1 not 0 `

shortthefomo:
wait your right this is wrong.

mvadari:
In case additional proof is needed - 29 valid validators, with length 29 (sorted so you can check for duplicates easily):
image

Issue #1058, opened 2024-10-18
Feature request: Page for fee voting


mvadari:
There is a page showing the current voting state for amendments: https://livenet.xrpl.org/amendments

I would like to see a page showing the current voting state for fees and reserves.

ckeshava:
It's also difficult for me to understand projections for the fees and reserves. If we could explain the process of consensus in a more easier/visual manner, that would be helpful.

I don't have good ideas on how to visualize that process

pdp2121:
> It's also difficult for me to understand projections for the fees and reserves. If we could explain the process of consensus in a more easier/visual manner, that would be helpful.
> 
> I don't have good ideas on how to visualize that process

@ckeshava probably a graph similar to rippled version upgrade status page would work: https://livenet.xrpl.org/network/upgrade-status

We will be working on a design

mrbeastmcnasty:
I think your upgrade is a huge help. 

feat: Add validators fee voting tab & reorganize Network tab #1098

PR #1098, opened 2024-12-11
feat: Add validators fee voting tab & reorganize Network tab

pdp2121:

This PR is trying to achieve the following:
- Add voting (for fee voting) and uptime (similar to the existing page) for validators page.
- Remove "validators", "nodes", and "upgrade status" tabs for Network. Each now would be a separate page under Network in top nav bar.
- Refactor ValidatorsTable.jsx into TypeScript hooks.

Resolve: https://github.com/ripple/explorer/issues/1058

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### TypeScript/Hooks Update

## Before / After

Validators page (Uptime):
Screenshot 2024-12-11 at 11 54 28 AM

Validators page (Voting):
Screenshot 2024-12-11 at 1 39 56 PM

Voting Mobile View:

Screenshot 2024-12-11 at 1 40 39 PM

Upgrade Status:
Screenshot 2024-12-11 at 1 41 16 PM


src/rippled/lib/rippled.js
```diff
@@ -509,6 +509,17 @@ const getServerInfo = (rippledSocket) =>
     return resp
   })
 
+const getServerState = (rippledSocket) =>
```
achowdhry-ripple, line 512: nit - might be worth it to add comments between this and getServerInfo, at a glance it is hard to tell what each is for. I don't feel strongly about this though so either way

src/containers/shared/vhsTypes.ts
```diff
@@ -101,6 +101,15 @@ export interface StreamValidator extends ValidatorResponse {
   ledger_hash?: string
   pubkey?: string
   time?: string
+  base_fee?: number
+  reserve_base?: number
+  reserve_inc?: number
```
achowdhry-ripple, line 106: could you make this a union of objects that includes feesettings? seems like all 3 would always be included together (you wouldn't just have base fee or reserve fee for example) since its all part of the same object below

src/containers/shared/utils.js
```diff
@@ -22,6 +22,7 @@ export const FETCH_INTERVAL_VHS_MILLIS = 60 * 1000 // 1 minute
 export const FETCH_INTERVAL_NODES_MILLIS = 60000
 export const FETCH_INTERVAL_ERROR_MILLIS = 300
 export const FETCH_INTERVAL_XRP_USD_ORACLE_MILLIS = 60 * 1000
```
achowdhry-ripple, line 24:
```suggestion
export const FETCH_INTERVAL_XRP_USD_ORACLE_MILLIS = 60 * 1000 // 1 minute
```

src/containers/Network/ValidatorsTable.tsx
```diff
@@ -0,0 +1,196 @@
...
+import UpIcon from '../shared/images/ic_up.svg'
+import DownIcon from '../shared/images/ic_down.svg'
+import DomainLink from '../shared/components/DomainLink'
+import InfoIcon from '../shared/images/info.svg'
+import { Loader } from '../shared/components/Loader'
+import './css/validatorsTable.scss'
+import { useLanguage } from '../shared/hooks'
+import { renderXRP } from '../shared/utils'
+
+const DROPS_TO_XRP_FACTOR = 1000000
```
achowdhry-ripple, line 15: nit: is this not a const somewhere global for the repo? seems like it would be used elsewhere

src/containers/Network/ValidatorsTable.tsx
```diff
@@ -0,0 +1,196 @@
...
+export const ValidatorsTable = (props: ValidatorsTableProps) => {
+  const { validators: rawValidators, metrics, tab, feeSettings } = props
+  const validators = rawValidators ? sortValidators(rawValidators) : undefined
+  const { t } = useTranslation()
+  const language = useLanguage()
+
+  const renderDomain = (domain) => domain && <DomainLink domain={domain} />
+
+  const renderAgreement = (className, d) =>
+    d ? (
```
achowdhry-ripple, line 63: nit: can we rename variable d to a more interpretive param name

achowdhry-ripple (approved):
Only nitpicks on my end, looks good overall though

src/containers/Network/Validators.tsx
```diff
@@ -55,6 +80,20 @@ export const Validators = () => {
     return updated
   }
 
+  function fetchFeeSettingsData() {
+    if (tab === 'voting') {
+      getServerState(rippledSocket)
+        .then((res) => res.state)
```
achowdhry-ripple, line 86: tiny nit: since line 86 isnt an async operation, can we just extract in the following then clause instead of separately here

achowdhry-ripple (approved):
lgtm, thanks for this

src/containers/shared/vhsTypes.ts
pdp2121: FeeSettings refer to network fee, which should always be available, while not all Validators has their votes for fee

src/containers/Network/css/style.scss
```diff
@@ -39,4 +39,20 @@
...
     margin: auto;
   }
+
+  .type {
+    display: inline-block;
+    margin-top: 80px;
+    margin-bottom: 32px;
+    margin-left: 16px;
+    margin-left: calc((100vw - 1500px) / 2);
+    margin-left: clamp(16px, calc((100vw - 1500px) / 2), calc((100vw - 1500px) / 2)); // Adjust based on wrap margin with min 16px
```
Patel-Raj11, line 49: nit: we can define these magic numbers as SCSS variables. I see them in other .scss files as well, so can be cleaned up in a separate effort.

Patel-Raj11 (approved):

src/containers/Network/css/style.scss
pdp2121: Great call! I will add a clean up tix for it

PR #1153, opened 2025-04-08
fix: remove BigQuery from Token page

pdp2121:

We are moving away from using Google Cloud for the Explorer.
The Explorer is using GCP credential to query top trading pairs data from this public dataset:
[GitHub - WietseWind/fetch-xrpl-transactions: Fetch (and store) all XRPL transactions (to a custom datastore)](https://github.com/WietseWind/fetch-xrpl-transactions) 
which has not been maintained recently. It would make sense to remove this part completely from the token page.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

## Before / After

Token page will no longer contain `Top Trading Pairs` section.


ckniffen (approved):
Great change.  Love how many dependencies it eliminates which will make staying up to date much easier.

achowdhry-ripple (approved):
lgtm

Issue #1145, opened 2025-03-31
Add custom network field uses incorrect URL


legleux:
Inputting the URL for a custom network as `localhost:6005`, when the resulting page is opened the URL gets translated to `https://custom.xrpl.org//localhost:6005` (with an extra `/`) resulting in the "UH-OH Page Not Found Please double check your URL" message.

Removing the extra `/` in the URL allows the page to load.

PR #1168, opened 2025-05-06
Fix: Custom network invalid URL generation

achowdhry-ripple:

Related Issue: https://github.com/ripple/explorer/issues/1145

URL was not accounting for how certain environments have the href base with a trailing slash by default, causing a double slash and invalid URL construction. This was causing the URL to generate successfully in our local environment, and to fail in our hosted ones. This small change checks for a pre-existing slash before appending an additional.

Also upgraded 2 dependencies for security vulnerabilities
https://github.com/advisories/GHSA-67mh-4wv8-2f99
https://github.com/advisories/GHSA-vg6x-rcgg-rjx6

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### TypeScript/Hooks Update


package.json
```diff
@@ -42,7 +42,7 @@
     "ripple-address-codec": "^5.0.0",
     "topojson-client": "^3.0.0",
     "usehooks-ts": "^3.1.0",
-    "vite": "^5.4.8",
+    "vite": "^5.4.12",
```
mvadari, line 45: Are these upgrades necessary as a part of this PR?

mvadari (commented):
Please add a testcase.

package.json
achowdhry-ripple: not necessary, just were small improvements I tucked into this PR since it was also small. Can separate it into a different PR if you would prefer that instead

mvadari: I'd prefer it in a separate PR, but I don't care that much if you want to leave it in

achowdhry-ripple: Will leave it in for now since its already there then, and keep it in mind for future PRs

Patel-Raj11 (approved):

pdp2121 (approved):

PR #1169, opened 2025-05-07
feat: Add new design & revision for ledgers metrics

pdp2121:

Add new design & descriptions for ledgers metrics on index page.
Also, the old code trying to calculate metrics for previous 100 ledgers. However, it stores history of 2 minutes only (~33 ledgers). This PR will extend the cache time to 5 minutes to try to capture 50 previous ledgers and reflect this change in the description for clarification.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Large Screen

Screenshot 2025-05-07 at 5 19 38 PM

### Tooltip on hover
Screenshot 2025-05-07 at 5 20 10 PM

### Midsize Screen (flex wrap)

Screenshot 2025-05-07 at 5 20 44 PM

### Mobile Screen
Screenshot 2025-05-07 at 5 21 33 PM


server/lib/streams.js
```diff
@@ -3,7 +3,7 @@ const utils = require('./utils')
 const log = require('./logger')({ name: 'streams' })
 
 const PURGE_INTERVAL = 10 * 1000
-const MAX_AGE = 120 * 1000
+const MAX_AGE = 5 * 60 * 1000
```
achowdhry-ripple, line 6: just curious why the increase?

src/containers/Ledgers/css/ledgerMetrics.scss
```diff
@@ -2,59 +2,146 @@
...
+        width: 1px;
+        height: 80%;
+        background-color: $black-60;
+        content: '';
+        transform: translateY(-50%);
+      }
+
+      // Hacky way to get rid of the double border when breaking to second row. TODO: cleanup
+      // 3 columns per row
+      @media (width >= 544px) and (width <= 703px) {
```
achowdhry-ripple, line 55: how were these width breakpoints chosen? can we make them variables if they are standard screen sm med lg values

achowdhry-ripple (approved):
love the new design!

server/lib/streams.js
pdp2121: 120 seconds would only suffice to get around 35 ledgers, but we would want to include 50 ledgers for our metric calculation

src/containers/Ledgers/css/ledgerMetrics.scss
pdp2121: It's 2*margin + n*min-width. I'm not sure if there's a clean way to handle this. I will create an issue for cleaning up these. We need to add more variables

Patel-Raj11 (approved):
LGTM

PR #1215, opened 2025-07-30
remove cheerio dev-dependency

ckeshava:

This is an attempt to supercede this PR: https://github.com/ripple/explorer/pull/1202. `cheerio` is a dependency of `enzyme` library. The explorer code base does not have usages of the `cheerio` library.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "@typescript-eslint/eslint-plugin": "^8.32.1",
     "@typescript-eslint/parser": "^8.32.1",
     "babel-jest": "^29.7.0",
-    "cheerio": "1.0.0-rc.3",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-17-updated": "^1.0.2",
     "eslint": "^8.57.0",


mvadari:
I disagree with this PR. The reason it's in the package.json is because the higher versions break tests. And the only way to pin a version is to specify it like that - otherwise npm will automatically try to update it. Tests only pass because npm uses `npm ci` to install packages, and you haven't changed the version in `package-lock.json`.

ckeshava:
> Tests only pass because npm uses npm ci to install packages, and you haven't changed the version in package-lock.json

I used `npm i --legacy-peer-deps` on my local machine. I don't see any errors. The `legacy-peer-deps` switch is required to resolve other dependencies (`react@16.14.0`, `airbnb-prop-types@2.16.0`).

> And the only way to pin a version is to specify it like that - otherwise npm will automatically try to update it.

If enzyme wants to use versions of cheerio only up until `1.0.0-rc.3`, it is enzyme's responsibility to specify the correct semantic-version in its `package.json` file. Why should we bear the responsibility of pinning the correct version of a secondary dependency?

ckeshava:
alright, I didn't realize that we were using an old version of React. I believe the latest version is at `react@19.1.1`. 

I do not have time to pick up on your PR, I'll keep that in mind 👍

PR #1216, opened 2025-07-31
Use drops to represent txn-fees and reserves

### LINKED ISSUE #1206: Use drops for `BASE FEE` in Fee Voting table
The `BASE FEE` (reference fee) in our Validator Voting table is often small and should be converted to drops for readability.

ckeshava:
This PR changes the unit-of-representation to drops, instead of 1 XRP. Certain quantities (like fees, account-reserve and object reserve) are very minute. For the purposes of readability, it is better represented as drops. 
(For reference: 1 XRP = 10 ^ 6 drops)

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

## Before / After

At the present moment, the validator voting tab looks like this:

image

This PR enables the voting tab to look like this:
image

### DIFF
--- a/src/containers/Validators/VotingTab.tsx
+++ b/src/containers/Validators/VotingTab.tsx
@@ -12,7 +12,6 @@ import {
   renderXRP,
 } from '../shared/utils'
 import { useAnalytics } from '../shared/analytics'
-import { XRP_BASE } from '../shared/transactionUtils'
 
 import './votingTab.scss'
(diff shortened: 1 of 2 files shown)


ckeshava:
There is a lack of consensus on the need for this change. I'll close this PR at the present time. If there is any interest in the future, we can revisit this PR.

PR #1226, opened 2025-08-12
fix: use VHS for `EnableAmendment` rippled version

pdp2121:

`EnableAmendment` transaction used to fetch rippled version from xrpl.org but the page no longer support this. Use VHS as source of truth (fetching from XRPScan)

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -74,6 +74,7 @@ jobs:
           VITE_MAINNET_LINK: mainnet.example.com
           VITE_TESTNET_LINK: testnet.example.com
           VITE_CUSTOMNETWORK_LINK: custom.example.com
+          VITE_DATA_URL: https://data.xrpl.org/v1/network
 
   build:
     runs-on: ubuntu-latest


.github/workflows/nodejs.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           VITE_MAINNET_LINK: mainnet.example.com
           VITE_TESTNET_LINK: testnet.example.com
           VITE_CUSTOMNETWORK_LINK: custom.example.com
+          VITE_DATA_URL: https://data.xrpl.org/v1/network
```
Patel-Raj11, line 77: Looks like this is the only place where an actual integration test is being done in the Explorer codebase. Otherwise other env variables are dummy.

Patel-Raj11 (approved):

.github/workflows/nodejs.yml
pdp2121: Yes correct

Issue #1210, opened 2025-07-23
tfWithdrawAll flag on AMMWithdraw displays negative balance on explorer


hazardcookie:
Transactions using this flag to withdraw all and delete the amm cause a negative balance to appear on the explorer. This value should be positive, so this just seems like a front-end parsing error:

https://livenet.xrpl.org/transactions/68AD5C03CCF55D84EDB68E92E014716E6D059A4D6779E1E1595EBCC2E14E8B39/simple

Issue #1207, opened 2025-07-16
Negative balance on AMMDeposit minted LPtokens


hazardcookie:
While looking into some data i noticed a AMMDeposit transaction on livenet explorer that was displaying a negative amount for the LP Tokens minted. This number should be positive afaik (and in most cases the explorer reflects that correctly).

https://livenet.xrpl.org/transactions/6D85FF2BEE30E53F690A0DBEBCAAB83AD207E264A49EA83DE29A30A1BD36D06D/simple

Seems like a edge case.

PR #1229, opened 2025-08-15
Fix negative LPTokens and XRP assets for AMMDeposit and AMMWithdraw transactions

Patel-Raj11:
Fixes https://github.com/ripple/explorer/issues/1207 and fixes https://github.com/ripple/explorer/issues/1210
1. Refactored, the code to compute LPTokens balance directly from AMM ledger LedgerEntryType from Meta fields. AMMDeposit and AMMWithdraw will have exactly one AMM ledger LedgerEntryType in meta fields and will show the LPTokens moved from that AMM.

2. Fixed an edge case when transaction fees are higher than the XRP asset being withdraw/deposited to AMM. Previously, it was showing -ve  XRP amount if transaction fee was more than the absolute value of XRP asset moved.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

## Before / After

Before
image
image

After
image
image

## Test Plan

Added test cases


src/containers/shared/metaParser.tsx
```diff
@@ -103,7 +104,7 @@ function findXRPAmount(
       ? Number(xrp.FinalFields.Balance) - Number(xrp.PreviousFields.Balance)
       : Number(xrp?.NewFields?.Balance),
   )
-  balance -= Number(tx.Fee)
+  balance = Math.abs(balance - Number(tx.Fee))
```
ckeshava, line 107: Why are you applying `Math.abs(...)` here? 

Suppose the fees of a transaction is higher than its balance, this code change will display a positive balance, however this is incorrect behavior.

Note: It is possible for a transaction to fail due to insufficient-reserve error.

src/containers/shared/components/Transaction/AMMWithdraw/test/AMMWithdraw.test.tsx
```diff
@@ -70,4 +71,17 @@ describe('AMM Withdraw Tests', () => {
...
+  it('renders LP Tokens properly', () => {
+    const wrapper = createWrapper(withdrawAll)
+    expectSimpleRowText(wrapper, 'lp_tokens', '4.77')
+    wrapper.unmount()
+  })
+
+  it('renders positive XRP amount even if transaction fee is greater than XRP taken out of AMM', () => {
+    const wrapper = createWrapper(withdrawAll)
+    expectSimpleRowNotToExist(wrapper, 'asset2')
+    expectSimpleRowText(wrapper, 'asset1', '\uE9000.000005 XRP')
```
ckeshava, line 84: Can you explain how you obtained this value? I would have expected this to equal to
```
const node = meta['AffectedNodes'][1].DeletedNode
console.log(node.FinalFields.Balance - node.PreviousFields.Balance) // 4.99 XRP (or) 4999900 drops
```

Here is the relevant snippet from the `withdraw_all.json` file:

Image

src/containers/shared/metaParser.tsx
Patel-Raj11: @ckeshava I see, missing Before screenshot made the change less clear. I have added those, and the `after` ones are for the exact same transactions.

Patel-Raj11: > Why are you applying Math.abs(...) here?

`balance` at line 102 will give the total XRP balance that changed for this account. Now, we want to show the XRP asset that moved in/out of AMM excluding the fees, thus we are subtracting the fees. -ve XRP balances creates confusions here and it's clear from `Asset 1 Out` text on the landing page that X amount of XRP moved out of the AMM.

> Note: It is possible for a transaction to fail due to insufficient-reserve error.

Yes, if it had failed in the validated ledger, the balance would be 0 and we won't show the `Asset 1 Out` at all. In this case, it succeeded in the validated ledger.

src/containers/shared/components/Transaction/AMMWithdraw/test/AMMWithdraw.test.tsx
Patel-Raj11: `rGjXpVMZtX5KFPNxt7srucLuLsPXxHLCPe` is not the transaction executor, its the account that owns the AMM. We should check the AccountRoot of `rhaAaAaaSDNiZpvUPQN5H5aP6kpZ2brare`, since we want to see the asset it gained from the AMM once the LPTokens were deposited.

Patel-Raj11: > Can you explain how you obtained this value? I would have expected this to equal to

It's computed here `findXRPAmount`. I have explained it in my other comment. It just computes the XRP assets moved in/out of AMM.

Patel-Raj11: After scratching my head for a few minutes, I just realized that we both are talking the same thing. Your calculation is from AMM owner perspective and the current code computes from transaction executor's perspective. i.e. movement of 5 drops (0.000005 XRP).

ckeshava: Hello Raj, I understand your points. However, I don't understand the test (with the mocked data) which prints the text `\uE9000.000005 XRP` -- How can I compute this value from the `withdraw_all.json` file? 

Also, where does this text appear on the `AMMWithdraw` transaction's `Simple` tab screenshot?

Patel-Raj11: 1. To get the value, use the AccountRoot node of the transaction sender (`rhaAaAaaSDNiZpvUPQN5H5aP6kpZ2brare`).

```
{
        "ModifiedNode": {
          "FinalFields": {
            "Account": "rhaAaAaaSDNiZpvUPQN5H5aP6kpZ2brare",
            "Balance": "94165821",
            "Flags": 0,
            "OwnerCount": 2,
            "Sequence": 94287973
          },
          "LedgerEntryType": "AccountRoot",
          "LedgerIndex": "A94F7C14018E704F0F99B5731ABD9DF179CFD17DEEB09B64DC58695ADA454949",
          "PreviousFields": {
            "Balance": "94165916",
            "Sequence": 94287972
          },
          "PreviousTxnID": "F432C4C2BAFE4CF8515E60925D4374D7CB067B68362E55920055B6FCA589DA59",
          "PreviousTxnLgrSeq": 97666613
        }
      },
```
Math.abs(Math.abs((FinalFields.Balance - PreviousFields.Balance)) - Fee) = 5 drops
|(|94165821 - 94165916|) - 100| = 5

2. The text appears here [`Asset 1 Out`](https://livenet.xrpl.org/transactions/68AD5C03CCF55D84EDB68E92E014716E6D059A4D6779E1E1595EBCC2E14E8B39/simple).

ckeshava: Yes, I understand. The source of my confusion is the strange character that appears in the output html. I had expected the output to be:
```
    expectSimpleRowText(wrapper, 'asset1', '0.000005 XRP')
```
However, there is this ascii character before the XRP value: `\ue900`. I mis-interpreted this as `9000` XRP in the balance field.

It turns out that this is the ascii character for the XRP display symbol which is [introduced in the currency-code here](https://github.com/ripple/explorer/blob/69256af9a5f97e0ea5a61eab82593b7b7433bdc4/src/containers/shared/components/Currency.tsx#L64).

ckeshava (approved):

src/containers/shared/metaParser.tsx
```diff
@@ -25,29 +25,30 @@ export function getAMMAccountID(meta: any) {
 }
 
 export function getLPTokenAmount(meta: any) {
-  // TODO: possibility of bug if currency that isnt LP token has 03 at the start
```
achowdhry-ripple, line 28: is this potential bug still a risk here?

achowdhry-ripple (approved):
lgtm overall

src/containers/shared/metaParser.tsx
ckeshava: Since the objects are filtered based on the AMM Ledger object type, this bug should not be possible. We are not checking the `03` initial substring in this piece of code.

PR #1245, opened 2025-10-01
Explorer account page refactoring

kuan121:
### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization


public/locales/en-US/translations.json
```diff
@@ -617,5 +613,90 @@
...
+  "account_flag_title_asf_authorized_nft_token_minter": "NFT Minter",
+  "account_flag_title_lsf_disallow_incoming_check": "No Checks Allowed",
+  "account_flag_title_lsf_deposit_auth": "Deposit Authorization",
+  "account_flag_title_asf_account_txn_id": "Track Account Latest Transaction",
+  "account_flag_description_lsf_global_freeze": "Freeze all assets issued by this account.",
+  "account_flag_description_lsf_disable_master": "Disallow use of the master key pair.",
+  "account_flag_description_lsf_default_ripple": "Enable rippling on this account's trust lines by default.",
+  "account_flag_description_lsf_allow_trustline_clawback": "Allow account to claw back tokens it has issued.",
+  "account_flag_description_lsf_require_destination_tag": "Require a destination tag to send transactions to this account.",
+  "account_flag_description_lsf_no_freeze": "Permanently give up the ability to freeze individual trust lines or disable Global Freeze. This flag can never be disabled after being enabled.",
```
Patel-Raj11, line 637: For my understanding: Where did we get these descriptions from?

src/containers/Accounts/AccountAsset/FutureDataIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const FutureDataIcon = ({
+  message = 'This data will be provided in a future release.',
+}: FutureDataIconProps) => {
+  const { showTooltip, hideTooltip } = useTooltip()
+
+  const handleShow = (e: React.MouseEvent | React.FocusEvent) => {
+    const rect = e.currentTarget.getBoundingClientRect()
+    showTooltip('text', e as any, message, {
+      // The tooltip position is adjusted to appear below and slightly to the right of the icon
+      x: rect.left + window.scrollX + 10,
```
Patel-Raj11: nit: Replace these magic numbers with named constants outside the component.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,331 @@
...
+      marker = accountLinesResponse?.marker || ''
+    } catch (error) {
+      log.error(
+        `Error fetching account lines for account ${accountId}: ${JSON.stringify(error)}`,
+      )
+      // Break the loop on error to avoid infinite retry
+      break
+    }
+  } while (marker)
+  log.info(`Found ${allTrustLines.length} trust lines`)
```
Patel-Raj11: Do we need these console messages? If not then we should remove those from all components.

public/locales/en-US/translations.json
kuan121: They're the first sentence of the description of the `asf` flags mentioned on this [page](https://xrpl.org/docs/references/protocol/transactions/types/accountset#accountset-flags)

src/containers/Accounts/AccountAsset/FutureDataIcon.tsx
kuan121: Sounds good. Will do.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,331 @@
...
+  transferFee?: string
+  frozen?: string
+}
+
+const fetchAccountHeldIOUs = async (
+  rippledSocket: any,
+  accountId: string,
+): Promise<IOU[]> => {
+  log.info(`Finding held IOUs via 'gatewayBalance' for account ${accountId}`)
+  const balancesResponse = await getBalances(rippledSocket, accountId)
```
Patel-Raj11: `getBalances` could throw an exception. We should try..catch it.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: I’ve noticed that we don’t usually write logs in the Explorer frontend. Because of this, when a page loads incorrectly or a section shows the wrong data, it’s hard to know what happened. If you open the Developer Tools console on [this page](https://livenet.dev.ripplex.io/accounts/r3hnq9k9NcvAMPfD2DHjFpzDQLxS2xAaki) and reload it, you can immediately see what actions are being attempted and whether something fails.

I also saw a [TODO](https://github.com/ripple/explorer/blob/explorer-account-page-refactoring/src/rippled/lib/logger.js#L1-L3) to actually ship those logs to backend.

Let me know if you have some concerns about writing those console logs.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: Good catch, I will wrap it in a try/catch then.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+  const iouTokens: any[] = []
+  for (const assets of Object.values(balancesResponse?.assets || {})) {
```
Patel-Raj11: nit: Nullish Coalescing Operator (??) should be preferred.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,331 @@
...
+    for (const asset of assets as any[]) {
+      if (asset.currency && !asset.currency.startsWith(LP_TOKEN_IDENTIFIER)) {
```
Patel-Raj11: Non LPTokens can have currency that starts with 03. Not thought much about it, but might need another API [call](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_info) to figure out if the issuer is an AMM account.

Example: https://testnet.xrpl.org/transactions/054C2B1D216893010D084C9A9D50BFB18E0D0C5B7EF09650C4188104A57D183B/simple

Patel-Raj11: Same suggestions for all pages that filters out LPToken as per this logic.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: OK. Let me give it a try to use ?? instead.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: Interesting. How did you find the token in testnet? Are you able to find some in mainnet? I wonder if it's legit to have IOU that starts with `03`. Let me check with the core ledger team as well.

References:
1. https://xrpl.org/docs/concepts/tokens/decentralized-exchange/automated-market-makers#lp-token-currency-codes
2. https://xrpl.org/docs/references/protocol/data-types/currency-formats#nonstandard-currency-codes

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
Patel-Raj11: They will end-up in the users Developer Tools console. So I think if they are only for debugging purpose, we can add them while debugging any issue and use developer tools to figure out the cause. No strong opinions but I would only ship error log statements to production.

If keeping these helps for analytics (once they are pushed to some database using API call as mentioned in TODO) then we should keep those as it is.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
Patel-Raj11: I created a transaction on Testnet using 03 currency code prefix. Occurred to me while working on https://github.com/ripple/explorer/pull/1229/files#r2291407675

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: >If keeping these helps for analytics (once they are pushed to some database using API call as mentioned in TODO) then we should keep those as it is.

The logs are mostly for debugging purpose. I will remove the `log.info(~)` statements for now and keep the `log.warn(~)` and `log.error(~)` logs.

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: Although it’s uncommon for IOU tokens to start with 03, we have to to handle that case to ensure full reliability.
1. For HeldIOUs.tsx, I dded an `account_info` call for tokens starting with `03` to check whether the corresponding account has `AMMID`.
2. For IssuedIOUs.tsx, I removed the `03` prefix check since it’s redundant. If an account issues an LP token, it will be displayed on the AMMAccount page instead of the regular account page.
3. For HeldLPTokens.tsx — No changes needed, as it already ignores IOUs that start with `03`.

.env.example
```diff
@@ -28,3 +28,6 @@ VITE_DATA_URL=https://data.xrpl.org/v1/network
 # Whether to use ws instead of wss (boolean)
 # Only used locally (the deployed Explorer requires wss)
 VITE_INSECURE_WS=0
+
+# LOS Endpoint URL
+VITE_LOS_URL=https://los.dev.ripplex.io
```
achowdhry-ripple, line 33: nit: could we update the example to the prod url just in the rare case that someone tries to use it?

src/containers/Accounts/AccountAsset/FutureDataIcon.scss
```diff
@@ -0,0 +1,14 @@
+@use '../../shared/css/variables' as *;
+
+.future-data {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  cursor: default;
+
+  .clock-icon {
+    width: px;
```
achowdhry-ripple: nit: looks like there is a missing px value

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,357 @@
...
+): Promise<IOU[]> => {
+  let balancesResponse
+  try {
+    balancesResponse = await getBalances(rippledSocket, accountId)
+  } catch (error) {
+    log.error(
+      `Error calling 'gatewayBalances' for account ${accountId}: ${JSON.stringify(error)}`,
+    )
+    return []
+  }
```
achowdhry-ripple, line 64: I've seen these types of functions saved in the rippled folder in this repo. I don't feel strongly here but might be worth adding this there and importing for consistency

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    const uniqueIssuers = [...new Set(sortedIOUs.map((line) => line.issuer))]
+    for (const issuer of uniqueIssuers) {
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        const accountInfo = await getAccountInfo(rippledSocket, issuer, false)
+
+        const transferFee = formatTransferFee(accountInfo?.TransferRate, 'IOU')
+
```
achowdhry-ripple, line 237: could some sort of parallelism here save time? not sure how long this process takes currently from a UI experience

src/containers/Accounts/AccountAsset/assetTables/HeldMPTs.tsx
```diff
@@ -0,0 +1,219 @@
...
+                <td>
+                  <RouteLink to={MPT_ROUTE} params={{ id: token.tokenId }}>
+                    {shortenMPTID(token.tokenId)}
+                  </RouteLink>
+                </td>
+                <td>{token.ticker ? token.ticker : <FutureDataIcon />}</td>
+                <td>
+                  <Account
+                    account={token.issuer}
+                    shortAccount={shortenAccount(token.issuer)}
```
achowdhry-ripple: nit: passing both of these in seems redundant?

src/containers/Accounts/AccountAsset/assetTables/HeldMPTs.tsx
```diff
@@ -0,0 +1,219 @@
...
+                  />
+                  {token.issuerName && (
+                    <div className="issuer-name">{token.issuerName}</div>
+                  )}
+                </td>
+                <td>
+                  <FutureDataIcon />
+                </td>
```
achowdhry-ripple, line 183: how come we are unable to provide the usd prices right now?

src/containers/Accounts/AccountAsset/assetTables/HeldNFTs.tsx
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    const allNFTs: any[] = []
+    let marker = ''
+    do {
+      // eslint-disable-next-line no-await-in-loop
+      const response = await getAccountNFTs(
+        rippledSocket,
+        accountId,
+        marker,
+        10, // Not 10 NFTs, but 10 pages of NFTs
```
achowdhry-ripple, line 26: why specifically 10?

achowdhry-ripple: nvm, I see you are calling it several times in batch in the table

src/containers/Accounts/AccountAsset/assetTables/NFTTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+            ),
+          )
+
+          // Mark as fetched
+          setFetchedNFTs((prev) => new Set([...prev, nft.nftId]))
+        } catch (error) {
+          // Handle individual NFT errors, continue with others
+          log.error(
+            `Error fetching offers for NFT ${nft.nftId}: ${JSON.stringify(error)}`,
+          )
```
achowdhry-ripple, line 151: should we add retry mechanisms here? or would a page reload be sufficient

src/containers/shared/NumberFormattingUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  let options
+  if (balance >= 1) {
+    options = USD_CURRENCY_OPTIONS
+  } else if (balance >= 0.0001) {
+    options = USD_SMALL_BALANCE_CURRENCY_OPTIONS
+  } else {
+    options = USD_EXTRA_SMALL_BALANCE_CURRENCY_OPTIONS
+  }
```
achowdhry-ripple: I do believe there are exisitng utils in this repo to display amounts, were you able to try those / is this logic significantly different?

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: `getBalances` is defined in `rippled/`. Here, we have a special use case of finding held IOUs from the response of `getBalances`. I don't think it's generic enough, unlike getBalances`, to be in `rippled/`

src/containers/Accounts/AccountAsset/assetTables/HeldIOUs.tsx
kuan121: Based on my tests, parallel requests quickly get throttled by Clio. It’s fine to make these calls sequentially here: we display the Held IOU table with other fields almost immediately, then fetch the transfer fee and frozen status one by one, updating the columns as the data arrives. From the user’s perspective, the table shows up immediately, so the experience isn’t blocked.

src/containers/Accounts/AccountAsset/assetTables/HeldMPTs.tsx
kuan121: It's not redundant since the caller of `<Account />` can pass in a short version of account, e.g., `rK2Y8ng...ZkyFm` or an issuer name like `Ripple` as `shortAccount`. I will rename `shortAccount` to `displayText` since issuer name like `Ripple` is definitely not a short account.

src/containers/Accounts/AccountAsset/assetTables/HeldMPTs.tsx
kuan121: Unlike IOU, MPT isn't being traded on the ledger at the moment. We need to have MPT DEX in mainnet so that MPT can be traded. Only after that will we have an MPT price in XRP and USD.

src/containers/Accounts/AccountAsset/FutureDataIcon.scss
kuan121: Good catch!

.env.example
kuan121: For testing purpose, I think dev endpoint should be sufficient. In general, it's not a good idea to expose prod endpoint even though we have DDoS protection in place.

src/containers/Accounts/AccountAsset/assetTables/NFTTable.tsx
kuan121: Based on my tests, we don't need a retry for now since I have tested some accounts with large number of tokens and I don't see any throttle. I plan to add retry as a follow-up.

src/containers/shared/NumberFormattingUtils.ts
kuan121: You’re right — we already have [localizeNumber](https://github.com/ripple/explorer/blob/explorer-account-page-refactoring/src/containers/shared/utils.js#L146), and the functions in this file use it under the hood. However, I’ve added some custom logic, after discussed with Julian, to control how many decimals we display for different prices and balances. We can also update the Token and Token Ranking pages to use these functions for consistency.

Here’s the custom logic:
Token balance:
1. If the balance is greater than `999`, keep 2 decimals (e.g., `100,001.94`)
2. If the balance is less than or equal to `999`, keep 2 to 4 decimals, (e.g., `151.194`)

USD price or balance:
1. If the value is greater than `1`, keep at most 2 decimals (e.g., `$1.02`).
2. If the value is between `0.0001` and `1`, keep 2–4 decimals (e.g., `$0.004`, `$0.02`).
3. If the value is less than `0.0001`, keep up to 10 decimals (e.g., `$0.00004`).

src/containers/Accounts/AccountAsset/assetTables/HeldNFTs.tsx
kuan121: It's just a number I pick to fetch data quick enough without being throttled by Clio if request are made sequentially.

package.json
```diff
@@ -4,8 +4,6 @@
   "version": "1.5.0",
   "dependencies": {
     "@rollup/plugin-inject": "^5.0.5",
-    "@testing-library/jest-dom": "^5.17.0",
-    "@testing-library/react": "^12.1.5",
     "@vitejs/plugin-react": "^4.2.1",
     "@xrplf/isomorphic": "^1.0.0-beta.1",
     "@xrplf/prettier-config": "^1.9.1",
```
Patel-Raj11: @kuan121 While you are at it, can you move `@xrplf/prettier-config` to devDependencies?

kuan121: Will do

Patel-Raj11 (approved):
LGTM!

PR #1248, opened 2025-10-08
feat: Tokens ranking page

pdp2121:

Create a tokens ranking page to show ranking of tokens based on multiple metrics (only showing tokens with trust scores 1-3 using XRPL Meta's methodology).

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

## Before/After

Page:

livenet dev ripplex io_tokens

Filter:

livenet dev ripplex io_tokens (1)

Pagination:

Screenshot 2025-10-08 at 5 25 19 PM

Tooltip:

Screenshot 2025-10-08 at 5 28 40 PM (3)

### DIFF
--- a/src/containers/shared/test/utils.test.ts
+++ b/src/containers/shared/test/utils.test.ts
@@ -39,12 +39,12 @@ describe('utils', () => {
   it('formatLargeNumber format numbers correctly', () => {
     expect(formatLargeNumber()).toEqual({ num: '0.0', unit: '' })
     expect(formatLargeNumber(2000000000000)).toEqual({
-      num: '2.0000',
+      num: '2.0',
       unit: 'T',
     })
-    expect(formatLargeNumber(3300000000)).toEqual({ num: '3.3000', unit: 'B' })
-    expect(formatLargeNumber(44400000)).toEqual({ num: '44.4000', unit: 'M' })
-    expect(formatLargeNumber(555500)).toEqual({ num: '555.5000', unit: 'K' })
+    expect(formatLargeNumber(3300000000)).toEqual({ num: '3.3', unit: 'B' })
+    expect(formatLargeNumber(44400000)).toEqual({ num: '44.4', unit: 'M' })
+    expect(formatLargeNumber(555500)).toEqual({ num: '555.5', unit: 'K' })
     expect(formatLargeNumber(66.666, 2)).toEqual({ num: '66.67', unit: '' })
   })
 


server/routes/v1/tokens.js
```diff
@@ -27,19 +26,10 @@ const parseCurrency = (currency) => {
     : currency
 }
 
-async function fetchXRPLMetaTokens(offset) {
-  log.info(`caching tokens from ${process.env.XRPL_META_URL}`)
+async function fetchXRPLMetaTokens() {
```
kuan121: nit: Can we rename this method to `fetchTokens` or `fetchTokensFromBackend`?

server/routes/v1/tokens.js
```diff
@@ -48,35 +38,22 @@ async function fetchXRPLMetaTokens(offset) {
...
-  tokensDataBatch = await fetchXRPLMetaTokens(0)
-  const { count } = tokensDataBatch
-  while (offset < count) {
-    allTokensFetched.push(...tokensDataBatch.tokens)
-    offset += XRPLMETA_QUERY_LIMIT
-    // eslint-disable-next-line no-await-in-loop
-    tokensDataBatch = await fetchXRPLMetaTokens(offset)
-  }
+  const losTokens = await fetchXRPLMetaTokens()
+  log.info(`Fetched ${losTokens.tokens.length} tokens from LOS...`)
```
kuan121: If there is an error in fetchXRPLMetaTokens, `tokens` doesn't exist. We need to update the error handling code in fetchXRPLMetaTokens

src/containers/App/App.tsx
```diff
@@ -5,16 +5,19 @@ import './app.scss'
...
 // memoize to prevent react-router from creating a new socket ever single time a rew route is loaded
 export const App: FC<{ rippledUrl: string }> = memo(
   ({ rippledUrl }: { rippledUrl: string }) => (
     <SocketProvider rippledUrl={rippledUrl}>
       <NetworkProvider rippledUrl={rippledUrl}>
-        <Header inNetwork />
-        <div className="content">
-          <Outlet />
-        </div>
+        <TooltipProvider>
```
kuan121, line 15: If we wrap the entire App in `<TooltipProvider>` and there's [an existing component inside `<TooltipProvider>`](https://github.com/ripple/explorer/blob/1d1f100f22ddb5f4c1c941348a43a940021aa55a/src/containers/Ledgers/LedgerListEntry.tsx#L92-L93), does that mean we will have nested `<TooltipProvider>`? Will things break?

src/containers/shared/components/Pagination/index.ts
```diff
@@ -0,0 +1 @@
+export * from './Pagination'
```
kuan121, line 1: I’m not sure if this is common in the JS/TS world, but this seems redundant since we already have `export const Pagination` in `Pagination.tsx`?

src/containers/shared/css/tooltip.scss
```diff
@@ -54,6 +54,8 @@
   }
 
   &.tooltip-text {
+    position: fixed;
+    z-index: 9999;
```
kuan121: Do you check all the places using tooltip that these changes won't break any of the existing tooltips?

src/containers/shared/components/Account.tsx
```diff
@@ -1,14 +1,17 @@
+import { MouseEventHandler, ReactNode } from 'react'
```
kuan121: I made similar changes to Account.tsx recently. Can you rebase on top of `staging` to include my Account Page refactoring changes and resolve conflicts?

server/routes/v1/index.js
```diff
@@ -1,7 +1,7 @@
 const api = require('express').Router()
 const getHealth = require('./health')
 const getCurrentMetrics = require('./currentMetrics')
-const getTokensSearch = require('./tokens')
+const getTokens = require('./tokens')
```
kuan121: nit: Can we change this to `const tokens = require('./tokens')`? It feels a bit strange to see two verbs `getTokens.getAllTokens`.

server/routes/v1/tokens.js
```diff
@@ -96,32 +73,80 @@ function queryTokens(tokenList, query) {
...
+        .toFixed(6),
+      stablecoin: cachedTokenList.tokens
+        .reduce((sum, token) => {
+          const cap = Number(token.market_cap_usd) || 0
+          return token.asset_subclass === 'stablecoin' && cap > 0
+            ? sum + cap
+            : sum
+        }, 0)
+        .toFixed(6),
+    }
```
kuan121: Can `metrics` be cached as well to avoid re-computation for every request?

src/containers/shared/components/TokenSearchResults/TokenSearchResults.tsx
```diff
@@ -9,9 +9,11 @@ import { TokenSearchRow } from './TokenSearchRow'
 import SocketContext from '../../SocketContext'
```
kuan121, line 9: Searching for `RLUSD` on https://livenet.dev.ripplex.io and https://livenet.xrpl.org gives me different results. For example, searching `RLUSD` on https://livenet.dev.ripplex.io gives me only two tokens and Ripple RLUSD is missing:
Image

src/containers/shared/components/middleEllipsis.tsx
```diff
@@ -0,0 +1,133 @@
+import { useLayoutEffect, useRef } from 'react'
```
kuan121: I only see `AddressMiddleEllipsis` used once in `<Account ...>` in this PR:
```
      <Account
        account={token.issuer_account}
        onClick={(e) => e.stopPropagation()}
        displayText={
          <AddressMiddleEllipsis text={token.issuer_account} tailLen={3} />
        }
      />
```
The implementation seems complex and unnecessary for the purpose of generating texts like `GateHub Crypto-rchGBxcD1A1C2tdxF6papQYZ8k…YcL`. Are we just formatting and concatenating two strings?

Is it possible to re-use [shortenAccount](https://github.com/ripple/explorer/blob/1d1f100f22ddb5f4c1c941348a43a940021aa55a/src/containers/shared/utils.js#L405C14-L405C28) to get a short account ID instead? On the Account Page, account IDs are shortened to keep 7 characters at the start and 5 at the end — for example, `rsoLo2S1kiGeCcn6hCUXVrCpGMWLrRrLZz` becomes `rsoLo2S...RrLZz`. Feel free to reuse it after modifying it to make the numbers of characters kept at the start and end two variables, e.g., `shortenAccount(addr = '', startChars = 7, endChars = 5)`.

src/containers/Tokens/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  )
+
+  const getSortValue = (token, field) => {
+    switch (field) {
+      case 'price':
+        return Number(token.price ?? 0)
+      case 'holders':
+        return Number(token.holders) ?? 0
+      case 'market_cap':
+      case 'marketcap':
```
kuan121: market_cap and marketcap? Is it a typo?

public/locales/en-US/translations.json
```diff
@@ -592,6 +592,22 @@
...
+  "market_cap_metric_description": "Circulating supply multiplied by price of all trusted tokens",
+  "market_cap_description": "Circulating supply multiplied by price",
+  "24h_description": "Price change over the past 24h",
+  "volume_description": "Total USD value of all DEX trades in the past 24h",
+  "trades_description": "Number of DEX trades executed in the past 24h",
+  "tvl_description": "USD value of token in AMMs",
+  "stablecoin_description": "Market cap of all trusted tokens classified as stablecoins",
+  "stablecoin": "Stablecoin",
+  "wrapped": "Wrapped",
+  "tokens_footnote": "Only includes trust scores 1-3 using <Link/>'s methodology.",
```
kuan121: Can we improve the footnote? The current version offers very little information. For example,
> This table includes only tokens with XRPL Meta trust levels 1–3 — those verified through KYC or reviewed by trusted publishers. Learn more in [XRPL Meta’s trust guidelines](https://xrplmeta.org/).

src/containers/Tokens/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+            setSortField={setSortField}
+            sortOrder={sortOrder}
+            setSortOrder={setSortOrder}
+            setPage={setPage}
+          />
+          <div className="footnote">
+            <Trans
+              i18nKey="tokens_footnote"
+              components={{
+                Link: <Link to="https://xrplmeta.org">XRPLMeta</Link>,
```
kuan121: I couldn’t find the “Safety Disabled” toggle shown in the Figma design. Without it and the `*`, how will users recognize this as a footnote?

public/locales/en-US/translations.json
```diff
@@ -592,6 +592,22 @@
...
+  "trades": "Trades",
+  "no_of_tokens": "# of Tokens",
+  "volume_24h_total": "DEX Traded Volume (24H)",
+  "volume_24h_total_description": "Total USD value of trades involving trusted assets on the DEX in the past 24h",
+  "market_cap_metric_description": "Circulating supply multiplied by price of all trusted tokens",
+  "market_cap_description": "Circulating supply multiplied by price",
+  "24h_description": "Price change over the past 24h",
+  "volume_description": "Total USD value of all DEX trades in the past 24h",
+  "trades_description": "Number of DEX trades executed in the past 24h",
+  "tvl_description": "USD value of token in AMMs",
```
kuan121, line 604: For my understanding, regarding TVL, do we only consider the token’s value in AMM pool?

src/containers/Tokens/TokensTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+      </td>
+      <td className="holders">
+        {token.holders ? parseAmount(token.holders) : DEFAULT_EMPTY_VALUE}
+      </td>
+      <td className="tvl">
+        {token.tvl_usd
+          ? parseCurrencyAmount(token.tvl_usd)
+          : DEFAULT_EMPTY_VALUE}
+      </td>
+      <td className="market-cap">
```
kuan121, line 127: From a user experience perspective, would it be better to include a `circulating supply` column alongside the `market cap` column?

src/containers/Tokens/TokensTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+  sortOrder: SortOrder
+  setSortField: (field: string) => void
+  setSortOrder: (order: SortOrder) => void
+  setPage: (page: number) => void
+}
+
+const DEFAULT_DECIMALS = 1
+const DEFAULT_EMPTY_VALUE = '--'
+
+export const parseCurrencyAmount = (
```
kuan121: Can we move all parsing and formatting logic to [NumberFormattingUtils](https://github.com/ripple/explorer/blob/staging/src/containers/shared/NumberFormattingUtils.ts) so that it can be unit tested for correctness and reused to prevent inconsistencies and duplicated effort.

cc: @achowdhry-ripple

src/containers/Tokens/index.tsx
```diff
@@ -0,0 +1,247 @@
+import axios from 'axios'
```
kuan121, line 1: Do we have any new components and functionalities can be re-used by Ashray for the Token Page?

pdp2121: Yes there will be plenty, he will rebase his on top of my branch so that should be the lean up needed for his PR

src/containers/Tokens/TokensTable.tsx
pdp2121: I think let's just stick to the vetted design. Also market cap also related to price and I think price at leftmost and marketcap rightmost make sense

public/locales/en-US/translations.json
pdp2121: Yes, I think that's the definition of TVL. I have talked with Julian and the DS team about this

src/containers/Tokens/index.tsx
pdp2121: We decided to not include non-KYC tokens for the ranking page in this release, but yes I should include the *

public/locales/en-US/translations.json
pdp2121: I'm sticking to the design here

kuan121: Not a blocker, but IMO, we need a better description for our users, and since we understand them better than the designer does, we should take the lead on this.

src/containers/shared/components/middleEllipsis.tsx
pdp2121: Yes initially I want it to be dynamic based on the screen size so user can see the full account when the screen is large and when ellipsis happens we always keep the last 3 chars (user can see more columns on mobile). But since Account page and token page both uses the same approach and this does not consume much space, probably I will use shortenAccount as well and remove this complicated logic

src/containers/Tokens/tokens.scss
```diff
@@ -0,0 +1,285 @@
+@use '../shared/css/variables' as *;
```
kuan121, line 1: It seems we’re not following the font sizes specified in the design. Most noticeably, the text in the Token table items is 12 px, which feels too small in my opinion.

server/routes/v1/tokens.js
pdp2121: Good call. I have fixed in https://github.com/ripple/explorer/pull/1248/commits/2cac44d2c1ff98658e869c2f645682e77059cfb3. If fetching failed we will use stale data

src/containers/App/App.tsx
pdp2121: I have removed all inner `TooltipProvider` in https://github.com/ripple/explorer/pull/1248/commits/075bbdb30187a804457cf45fd8ba9490c3e4c855. We should only have one provider at the top level following React Context best practice. All should still working fine as checked

src/containers/shared/components/Account.tsx
pdp2121: Rebased to use your change

server/routes/v1/tokens.js
pdp2121: Good call! Fixed in https://github.com/ripple/explorer/pull/1248/commits/141a35207f39ec299063cfce1a31ae1ac8779bca

src/containers/shared/components/middleEllipsis.tsx
pdp2121: Fixed in https://github.com/ripple/explorer/pull/1248/commits/816a60c3b16b1eb626b999a1c46ff25aa9537d18

src/containers/Tokens/index.tsx
pdp2121: I left both just in case since we have it as market_cap in LOS but XRPLMeta uses marketcap. But we should just use `market_cap`, fixed

public/locales/en-US/translations.json
pdp2121: The design has been vetted by Julian, so we should run it through him again if we want to change a message, since things can be controversial

src/containers/Tokens/TokensTable.tsx
pdp2121: Fixed in https://github.com/ripple/explorer/pull/1248/commits/d43cd13c6fcc763bf6c528af2a1e12fe38afdf3a

src/containers/shared/css/tooltip.scss
pdp2121: This styling only apply to plain text tooltips that we recently added, and should not affect other tooltips

src/containers/shared/components/Pagination/index.ts
pdp2121: This is common practice in JS/TS for clean and consistent import paths (otherwise the import would be `from './Pagination/Pagination'`)

src/containers/Tokens/index.tsx
pdp2121: Fixed in https://github.com/ripple/explorer/pull/1248/commits/8875e3e92ad770aa8ae110e8371752e646fc0e1a

server/routes/v1/index.js
pdp2121: Agree! Fixed in https://github.com/ripple/explorer/pull/1248/commits/ee317e40baf75aebad3016b7f48706685cb949fb to use destructuring import

src/containers/shared/components/TokenSearchResults/TokenSearchResults.tsx
pdp2121: Great catch! Fixed in https://github.com/ripple/explorer/pull/1248/commits/605a5ad43d8a89587f96d9b2bc2f0c1bd346471b

src/containers/Tokens/tokens.scss
pdp2121: Agree! I will use font-size 14px for screen larger than tablet-landscape-up and 12px for smaller screen so that people can see more columns

src/containers/shared/css/tooltip.scss
kuan121: I noticed a difference in the tooltips for the metric fields on the landing page between:
1. https://livenet.dev.ripplex.io/
2. https://livenet.xrpl.org/

It seems that either your change or Ashray’s change is affecting this. In my view, this involves the CSS of a shared component, so it shouldn’t be modified unless there’s a clear reason and we fully understand the impact.

.env.example
```diff
@@ -16,8 +16,8 @@ VITE_XAHAU_TESTNET_LINK=
 VITE_CUSTOMNETWORK_LINK=
 VITE_VALIDATOR=vl.ripple.com
 
-#External data source - XRPLMeta node for token search results
-XRPL_META_URL=s1.xrplmeta.org
```
kuan121: Ashray is using this to call the holder. Let's keep it.

kuan121, line 20: Ashray is using this to call the holders endpoint. Let's keep it.

pdp2121: Sure

server/routes/v1/tokens.js
```diff
@@ -27,101 +26,185 @@ const parseCurrency = (currency) => {
     : currency
 }
 
-async function fetchXRPLMetaTokens(offset) {
-  log.info(`caching tokens from ${process.env.XRPL_META_URL}`)
+const calculateMetrics = (tokens) => ({
+  count: tokens.length,
+  market_cap: tokens
+    .reduce((sum, token) => {
```
achowdhry-ripple, line 32: not following this part, what is sum and why is it being added to the metrics here?

server/routes/v1/tokens.js
```diff
@@ -27,101 +26,185 @@ const parseCurrency = (currency) => {
...
+      return cap > 0 ? sum + cap : sum
+    }, 0)
+    .toFixed(6),
+  volume_24h: tokens
+    .reduce((sum, token) => sum + Number(token.daily_volume_usd || 0), 0)
+    .toFixed(6),
+  stablecoin: tokens
+    .reduce((sum, token) => {
+      const cap = Number(token.market_cap_usd) || 0
+      return token.asset_subclass === 'stablecoin' && cap > 0 ? sum + cap : sum
```
achowdhry-ripple, line 43: this logic is now being handled in the trusted-tokens side, could we remove it from the frontend now?

server/routes/v1/tokens.js
```diff
@@ -27,101 +26,185 @@ const parseCurrency = (currency) => {
...
+      (a, b) => Number(b.holders ?? 0) - Number(a.holders ?? 0),
+    )
 
-  cachedTokenSearchList.tokens = allTokensFetched.filter(
-    (result) =>
-      (result.metrics.trustlines > 50 &&
-        result.metrics.holders > 50 &&
-        result.metrics.marketcap > 0 &&
-        result.metrics.volume_7d > 0) ||
-      result.meta.issuer.trust_level === 3,
```
achowdhry-ripple, line 70: are we now removing all minimum filters? or is this done on the LOS side

server/routes/v1/tokens.js
```diff
@@ -27,101 +26,185 @@ const parseCurrency = (currency) => {
...
   }
 }
+
+const getAllTokens = async (req, res) => {
+  try {
+    log.info('getting tokens list for search')
+    while (cachedTokenList.tokens.length === 0) {
+      // eslint-disable-next-line no-await-in-loop -- necessary here to wait for cache to be filled
+      await sleep(1000)
+    }
```
achowdhry-ripple, line 193: should we have a timeout limit here to be safe?

src/containers/Tokens/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+            setSortOrder={setSortOrder}
+            setPage={setPage}
+          />
+          <div className="footnote">
+            <Trans
+              i18nKey="tokens_footnote"
+              components={{
+                Link: (
+                  <Link to="https://xrplmeta.org">
+                    XRPL Meta&apos;s guidelines
```
achowdhry-ripple: nit: can you add the link text to translations

achowdhry-ripple (approved):
lgtm. We will need to merge the token page PR into this branch first before merging

server/routes/v1/tokens.js
pdp2121: It's done on LOS side

server/routes/v1/tokens.js
pdp2121: It's the sum of market_cap of all tokens, of all volume 24h, and market_cap of all stablecoin

server/routes/v1/tokens.js
pdp2121: This is to calculate sum of market_cap of all stablecoin for metrics, not related to individual market_cap

src/containers/Tokens/index.tsx
pdp2121: Good catch. Fixed

server/routes/v1/tokens.js
pdp2121: Fixed

PR #1252, opened 2025-10-24
Empty table state component, remove search banner

achowdhry-ripple:

Minor cosmetic changes:
- AMM TVL now shows -- when there are no AMMs found
- When there are no dex trades or transfers found, empty state looks consistent with explorer design
- empty state separated into reusable component
- remove search banner ahead of token page launch

before:
Screenshot 2025-10-24 at 11 28 53 AM

after:
Screenshot 2025-10-24 at 11 28 18 AM

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization


src/containers/Token/components/DexTradeTable/styles.scss
```diff
@@ -57,6 +57,30 @@
   }
 }
 
+// Empty state message styling
+.empty-state-message {
```
kuan121: Should we extract this empty state thing into a separate component to avoid duplicating the HTML and SCSS multiple times? Also, maybe check if we already have a similar component in the codebase? I’m planning to create the same tables for the MPT Object page, so it’d be great if we could make this reusable.

pdp2121: Yes we should have it reusable in shared folder

achowdhry-ripple: done

src/containers/shared/components/EmptyStateMessage/emptyStateMessage.scss
```diff
@@ -0,0 +1,26 @@
+@use '../css/variables' as *;
```
pdp2121: Can you put this file into the css folder?

achowdhry-ripple: done

kuan121 (approved):

PR #1253, opened 2025-10-30
fix: add non-mainnet token page

pdp2121:

We are not currently fetching non-mainnet data in LOS. This causes the token page for non-mainnet to break. Use the old token page for non-mainnet networks until they are supported by LOS.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### DIFF
--- a/src/containers/TokenNonMain/TokenHeader/index.tsx
+++ b/src/containers/TokenNonMain/TokenHeader/index.tsx
@@ -177,7 +177,9 @@ export const TokenHeader = ({
   return (
     <div className="box token-header-non-main">
       <div className="section box-header">
-        <Currency currency={currency} />
+        <h1>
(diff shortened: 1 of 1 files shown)


src/rippled/token.ts
```diff
@@ -0,0 +1,71 @@
+import logger from './lib/logger'
```
achowdhry-ripple, line 1: nit: can you add a TODO comment to remind us to remove this once we switch to a permanent solution?

achowdhry-ripple (approved):

src/rippled/token.ts
pdp2121: I think we can add an issue and have the details there instead of a TODO

PR #1264, opened 2025-11-28
MPT Object Page Revamp

kuan121:
1. Revamp the MPT Object page
   * Header has the `General Overview`, `Market Data` and `Settings` sections
   * Add tables displaying all transactions and transfers involving the MPT
   * Add a holders table, sorted in descending order by amount
2. Refactor the IOU Token page in order to reuse its components for MPT since both MPT and IOU are tokens. This involves moving reusable components to `src/containers/Token/shared/` and placing IOU specific logic in `src/containers/Token/IOU/`

### Context of Change

The current XRPL Explorer Multi-Purpose Token (MPT) page provides only on-ledger metadata of MPT tokens, i.e., Token ID, Issuer, Flags. It lacks interesting information such as market and transaction data. This PR revamp the XRPL Explorer MPT Token page to deliver a more organized, performant, and informative experience. The new page will:
1. Present data in an organized format, e.g., General Overview, Market Data, Settings, Transactions/transfers
2. Show relevant transaction and transfer data
3. Maintain fast load times, even for tokens with high data volume

### Type of Change

- [ ] Documentation Updates
- [x] Translation Updates
- [ ] Release

### Codebase Modernization

## Before / After
1. This MPT has metadata that follows the recommended metadata standards. Its metadata was provided using short keys. The new page shows the metadata in long-key format, but the old page shows the short keys.
    * New: https://testnet.dev.ripplex.io/mpt/00C81C870778A061899843D4E120E4CDF71BE9613D49A3B7
    * Old: https://testnet.xrpl.org/mpt/00C81C870778A061899843D4E120E4CDF71BE9613D49A3B7

2. This MPT has metadata that does not follow the recommended metadata standards.
   * New: https://testnet.dev.ripplex.io/mpt/00C82BB9B57025920E7575FB97A3930619B73F3DFC630C29
   * Old: https://testnet.xrpl.org/mpt/00C82BB9B57025920E7575FB97A3930619B73F3DFC630C29

3. This MPT's metadata is a non JSON string
   * New: https://testnet.dev.ripplex.io/mpt/00C84CF7586BF37DFD607FA80A5A5FB07D8DD3AAEBE5E831
   * Old: https://testnet.xrpl.org/mpt/00C84CF7586BF37DFD607FA80A5A5FB07D8DD3AAEBE5E831

4. This MPT doesn't have metadata.
   * New: https://testnet.dev.ripplex.io/mpt/00C88EE826D4F4C72605258FB1A895A16343C11CADD98B55
   * Old: https://testnet.xrpl.org/mpt/00C88EE826D4F4C72605258FB1A895A16343C11CADD98B55

5. This MPT has transfer data
   * New: https://livenet.dev.ripplex.io/mpt/057D54EE52340A529B506C4B630ADE0EEE8DA51654A41628
   * Old: https://livenet.xrpl.org/mpt/057D54EE52340A529B506C4B630ADE0EEE8DA51654A41628

For reference:
1. MPT Token Figma Design: https://www.figma.com/design/ANoKE7Qr7gsNrRDAyjVTkH/Tokens-and-search?node-id=2736-822&t=0GhNSGgvtMrCWYTy-0
2. RLUSD Token Page: https://livenet.xrpl.org/token/524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De

## Test Plan

* [Done] Unit Tests
* [Done] Manual Tests

### DIFF
--- a/src/containers/Token/MPT/Header/Metadata.tsx
+++ b/src/containers/Token/MPT/Header/Metadata.tsx
@@ -18,7 +18,7 @@ export const Metadata = ({
         {isString ? (
           <div className="metadata-string">{decodedMPTMetadata}</div>
         ) : (
-          <JsonView data={decodedMPTMetadata} showExpandButton />
+          <JsonView data={decodedMPTMetadata} showExpandButton showBackground />
         )}
       </div>
     </div>


src/containers/App/index.tsx
```diff
@@ -34,13 +34,13 @@ import { Ledger } from '../Ledger'
 import { AccountsRouter } from '../Accounts/AccountsRouter'
 import { Transaction } from '../Transactions'
 import { Validator } from '../Validators'
-import { Token } from '../Token'
+import { Token } from '../Token/IOU'
```
Patel-Raj11: nit: Should we rename export from `../Token/IOU` to IOU and use `IOU` here?

src/containers/Token/MPT/Header/Header.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ * Check if MPT metadata conforms to XLS-0089 standard
+ * Required fields: ticker (t), name (n), icon (i), asset_class (ac), issuer_name (in)
+ */
+const isXLS0089Compliant = (metadata?: MPTMetadata): boolean => {
+  if (!metadata) return false
+
+  // Check for required fields (both short and long key formats)
+  const hasTicker = !!(metadata.t || metadata.Ticker)
```
Patel-Raj11: Long form keys are in snake case as per the spec.

src/containers/Token/MPT/Header/Header.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const hasAssetClass = !!(metadata.ac || metadata.AssetClass)
+  const hasIssuerName = !!(metadata.in || metadata.IssuerName)
+
+  return hasTicker && hasName && hasIcon && hasAssetClass && hasIssuerName
+}
+
+/**
+ * Extract metadata fields supporting both short and long key formats
+ */
+const extractMetadataFields = (metadata?: MPTMetadata) => {
```
Patel-Raj11: 1. I don't see logic that extracts websiteUrl and socialLinks from `formatMPTIssuance`. Did I miss the intent here?
2. Ideally `isXLS0089Compliant` should check if `ticker` is up to 6 characters long and other checks as per XLS standard. Are we doing just the basic checks here?
3. We should prepend https:// to any URL where it's missing. I doubt that with the current logic Logo will be fetched and displayed.
4. Social URL does not takes me to the external website. Example MPT on testnet - 00C4CE21F66F8DB2F35177BB412357F68E4D117A5456D4CF

public/locales/en-US/translations.json
```diff
@@ -762,6 +762,46 @@
...
+  "mpt_page.asset_class": "Asset Class",
+  "mpt_page.all_tx": "All Token Transactions",
+  "mpt_page.transfers_tx": "Transfers",
+  "mpt_page.holders_table": "Holders",
+  "mpt_page.holders_rank": "Rank",
+  "mpt_page.holders_account": "Account",
+  "mpt_page.holders_num_tokens": "# of Tokens",
+  "mpt_page.holders_percent_supply": "% of Supply",
+  "mpt_page.holders_no_holders": "No holders found",
+  "mpt_page.transfers_no_transfers": "No transfers found",
```
achowdhry-ripple: could these reuse the ones from token page, or could we consolidate them into one?

public/locales/en-US/translations.json
```diff
@@ -762,6 +762,46 @@
...
+  "mpt_page.metadata_warning": "This MPT's metadata does not conform to the XLS-0089 standard. Some information may be missing or displayed incorrectly.",
+  "mpt_page.social_links": "Social Links",
+  "mpt_page.enter_id": "Enter a MPT Issuance ID in the search box",
+  "mpt_page.no_holders": "No holders found",
+  "mpt_page.no_transfers": "No transfers found",
```
achowdhry-ripple: these seem like duplicates, could we use one definition for these copies?

public/locales/en-US/translations.json
```diff
@@ -762,6 +762,46 @@
...
+  "mpt_page.can_lock": "Can Lock",
+  "mpt_page.require_auth": "Require Auth",
+  "mpt_page.can_escrow": "Can Escrow",
+  "mpt_page.can_trade": "Can Trade",
+  "mpt_page.can_transfer": "Can Transfer",
+  "mpt_page.can_clawback": "Can Clawback",
+  "mpt_page.issuer_name": "Issuer Name",
+  "mpt_page.description": "Description",
+  "mpt_page.asset_class": "Asset Class",
+  "mpt_page.all_tx": "All Token Transactions",
```
achowdhry-ripple: a lot of these seem like duplicated copy, could we make the names more generic and reuse them in multiple places for cleanliness?

public/locales/en-US/translations.json
```diff
@@ -762,6 +762,46 @@
   "token_page.transfers_no_transfers": "No transfers found",
   "token_page.token_label": "Token",
   "token_page.category_text": "IOU",
-  "refresh_data": "Refresh data"
+  "refresh_data": "Refresh data",
+  "mpt_page.token_label": "Token",
+  "mpt_page.category_text": "MPT",
+  "mpt_page.mpt_issuance_id": "MPT Issuance ID",
+  "mpt_page.issuer": "Issuer",
```
achowdhry-ripple: nit: add nulls for new definitions in the other translations files

kuan121: I will add translations for other languages in a subsequent commit.

public/locales/en-US/translations.json
kuan121: OK. I will consolidate them.

src/containers/App/index.tsx
kuan121: Fixed in [this commit](https://github.com/ripple/explorer/pull/1264/changes/2506b510b8bfaa10d232f3f12cbb7eb91a1240e1)

public/locales/en-US/translations.json
kuan121: When you said “multiple places,” did you mean locations beyond the IOU and MPT pages? I was planning to reuse them only for MPT and IOU, not elsewhere. For example, I would rename this to token_page.all_tx so it can be shared by both pages. Let me know if you meant something different.

src/containers/Token/MPT/Header/Header.tsx
```diff
@@ -0,0 +1,264 @@
...
+ */
+const isXLS0089Compliant = (metadata?: MPTMetadata): boolean => {
+  if (!metadata) return false
+
+  // Check for required fields (both short and long key formats)
+  const hasTicker = !!(metadata.t || metadata.Ticker)
+  const hasName = !!(metadata.n || metadata.IssuerName)
+  const hasIcon = !!(metadata.i || metadata.LogoURL)
+  const hasAssetClass = !!(metadata.ac || metadata.AssetClass)
+  const hasIssuerName = !!(metadata.in || metadata.IssuerName)
```
achowdhry-ripple: Am a little confused here -- by the spec design, does the metadata standard support both the long form keys and the short form keys? I was under the impression it only supports short form and we extract it and save as long form for readability?

src/containers/Token/MPT/Header/styles.scss
```diff
@@ -0,0 +1,335 @@
...
+      width: 12px;
+      height: 12px;
+    }
+  }
+
+  .social-dropdown-menu {
+    position: absolute;
+    z-index: 100;
+    top: 100%;
+    right: 0;
```
achowdhry-ripple, line 92: It seems like we had to dupllictae the full header boxes component because some of the fields are different, but could the styles be shared between the two since they have the same UX?

public/locales/en-US/translations.json
kuan121: Yes, I will reuse one for both MPT and IOU

src/containers/Token/MPT/Header/Header.tsx
kuan121: 1. Fixed. Note that, for MPT, we don't differentiate between websiteURL and social link. The metadata has all the websites under `uris`.
2. I copy/paste the code in [xrpl.js](https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/utils/mptokenMetadata.ts) and it should work now.
3. Fixed. I used the `DomainLink` component, which will prepend https:// if missing.
4. Fixed.

src/containers/Token/MPT/Header/Header.tsx
kuan121: Fixed.

src/containers/Token/MPT/Header/Header.tsx
kuan121: It support both short and long, but it advocates for short. That's why the spec use short keys as an example. If users provide metadata with short key, Explorer will display the corresponding long keys for better user experience.

src/containers/Token/MPT/Header/AdditionalInfo.tsx
```diff
@@ -0,0 +1,38 @@
+import { useTranslation } from 'react-i18next'
```
kuan121: I'm confirming with Julian now, but this component might be removed since we already show the metadata JSON anyway.

src/containers/Token/MPT/Header/index.tsx
```diff
@@ -0,0 +1,261 @@
...
+                    displayText={shortenAccount(issuer)}
+                  />
+                </div>
+              )}
+              <span className="paren">)</span>
+            </div>
+          )}
+        </div>
+
+        {allUris.length > 0 && (
```
Patel-Raj11, line 169: I think `00C81C870778A061899843D4E120E4CDF71BE9613D49A3B7` MPT has more than 1 uri and there is a dropdown icon as well. Did we meant to display all of those?

kuan121: Correct. We display all URLs in the Metadata section. If there is more than one, we show a dropdown; if there is only one, we display it directly without a dropdown. If there are no URLs, the component is hidden.

src/containers/Token/MPT/Header/MarketData.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const MarketData = ({
+  maxAmt,
+  outstandingAmt,
+  assetScale,
+}: MarketDataProps): JSX.Element => {
+  const { t } = useTranslation()
+
+  const formattedSupply = parseAmount(
+    convertScaledPrice(BigInt(maxAmt || '0'), assetScale ?? 0),
```
Patel-Raj11, line 19: Should this be `outstandingAmt`?  `outstandingAmt` is used in the MPT issued table.

src/containers/Token/MPT/Header/index.tsx
Patel-Raj11: Ya, actually that particular page is displaying only the first URL. I think we need to iterate and display all.

src/containers/Token/MPT/Header/MarketData.tsx
kuan121: Good catch.

The value in the MPT Issued table should be updated to Circulating Supply.
On the MPT page, we currently show Max Supply and Circ Supply. Max Supply uses maxAmt, and Circ Supply uses outstandingAmt. I will:
1. Update the Issued MPT table to use Circ Supply.
2. [Need to check with Julian] Update the SUPPLY label on the MPT page to MAX SUPPLY.

Patel-Raj11: I think Issued MPT table is already using outstandingAmt thus Circulating Supply.

src/containers/Token/MPT/Header/index.tsx
kuan121: https://testnet.dev.ripplex.io/mpt/00C81C870778A061899843D4E120E4CDF71BE9613D49A3B7 There is a dropdown and you can see all of them.

Patel-Raj11: Somehow clicking it was not showing the drop down before, maybe I was clicking just below the arrow. Ya, I see now it working properly.

src/containers/Token/MPT/Header/MarketData.tsx
kuan121: The current state
1. On Issued MPT table,  the table column shows **SUPPLY**, but use `outstandingAmt`. 
2. On MPT Object page, we show a **SUPPLY** field, but we use `maxAmt`.

What I meant is I will update the table column of Issued MPT table from **SUPPLY** to **CIRC SUPPLY**

Patel-Raj11: I see, thanks for clarifying.

public/locales/en-US/translations.json
achowdhry-ripple: that sounds good to me

src/containers/Token/shared/components/HoldersTable/HoldersTable.tsx
```diff
@@ -66,13 +75,21 @@ export const HoldersTable = ({
           displayText={shortenAccount(holder.account)}
         />
       </td>
-      <td className="tx-ledger">{parseIntegerAmount(holder.balance)}</td>
```
kuan121: Confirm with Ashray this is expected: IOU holder balance is only integer?

PR #1232, opened 2025-08-19
refactor: Streams to make homepage startup instantly

ckniffen:
## High Level Overview of Change

Switch streams logic to a context/provider based architecture.  This reduces a bunch of redundant logic and cleans up the state change tree. The moment the page loads the most recent validated ledger is loaded in to make the page look populated from the get go.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

- [x] Updated files to React Hooks
- [x] Updated files to TypeScript
- [ ] Updated tests to React Testing Library

PR #1265, opened 2025-12-02
Fix component load bugs

Patel-Raj11:
This PR addresses following items:
1. Loads Validator table with VHS data if data from live websocket connection is not available yet.
2. Load metrics from backend that is already calculated in backed when homepage loads.
3. Show loading icon, if websocket stream is not yet established on homepage.
4. Fix bug that accessed an undefined variable when hover over nUNL element when there are no nNUL present.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

## Before / After
Before and after are captured in this video

https://github.com/user-attachments/assets/98475aff-68dd-4957-b5b1-960c512d3b83

## Test Plan
Added and updated tests

### DIFF
--- a/src/containers/Ledgers/LedgerMetrics.tsx
+++ b/src/containers/Ledgers/LedgerMetrics.tsx
@@ -144,7 +144,7 @@ export const LedgerMetrics = ({
                 {renderTextTooltip(key)}
               </div>
             </div>
-            <span>{content ?? '--'}</span>
+            <span>{content}</span>
           </div>
         )
       } else {


src/containers/Network/Validators.tsx
```diff
@@ -36,42 +36,30 @@ export const ValidatorsData = () => {
...
+        fromStream[d].ledger_index
       ) {
         // VHS uses `current_index` instead of `ledger_index`
         // If `ledger_index` isn't defined, then we're still using the VHS data,
         // instead of the Streams data
-        newData.ledger_index = validatorsFromValidations[d].ledger_index
-      }
-      if (newData.current_index == null) {
-        newData.signing_key = validatorsFromValidations[d].signing_key
-      }
```
achowdhry-ripple, line 70: why can we remove this part? it seems like the properties that are being set have changed with this logic change now

Patel-Raj11: `signing_key` was never a property on `validatorsFromValidations[d]`. I don't know why this was added during recent refactor, but if we go further down the rendering part, `master_key` is given priority irrespective of the source of data.

src/containers/Ledgers/test/LedgersPage.test.js
```diff
@@ -147,11 +147,12 @@ describe('Ledgers Page container', () => {
     expect(wrapper.find('.txn').length).toBe(0)
 
     server.send(prevLedgerMessage)
+    await sleep(260)
     wrapper.update()
     expect(wrapper.exists()).toBe(true)
     expect(wrapper.find('.ledgers').length).toBe(1)
     expect(wrapper.find('.ledger-list').length).toBe(1)
-    expect(wrapper.find('.ledger').length).toBe(1)
+    expect(wrapper.find('.ledger').length).toBe(2)
```
ckeshava, line 155: What are the two .ledger react components in the wrapper? Since we send one `prevLedgerMessage`, shouldn't we expect to see one of the `.ledger` react component?

src/containers/Ledgers/LedgerMetrics.tsx
```diff
@@ -110,7 +110,7 @@ export const LedgerMetrics = ({
         content = renderXRP(data[key], language)
       } else if (key === 'ledger_interval' && data[key] !== '--') {
         content = `${data[key]} ${t('seconds_short')}`
-      } else if (key === 'nUnl' && data[key]?.length === 0) {
+      } else if (key === 'nUnl' && (!data[key] || data[key]?.length === 0)) {
```
ckeshava, line 113: Can you add a unit test that simulates this case? when there is no element on the nUNL and ensure that the react component displays `--`?

From the video in the Github PR, I didn't fully understand the solution for this problem? If the nUNL list is empty, are you displaying "--" (or) are you removing that widget from the display?

src/containers/shared/components/Streams/StreamsProvider.tsx
```diff
@@ -78,9 +78,7 @@ export const StreamsProvider: FC = ({ children }) => {
     { enabled: socket.getState().online },
   )
   const { data: serverRunningMetrics, refetch: refetchServerRunningMetrics } =
-    useQuery<string[]>('runningMetrics', () => fetchMetrics(), {
-      enabled: socket.getState().online,
```
ckeshava, line 82: `enabled` ensures that useQuery is not executed upon every reload of the react component. Since ledgers are accumulated from the web-socket stream, the `Ledgers` component might be reloaded multiple times.

Here are the docs for this method: https://tanstack.com/query/v5/docs/framework/react/reference/useQuery

src/containers/shared/components/Streams/StreamsProvider.tsx
```diff
@@ -55,7 +55,7 @@ const truncateLedgers = (ledgers: Record<string, Ledger>, count) =>
 export const StreamsProvider: FC = ({ children }) => {
   // In custom mode we populate metrics from ledgers loaded into memory
   const useServerMetrics = process.env.VITE_ENVIRONMENT !== 'custom'
-  const [ledgers, setLedgers] = useState<Record<number, Ledger>>([])
+  const [ledgers, setLedgers] = useState<Record<number, Ledger>>({})
```
ckeshava, line 58: hmm, nice catch! I wonder why the tsc did not complain about the type mismatch before. There must be some form of implicit type coercion in the compiler.

src/containers/Network/Validators.tsx
ckeshava: If `signing_key` property is not defined, then this line will not render the validator component. https://github.com/ripple/explorer/blob/8d317d957ee04fd8551a40ac1156e7e221e789dc/src/containers/Validators/index.tsx#L207

src/containers/Network/Validators.tsx
```diff
@@ -36,42 +36,30 @@ export const ValidatorsData = () => {
   const [feeSettings, setFeeSettings] = useState<FeeSettings>()
```
ckeshava, line 36: Amongst the diff'ed lines in this file, what is the fix for this bug:
> Loads Validator table with VHS data if data from live websocket connection is not available yet.

I'm trying to understand the goal of the changes in this file.

src/containers/Network/Validators.tsx
Patel-Raj11: > If signing_key property is not defined, then this line will not render the validator component.

`master_key` is given preference, and it will render the component. Won't render when both are not present.
In any case `signing_key` was not set by the code prior to https://github.com/ripple/explorer/pull/1232 and it's incorrectly set currently. So why the concern now?

src/containers/Ledgers/test/LedgersPage.test.js
Patel-Raj11: For the same reason 2 components were expected here - https://github.com/ripple/explorer/blob/8d317d957ee04fd8551a40ac1156e7e221e789dc/src/containers/Ledgers/test/LedgersPage.test.js#L253

One is the latest validated ledger and another from `prevLedgerMessage` on same reasoning - https://github.com/ripple/explorer/blob/8d317d957ee04fd8551a40ac1156e7e221e789dc/src/containers/Ledgers/test/LedgersPage.test.js#L245

src/containers/Ledgers/LedgerMetrics.tsx
Patel-Raj11: > From the video in the Github PR, I didn't fully understand the solution for this problem? If the nUNL list is empty, are you displaying "--" (or) are you removing that widget from the display?

I am removing the component. `--` is now no more necessary, will remove in next commit. I added it first to fix an alignment issue and then fixed a major bug seen in the video, that leads to addition of `--` change no more necessary.

src/containers/shared/components/Streams/StreamsProvider.tsx
Patel-Raj11: > enabled ensures that useQuery is not executed upon every reload of the react component

enabled ensure that the query function executes, when it needs to execute based (query key change, initial mount etc). and that's what we want and currently it does that once we connect to clio from frontend. 

This change makes, it do that even before we establish the frontend connection, since the query function fetches the data from backend.

src/containers/Network/Validators.tsx
Patel-Raj11: Goal is to return the data from VHS even if data from clio is yet not available. It used to do that previous to https://github.com/ripple/explorer/pull/1232. You can test it by running it locally on a previous commit.

The difference in livenet tab and localhost tab in the video will make it more clear.

PR #1267, opened 2025-12-05
Run `prettier` on all Explorer files

mvadari:

This PR runs `prettier` on all Explorer files (it seems that it wasn't properly run on non-JS/TS files before). It also configures CI and CLI scripts to check `prettier` as well, and sets up some basic pre-commit settings.

This PR does not make any code changes, only linting changes.

### Context of Change

General repo cleanup

### Type of Change

- [x] Chore

### Codebase Modernization

N/A

## Test Plan

Works locally. CI passes.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Use Node.js ${{ env.NODE_VERSION }}
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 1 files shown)


public/locales/ca-CA/translations.json
```diff
@@ -729,22 +729,6 @@
   "txn_sec_description": null,
   "load_fee_description": null,
   "nUnl_description": null,
-  "volume": null,
```
mvadari: These were all duplicates

.github/dependabot.yml
```diff
@@ -1,8 +1,8 @@
 version: 2
 updates:
```
kuan121, line 2: Thanks for opening the PR! Do you know how to set up an "apply prettier on file save" rule or formatter? Doing it in batch later makes git blame less useful.

mvadari: I have it in my IDE and am not quite sure how to share that. That's also the purpose of the pre-commit setting though - everyone can install it via `pre-commit install` and it'll automatically run when you try to `git commit`.

You can also remove this specific git hash from the git blame history.

kuan121 (dismissed):

PR #1270, opened 2025-12-11
refactor: convert new TS files to async/await

mvadari:

This PR is a fast follow from #1269 to refactor the files converted to TS to use proper async/await structure instead of `then` structure.

There is no functionality change, only refactors.

### Context of Change

Best practices, easier to read/write

### Type of Change

### Codebase Modernization

Already done ;)

## Before / After

No changes

## Test Plan

CI passes.

### DIFF
--- a/src/rippled/lib/summarizeLedger.ts
+++ b/src/rippled/lib/summarizeLedger.ts
@@ -19,7 +19,7 @@ const summarizeLedger = (ledger: any, txDetails: boolean = false) => {
     return summarizeTransaction(d, txDetails)
   })
 
-  // eslint-disable-next-line camelcase - TODO: fix later
+  // eslint-disable-next-line camelcase -- TODO: fix later
   const total_fees =
     ledger.transactions.reduce(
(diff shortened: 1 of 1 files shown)


src/rippled/lib/rippled.ts
kuan121, line 34: Can this be converted to async/await as well?

src/rippled/lib/txSummary/index.ts
```diff
@@ -26,14 +26,18 @@ const getInstructions = (tx: Transaction, meta: TransactionMetadata) => {
 }
```
kuan121, line 26: Should we also do the conversion for the files in https://github.com/ripple/explorer/tree/staging/server?

mvadari: Yeah probably, but IMO that's out of scope of this PR

kuan121 (dismissed):

kuan121 (approved):

PR #1281, opened 2026-01-23
Explorer Lending Protocol Object Page

ckeshava:

This PR implements an Explorer Lending-Protocol Object Page. It provides a readable version of Single-Asset-Vaults to the viewers.

If you would like to interact with the new Lending Protocol Object Page, please run a local instance of the Explorer with this PR's changes and navigate to http://localhost:3000/lend.devnet.rippletest.net:51233/vault/3E7B7873E696C8055302F2EA3669E6C447FB3606B92E6451AF1CD2BACFFB1F0E on your browser.

Refer to this screen recording video to see a visual preview of the new Explorer Vault page: https://github.com/user-attachments/assets/e7707a9f-6bde-40cf-8239-b700cfb02a24

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

## Test Plan

Appropriate tests have been added to the new feature.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -2,7 +2,7 @@ name: Explorer CI
 
 on:
   push:
-    branches: [main, staging]
+    branches: [main]
   pull_request:
   workflow_dispatch:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ You need only to do this once and you have probably already done this if you alr
 ### Clone the repository on your local machine.
 
 ```
-$ git clone git@github.com:username/explorer.git --branch staging
+$ git clone git@github.com:username/explorer.git --branch main
 $ cd explorer
 $ git remote add upstream git@github.com:ripple/explorer.git
 ```
@@ -38,17 +38,17 @@ $ git remote add upstream git@github.com:ripple/explorer.git
 ### Start work in a new branch.
 
 ```
-$ git fetch upstream staging
-$ git checkout staging
+$ git fetch upstream main
+$ git checkout main
 $ git checkout -b your-branch-name
 $ git push --set-upstream origin your-branch-name
 ```
 
-Bring in recent changes to the “staging” branch into your own branch
+Bring in recent changes to the “main” branch into your own branch
 
 ```
-$ git fetch upstream staging
-$ git pull upstream staging
+$ git fetch upstream main
+$ git pull upstream main
 ```
 
 ### Ask to submit code from a branch of your GitHub fork.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![Release](https://img.shields.io/github/v/release/ripple/explorer.svg)](https://github.com/ripple/explorer/releases)
 [![Build](https://github.com/ripple/explorer/actions/workflows/nodejs.yml/badge.svg)](https://github.com/ripple/explorer/actions)
-[![License](https://img.shields.io/github/license/ripple/explorer)](https://github.com/ripple/explorer/blob/staging/LICENSE)
+[![License](https://img.shields.io/github/license/ripple/explorer)](https://github.com/ripple/explorer/blob/main/LICENSE)
 
 # XRPL Explorer
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -277,9 +277,9 @@
       }
     },
     "node_modules/@babel/helper-define-polyfill-provider": {
-      "version": "0.6.6",
-      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.6.6.tgz",
-      "integrity": "sha512-mOAsxeeKkUKayvZR3HeTYD/fICpCPLJrU5ZjelT/PA6WHtNDBOE436YiaEUvHN454bRM3CebhDsIpieCc4texA==",
+      "version": "0.6.7",
+      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.6.7.tgz",
+      "integrity": "sha512-6Fqi8MtQ/PweQ9xvux65emkLQ83uB+qAVtfHkC9UodyHMIZdxNI01HjLCLUtybElp2KY2XNE0nOgyP1E1vXw9w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -3076,18 +3076,18 @@
       }
     },
     "node_modules/@jest/console": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/@jest/console/-/console-30.2.0.tgz",
-      "integrity": "sha512-+O1ifRjkvYIkBqASKWgLxrpEhQAAE7hY77ALLUufSk5717KfOShg6IbqLmdsLMPdUiFvA2kTs0R7YZy+l0IzZQ==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/console/-/console-30.3.0.tgz",
+      "integrity": "sha512-PAwCvFJ4696XP2qZj+LAn1BWjZaJ6RjG6c7/lkMaUJnkyMS34ucuIsfqYvfskVNvUI27R/u4P1HMYFnlVXG/Ww==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@jest/types": "30.2.0",
+        "@jest/types": "30.3.0",
         "@types/node": "*",
         "chalk": "^4.1.2",
-        "jest-message-util": "30.2.0",
-        "jest-util": "30.2.0",
+        "jest-message-util": "30.3.0",
+        "jest-util": "30.3.0",
         "slash": "^3.0.0"
       },
       "engines": {
@@ -3109,9 +3109,9 @@
       }
     },
     "node_modules/@jest/console/node_modules/@jest/types": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/@jest/types/-/types-30.2.0.tgz",
-      "integrity": "sha512-H9xg1/sfVvyfU7o3zMfBEjQ1gcsdeTMgqHoYdN79tuLqfTtuu7WckRA1R5whDwOzxaZAeMKTYWqP+WCAi0CHsg==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/types/-/types-30.3.0.tgz",
+      "integrity": "sha512-JHm87k7bA33hpBngtU8h6UBub/fqqA9uXfw+21j5Hmk7ooPHlboRNxHq0JcMtC+n8VJGP1mcfnD3Mk+XKe1oSw==",
       "dev": true,
       "license": "MIT",
       "peer": true,
@@ -3155,59 +3155,58 @@
       }
     },
     "node_modules/@jest/console/node_modules/jest-util": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/jest-util/-/jest-util-30.2.0.tgz",
-      "integrity": "sha512-QKNsM0o3Xe6ISQU869e+DhG+4CK/48aHYdJZGlFQVTjnbvgpcKyxpzk29fGiO7i/J8VENZ+d2iGnSsvmuHywlA==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/jest-util/-/jest-util-30.3.0.tgz",
+      "integrity": "sha512-/jZDa00a3Sz7rdyu55NLrQCIrbyIkbBxareejQI315f/i8HjYN+ZWsDLLpoQSiUIEIyZF/R8fDg3BmB8AtHttg==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@jest/types": "30.2.0",
+        "@jest/types": "30.3.0",
         "@types/node": "*",
         "chalk": "^4.1.2",
         "ci-info": "^4.2.0",
         "graceful-fs": "^4.2.11",
-        "picomatch": "^4.0.2"
+        "picomatch": "^4.0.3"
       },
       "engines": {
         "node": "^18.14.0 || ^20.0.0 || ^22.0.0 || >=24.0.0"
       }
     },
     "node_modules/@jest/core": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/@jest/core/-/core-30.2.0.tgz",
-      "integrity": "sha512-03W6IhuhjqTlpzh/ojut/pDB2LPRygyWX8ExpgHtQA8H/3K7+1vKmcINx5UzeOX1se6YEsBsOHQ1CRzf3fOwTQ==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/core/-/core-30.3.0.tgz",
+      "integrity": "sha512-U5mVPsBxLSO6xYbf+tgkymLx+iAhvZX43/xI1+ej2ZOPnPdkdO1CzDmFKh2mZBn2s4XZixszHeQnzp1gm/DIxw==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@jest/console": "30.2.0",
+        "@jest/console": "30.3.0",
         "@jest/pattern": "30.0.1",
-        "@jest/reporters": "30.2.0",
-        "@jest/test-result": "30.2.0",
-        "@jest/transform": "30.2.0",
-        "@jest/types": "30.2.0",
+        "@jest/reporters": "30.3.0",
+        "@jest/test-result": "30.3.0",
+        "@jest/transform": "30.3.0",
+        "@jest/types": "30.3.0",
         "@types/node": "*",
         "ansi-escapes": "^4.3.2",
         "chalk": "^4.1.2",
         "ci-info": "^4.2.0",
         "exit-x": "^0.2.2",
         "graceful-fs": "^4.2.11",
-        "jest-changed-files": "30.2.0",
-        "jest-config": "30.2.0",
-        "jest-haste-map": "30.2.0",
-        "jest-message-util": "30.2.0",
+        "jest-changed-files": "30.3.0",
+        "jest-config": "30.3.0",
+        "jest-haste-map": "30.3.0",
+        "jest-message-util": "30.3.0",
         "jest-regex-util": "30.0.1",
-        "jest-resolve": "30.2.0",
-        "jest-resolve-dependencies": "30.2.0",
-        "jest-runner": "30.2.0",
-        "jest-runtime": "30.2.0",
-        "jest-snapshot": "30.2.0",
-        "jest-util": "30.2.0",
-        "jest-validate": "30.2.0",
-        "jest-watcher": "30.2.0",
-        "micromatch": "^4.0.8",
-        "pretty-format": "30.2.0",
+        "jest-resolve": "30.3.0",
+        "jest-resolve-dependencies": "30.3.0",
+        "jest-runner": "30.3.0",
+        "jest-runtime": "30.3.0",
+        "jest-snapshot": "30.3.0",
+        "jest-util": "30.3.0",
+        "jest-validate": "30.3.0",
+        "jest-watcher": "30.3.0",
+        "pretty-format": "30.3.0",
         "slash": "^3.0.0"
       },
       "engines": {
@@ -3237,25 +3236,24 @@
       }
     },
     "node_modules/@jest/core/node_modules/@jest/transform": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/@jest/transform/-/transform-30.2.0.tgz",
-      "integrity": "sha512-XsauDV82o5qXbhalKxD7p4TZYYdwcaEXC77PPD2HixEFF+6YGppjrAAQurTl2ECWcEomHBMMNS9AH3kcCFx8jA==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/transform/-/transform-30.3.0.tgz",
+      "integrity": "sha512-TLKY33fSLVd/lKB2YI1pH69ijyUblO/BQvCj566YvnwuzoTNr648iE0j22vRvVNk2HsPwByPxATg3MleS3gf5A==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
         "@babel/core": "^7.27.4",
-        "@jest/types": "30.2.0",
+        "@jest/types": "30.3.0",
         "@jridgewell/trace-mapping": "^0.3.25",
         "babel-plugin-istanbul": "^7.0.1",
         "chalk": "^4.1.2",
         "convert-source-map": "^2.0.0",
         "fast-json-stable-stringify": "^2.1.0",
         "graceful-fs": "^4.2.11",
-        "jest-haste-map": "30.2.0",
+        "jest-haste-map": "30.3.0",
         "jest-regex-util": "30.0.1",
-        "jest-util": "30.2.0",
-        "micromatch": "^4.0.8",
+        "jest-util": "30.3.0",
         "pirates": "^4.0.7",
         "slash": "^3.0.0",
         "write-file-atomic": "^5.0.1"
@@ -3265,9 +3263,9 @@
       }
     },
     "node_modules/@jest/core/node_modules/@jest/types": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/@jest/types/-/types-30.2.0.tgz",
-      "integrity": "sha512-H9xg1/sfVvyfU7o3zMfBEjQ1gcsdeTMgqHoYdN79tuLqfTtuu7WckRA1R5whDwOzxaZAeMKTYWqP+WCAi0CHsg==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/types/-/types-30.3.0.tgz",
+      "integrity": "sha512-JHm87k7bA33hpBngtU8h6UBub/fqqA9uXfw+21j5Hmk7ooPHlboRNxHq0JcMtC+n8VJGP1mcfnD3Mk+XKe1oSw==",
       "dev": true,
       "license": "MIT",
       "peer": true,
@@ -3350,22 +3348,22 @@
       }
     },
     "node_modules/@jest/core/node_modules/jest-haste-map": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/jest-haste-map/-/jest-haste-map-30.2.0.tgz",
-      "integrity": "sha512-sQA/jCb9kNt+neM0anSj6eZhLZUIhQgwDt7cPGjumgLM4rXsfb9kpnlacmvZz3Q5tb80nS+oG/if+NBKrHC+Xw==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/jest-haste-map/-/jest-haste-map-30.3.0.tgz",
+      "integrity": "sha512-mMi2oqG4KRU0R9QEtscl87JzMXfUhbKaFqOxmjb2CKcbHcUGFrJCBWHmnTiUqi6JcnzoBlO4rWfpdl2k/RfLCA==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@jest/types": "30.2.0",
+        "@jest/types": "30.3.0",
         "@types/node": "*",
         "anymatch": "^3.1.3",
         "fb-watchman": "^2.0.2",
         "graceful-fs": "^4.2.11",
         "jest-regex-util": "30.0.1",
-        "jest-util": "30.2.0",
-        "jest-worker": "30.2.0",
-        "micromatch": "^4.0.8",
+        "jest-util": "30.3.0",
+        "jest-worker": "30.3.0",
+        "picomatch": "^4.0.3",
         "walker": "^1.0.8"
       },
       "engines": {
@@ -3387,28 +3385,28 @@
       }
     },
     "node_modules/@jest/core/node_modules/jest-util": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/jest-util/-/jest-util-30.2.0.tgz",
-      "integrity": "sha512-QKNsM0o3Xe6ISQU869e+DhG+4CK/48aHYdJZGlFQVTjnbvgpcKyxpzk29fGiO7i/J8VENZ+d2iGnSsvmuHywlA==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/jest-util/-/jest-util-30.3.0.tgz",
+      "integrity": "sha512-/jZDa00a3Sz7rdyu55NLrQCIrbyIkbBxareejQI315f/i8HjYN+ZWsDLLpoQSiUIEIyZF/R8fDg3BmB8AtHttg==",
       "dev": true,
       "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@jest/types": "30.2.0",
+        "@jest/types": "30.3.0",
         "@types/node": "*",
         "chalk": "^4.1.2",
         "ci-info": "^4.2.0",
         "graceful-fs": "^4.2.11",
-        "picomatch": "^4.0.2"
+        "picomatch": "^4.0.3"
       },
       "engines": {
         "node": "^18.14.0 || ^20.0.0 || ^22.0.0 || >=24.0.0"
       }
     },
     "node_modules/@jest/core/node_modules/pretty-format": {
-      "version": "30.2.0",
-      "resolved": "https://registry.npmjs.org/pretty-format/-/pretty-format-30.2.0.tgz",
-      "integrity": "sha512-9uBdv/B4EefsuAL+pWqueZyZS2Ba+LxfFeQ9DN14HU4bN8bhaxKdkpjpB6fs9+pSjIBu+FXQHImEg8j/Lw0+vA==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/pretty-format/-/pretty-format-30.3.0.tgz",
+      "integrity": "sha512-oG4T3wCbfeuvljnyAzhBvpN45E8iOTXCU/TD3zXW80HA3dQ4ahdqMkWGiPWZvjpQwlbyHrPTWUAqUzGzv4l1JQ==",
       "dev": true,
       "license": "MIT",
       "peer": true,
@@ -3473,9 +3471,9 @@
       }
     },
     "node_modules/@jest/diff-sequences": {
-      "version": "30.0.1",
-      "resolved": "https://registry.npmjs.org/@jest/diff-sequences/-/diff-sequences-30.0.1.tgz",
-      "integrity": "sha512-n5H8QLDJ47QqbCNn5SuFjCRDrOLEZ0h8vAHCK5RL9Ls7Xa8AQLa/YxAc9UjFqoEDM48muwtBGjtMY5cr0PLDCw==",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/diff-sequences/-/diff-sequences-30.3.0.tgz",
+      "integrity": "sha512-cG51MVnLq1ecVUaQ3fr6YuuAOitHK1S4WUJHnsPFE/quQr33ADUx1FfrTCpMCRxvy0Yr9BThKpDjSlcTi91tMA==",
       "dev": true,
       "license": "MIT",
       "peer": true,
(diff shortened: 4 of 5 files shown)


public/locales/en-US/translations.json
```diff
@@ -69,6 +69,7 @@
   "transaction_legend_toggle_show": "Show Legend",
   "transactions.date_header": "Date/time (UTC)",
   "no_transactions_message": "No transactions found.",
+  "get_vault_transactions_failed": "Unable to load vault transactions at this time.",
```
pdp2121, line 71: Can you add translations with null values to other languages as well?

ckeshava: ok, I have added them here: 047c3f00. Thanks for raising this!

src/rippled/lib/rippled.ts
```diff
@@ -899,6 +904,37 @@ const getLoanBroker = (rippledSocket, loanBrokerId) =>
     return resp.node
   })
 
+// get account loans
+const getAccountLoans = async (
```
kuan121: Is this new method necessary? I noticed that `getAccountObjects(..., 'loan')` and `getAccountObjects(..., 'loan_broker')` are used elsewhere. Perhaps we could just use `getAccountObjects(..., 'loan')`. Additionally, this implementation is missing the marker feature.

src/containers/shared/components/TransactionTable/TransactionTable.tsx
```diff
@@ -41,11 +47,15 @@ export const TransactionTable: TransactionTableComponent = ({
...
       <ol className="transaction-table">
         <li className="transaction-li transaction-li-header">
+          {hasHashColumn && <div className="col col-hash">{t('tx_hash')}</div>}
           {hasTokensColumn && (
             <div className="col col-token"> {t('token')} </div>
           )}
           <div className="col col-account">{t('account')}</div>
           <div className="col col-type">{t('transaction_type')}</div>
+          {hasAmountColumn && (
+            <div className="col col-amount">{t('amount')}</div>
```
kuan121: Do we need the amount column? That information is already displayed in the bottom section of each transaction row. For most rows the value is empty, which ends up creating a sparse column.

src/containers/shared/hooks/useTokenToUSDRate.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Fetches token price from xrplmeta.org API
+ * The API returns the price of the token in XRP
+ */
+const fetchTokenPriceInXRP = async (
+  currency: string,
+  issuer: string,
+): Promise<number> => {
+  const identifier = `${currency}:${issuer}`
+  const response = await fetch(
+    `https://s1.xrplmeta.org/token/${encodeURIComponent(identifier)}`,
```
kuan121: We should use LOS instead, i.e., `${process.env.VITE_LOS_URL}/tokens/${currency}.${issuer}`, since it’s a service under our control. We are gradually moving away from XRPLMeta in Explorer.

Example - https://los.dev.ripplex.io/tokens/USD.rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B

src/containers/Vault/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @param value - The number or string value to format
+ * @param language - Locale string for number formatting
+ * @param options - Optional configuration
+ * @param options.currency - Currency suffix to append (e.g., "XRP")
+ * @param options.prefix - Prefix to prepend (e.g., "$")
+ * @param options.fallback - Value to return if input is invalid (default: "-")
+ * @param options.maxFractionDigits - Maximum fraction digits (default: 2)
+ */
+export const formatCompactNumber = (
```
kuan121: 1. We already have something similar `formatLargeNumber`
2. When it comes to formatting numbers, monetary values and balances displayed on Explorer, make sure that you use the functions defined in `NumberFormattingUitls.ts` to avoid reinventing the wheel and ensure consistency.

src/containers/Vault/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </div>
+    )
+  }
+
+  if (error) {
+    return <Page vaultId={vaultId}>{renderError()}</Page>
+  }
+
+  // Get the account ID for transactions (PseudoAccount or Owner)
+  const transactionAccountId = vaultData?.PseudoAccount || vaultData?.Owner
```
kuan121: 1. `PseudoAccount` or `Account`? https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/vault
2. Why can the account be either of the two? If we use the owner account and the owner also performs an NFT mint transaction, we would end up showing that transaction as well. Is that expected behavior?

src/containers/Vault/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+    return ''
+  }
+
+  return (
+    <Page vaultId={vaultId}>
+      {!vaultId && (
+        <div className="vault-warning">
+          <h2>Enter a Vault ID in the search box</h2>
+        </div>
+      )}
```
kuan121: 1. When will users see this warning? When visiting `https://livenet.xrpl.org/values/`?
2. Can we make this a translation key instead of a hardcoded value?

src/containers/Vault/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+      {vaultId && loading && <Loader />}
+      {vaultId && !loading && vaultData && (
+        <>
+          <div className="vault-title-section">
+            <div className="vault-title-left">
+              <h1 className="vault-title">
+                {decodeVaultData(vaultData.Data) || t('yield_pool')}
```
kuan121: Are we sure that we can use `Data` as the title? https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/vault#data-field-format

src/containers/Vault/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const getAssetCurrencyDisplay = () => {
+    const asset = vaultData?.Asset
+    if (!asset) return ''
+    if (asset.currency) return asset.currency
+    if (asset.mpt_issuance_id) {
+      // For MPT, show truncated ID
+      return `MPT (${asset.mpt_issuance_id.substring(0, 6)}...)`
+    }
+    return ''
+  }
```
kuan121: Are we testing `getAssetCurrencyDisplay` with assets other than XRP? For example, how does it handle an IOU token or an MPT? Specifically, for the token RLUSD, are we confident it won’t display `524C555344000000000000000000000000000000` instead?

src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const VaultHeader = ({ data, vaultId, displayCurrency }: Props) => {
+  const { t } = useTranslation()
+
+  return (
+    <div className="vault-section">
+      <h2 className="vault-section-title">{t('vault')}</h2>
+      <div className="vault-section-divider" />
+      <Details
```
kuan121: Is the Details component necessary? It causes VaultData and Props to be defined twice.

src/containers/Vault/VaultHeader/Details.tsx
```diff
@@ -0,0 +1,243 @@
...
+// Withdrawal policy values from XLS-65d spec
+const WITHDRAWAL_POLICIES: { [key: number]: string } = {
+  1: 'first_come_first_served',
+}
+
+const formatAsset = (asset: VaultData['Asset']): string | React.ReactNode => {
+  if (!asset) return '-'
+  if (asset.currency === 'XRP') return 'XRP'
+  if (asset.mpt_issuance_id) {
+    const truncatedId = `${asset.mpt_issuance_id.substring(0, 8)}...${asset.mpt_issuance_id.substring(asset.mpt_issuance_id.length - 6)}`
```
kuan121: You can reuse `shortenMPTID`

src/containers/Vault/VaultDepositors/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+interface Props {
+  shareMptId: string
+  totalSupply: string | undefined
+  assetsTotal: string | undefined
+  displayCurrency: string
+  asset?: AssetInfo
+}
+
+export const VaultDepositors = ({
```
kuan121, line 27: Instead of reimplementing it, can we just reuse [HoldersTable](https://github.com/ripple/explorer/tree/staging/src/containers/Token/shared/components/HoldersTable) that is used on the IOU/MPT token page? A minor refactoring might be required.

src/containers/Vault/VaultDepositors/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const [currentPage, setCurrentPage] = useState(0)
+  const [markers, setMarkers] = useState<(string | undefined)[]>([undefined])
+
+  const pageSize = 5
+
+  const {
+    data,
+    isFetching: loading,
+    error,
+  } = useQuery<any, Error>(
```
kuan121: If a vault has 100 depositors, will this query return all of them or just 5? Remember, we need to fetch all depositor data before ranking. If we fetch only 5 and rely on lazy loading, the first page might not show the actual top 5 depositors.

src/containers/Vault/VaultLoans/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const { trackException } = useAnalytics()
+  const rippledSocket = useContext(SocketContext)
+  const [selectedBrokerIndex, setSelectedBrokerIndex] = useState(0)
+
+  const { data: loanBrokers, isFetching: loading } = useQuery<
+    LoanBrokerData[] | undefined
+  >(
+    ['getVaultLoanBrokers', vaultPseudoAccount],
+    async () => {
+      const resp = await getAccountObjects(
```
kuan121: The `getAccountObjects` method has a default limit of 400. If an account has more than 400 objects, a single call might not return all `loan_broker` objects. This is because the limit specifies how many **total objects to examine** on the server side, not how many `loan_broker` objects to return. For example, if an account has 800 objects, including 40 `loan_broker` objects, the first request will examine only 400 objects, find and return only 20 `loan_broker` objects—there’s no guarantee you’ll get all of them. To retrieve all `loan_broker` objects, you need to repeatedly call `getAccountObjects` until all account objects have been examined, e.g., `marker` is undefined ([code example](https://github.com/ripple/explorer/blob/staging/src/containers/Accounts/AccountAsset/assetTables/IssuedMPTs.tsx#L34-L54)).

This comment applies to the `getAccountObjects(..., 'loan')` as well.

src/containers/Vault/VaultLoans/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+        trackException(
+          `Error fetching Loan Brokers for account ${vaultPseudoAccount} --- ${JSON.stringify(e)}`,
+        )
+      },
+    },
+  )
+
+  // Fetch loan counts for each broker - must be called before any early returns
+  const loanCountQueries = useQueries(
+    (loanBrokers ?? []).map((broker) => ({
```
kuan121: If we have multiple loan brokers, do we send the `getAccountObjects` requests concurrently or sequentially? If we do it concurrently, we risk being throttled.

src/containers/Vault/VaultLoans/BrokerDetails.tsx
```diff
@@ -0,0 +1,181 @@
...
+    const numAmount = Number(amount)
+    if (Number.isNaN(numAmount)) return amount
+    return tokenToUsdRate > 0 ? String(numAmount * tokenToUsdRate) : undefined
+  }
+
+  // Fetch loans for this broker (used for both table display and default check)
+  const { data: loans, isFetching: loansLoading } = useQuery(
+    ['getBrokerLoans', broker.Account, broker.index],
+    async () => {
+      const resp = await getAccountObjects(
```
kuan121: Can we avoid calling `getAccountObjects(..., 'loan')` for each broker again and instead use the loan data already fetched in `index.tsx` from the parent component?

src/containers/Vault/index.tsx
ckeshava: You can find the tests in the file: `src/containers/Vault/test/Vault.test.tsx`. Please search for the section titled "SECTION 7: Asset Currency Display Tests". (line 560)

1. In case of MPT assets, the MPT issuance ID is truncated as described in line 113 above.
2. In the case of IOU, this code displays the `asset.currency` field of the Vault ledger object. Here is the description of the Vault ledger object fields: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/vault-methods/vault_info#asset-object

To answer your question in broader terms: This design forwards the exact currency code recorded in the Vault. That currency string could be any arbitrary 20-byte string. You can read more about [Non-Standard Currency Codes here.](https://xrpl.org/docs/references/protocol/data-types/currency-formats#currency-codes). How do you propose that we translate that 20-byte string into a human-readable form?

src/containers/Vault/VaultDepositors/index.tsx
ckeshava: Okay, however, at this point of time, it is not possible to view the VaultDepositors section. We have requested the Clio team to provide us with an improved `mpt_holders` API so that we can track the depositors of the Vault.

Is it okay if I keep this suggestion when I actually test this component later? At this point of time, none of the Clio nodes will be able to return the requested `mpt_holders` API data.

src/containers/Vault/VaultLoans/index.tsx
ckeshava: I couldn't find references to concurrency on the official documentation of the react-query project. However, [based on this Github discussion](https://github.com/TanStack/query/discussions/4943), it appears that all the API calls are fired off at the same time.

Even if we are throttled by the Clio node, this API call will retry and wait until all the Loans for all the LoanBrokers are fetched. The Product Managers for this feature strongly believe that each Vault will only have a handful of LoanBrokers only.

If we choose to use serial API calls for this component, it will delay the loading of the page for all the Vaults. However, the issue of throttling affects only Vaults which have > 30 loan-brokers. (I don't know the exact number of loan brokers that will invoke throttling behavior)

src/rippled/lib/rippled.ts
ckeshava: fixed in bc237597

src/containers/Vault/index.tsx
ckeshava: thanks. fixed in bc237597

src/containers/Vault/index.tsx
ckeshava: this was dead code. I have removed it. thanks. bc237597

src/containers/Vault/VaultHeader/index.tsx
ckeshava: thanks. fixed in bc237597

src/containers/Vault/VaultHeader/Details.tsx
ckeshava: thanks. fixed in bc237597

src/containers/Vault/VaultLoans/BrokerDetails.tsx
ckeshava: thanks. fixed in bc237597

src/containers/Vault/VaultLoans/index.tsx
ckeshava: thanks. updated in 22896a62

src/containers/Vault/VaultDepositors/index.tsx
ckeshava: I have not been able to test this component with a Clio instance. Here is the open Issue in Clio repo: https://github.com/XRPLF/clio/issues/2695. At present time, this section displays the message: "Unable to fetch depositors information from Clio node at this time."

To answer your question: One invocation of this API returns `5` mpt_holders, because `pageSize == 5`. When the user clicks on the `next` button, this API is triggered again with the marker corresponding to that specific page. Refer to the `markers` react-state variable. The benefit of this is faster page-load time instead of waiting to fetch all the potential depositors of a vault.

src/containers/Vault/utils.ts
ckeshava: thanks, fixed in dd04c151

src/containers/Vault/index.tsx
ckeshava: updated in a2958bf16806a30361c406d64ab953cbcbd8d1a2

src/containers/shared/hooks/useTokenToUSDRate.ts
ckeshava: ok, updated in cc703691

src/containers/Vault/VaultDepositors/index.tsx
kuan121: >Is it okay if I keep this suggestion when I actually test this component later?

Do we plan to merge this PR with the code for the depositor section? If yes, I think even if the API isn’t ready, we can prepare the frontend and replace VaultDepositors with HoldersTable.

src/containers/Vault/VaultDepositors/index.tsx
kuan121: > The benefit of this is faster page-load time instead of waiting to fetch all the potential depositors of a vault.

We need to decide between immediate page loads and data-rich views (e.g., ranked top users). We assumed a full fetch of depositors would be too slow, but a ~5s response time is acceptable if paired with a loading state. We don't need to achieve for instant loading if a manageable delay allows us to show the useful information.

src/containers/shared/hooks/useTokenToUSDRate.ts
```diff
@@ -0,0 +1,115 @@
...
+    },
+  )
+
+  // XRP: use direct XRP to USD rate
+  if (currency === 'XRP') {
+    return { rate: xrpToUsdRate, isAvailable: true, isLoading: false }
+  }
+
+  // RLUSD: stablecoin pegged 1:1 with USD
+  if (currency === 'RLUSD') {
```
pdp2121, line 84: Would the currency for non-standard code returned a hex or parsed in LOS? If hex, does it need to be converted first? Also, just wonder why we need a special case for RLUSD

src/containers/Vault/VaultDepositors/DepositorTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useTokenToUSDRate } from '../../shared/hooks/useTokenToUSDRate'
+import { localizeNumber } from '../../shared/utils'
+import { formatAmount } from '../utils'
+
+interface Holder {
+  account: string
+  mpt_amount: string
+}
+
+interface AssetInfo {
```
pdp2121: I think we have an interface for Currency already

src/containers/Vault/VaultLoans/index.tsx
kuan121: >Even if we are throttled by the Clio node, this API call will retry and wait until all the Loans for all the LoanBrokers are fetched. 

Where is the retry logic defined?

> If we choose to use serial API calls for this component, it will delay the loading of the page for all the Vaults.

I’m not entirely sure what you meant by “all the Vaults,” but I agree that using concurrent queries (as is the case here) is preferable to sequential ones, as long as we can ensure successful retries in case throttling occurs.

src/containers/Vault/VaultDepositors/DepositorTable.tsx
ckeshava: I could not find a `interface Currency` in the code base. Do you remember any other constituent fields / keywords that I can search for?

src/containers/Vault/index.tsx
kuan121: 1. For MPT, another option is to use the ticker field in its metadata, if it exists (see: [MPT metadata schema](https://xrpl.org/docs/concepts/tokens/fungible-tokens/multi-purpose-tokens#metadata-schema)). You can leverage mptUtils.ts to parse the MPT metadata and extract the relevant fields.

2. > How do you propose that we translate that 20-byte string into a human-readable form?

Let me explain with a concrete example. The IOU token `RLUSD` has a currency code of `524C555344000000000000000000000000000000`. On Explorer, we don’t display the raw 524C555344000000000000000000000000000000; instead, it’s converted to RLUSD using logic like [this](https://github.com/ripple/explorer/blob/staging/src/containers/shared/components/Currency.tsx#L49). I just want to make sure we apply the same transformation here.

src/containers/shared/hooks/useTokenToUSDRate.ts
ckeshava: > Also, just wonder why we need a special case for RLUSD

I had written a special case for RLUSD because it is a stablecoin pegged to a 1:1 rate with the USD. I did not want to make an extra network call to fetch the USD price of a stablecoin.

> Would the currency for non-standard code returned a hex or parsed in LOS? 
The currency code is fetched from the Vault ledger-object's Asset field. As per my understanding, it will be a 20-byte hex string.

For display purposes, are you aware of any mapping that converts the 20-byte hex strings into a human-friendly currency display format? If not, I'll need to fallback to displaying the hex value.

src/containers/Vault/VaultLoans/index.tsx
ckeshava: As per the documentation of useQuery and useQueries methods: (both of them use the same set of options) the default configuration allows for three retries. https://tanstack.com/query/v5/docs/framework/react/reference/useQueries
image

Do you suggest that I specify the retry count explicitly?

src/containers/Vault/VaultDepositors/index.tsx
ckeshava: Since the `mpt_holders` API is not yet ready, I don't know how much time it will take to load all the depositors of the vault. Once that Clio API is available, we will be able to measure the latency of these API calls.

Should I change the design to fetch all the vault_depositors now? Or should I wait until the Clio API is fully implemented?

src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+                label={t('total_value_locked')}
+                value={(() => {
+                  const convertedAmount = convertToDisplayCurrency(assetsTotal)
+                  if (
+                    convertedAmount === undefined &&
+                    displayCurrency === 'usd'
+                  ) {
+                    return '--'
+                  }
+                  return formatAmount(
```
pdp2121: Please use the functions in `NumberFormattingUtils.ts` here and in other places for consistency. It's a new formatting standard that we just created. Probably will need to document it somewhere for future reference

src/containers/Vault/VaultLoans/BrokerDetails.tsx
```diff
@@ -0,0 +1,148 @@
...
+  Data?: string
+  ManagementFeeRate?: number
+  CoverAvailable?: string
+  CoverRateMinimum?: number
+  CoverRateLiquidation?: number
+  DebtTotal?: string
+  DebtMaximum?: string
+}
+
+interface AssetInfo {
```
pdp2121, line 27: Same here. Can we use existing interface?

src/containers/Vault/VaultLoans/BrokerDetails.tsx
```diff
@@ -0,0 +1,148 @@
+import { useTranslation } from 'react-i18next'
+import { CopyableText } from '../../shared/components/CopyableText/CopyableText'
+import { useLanguage } from '../../shared/hooks'
+import { useTokenToUSDRate } from '../../shared/hooks/useTokenToUSDRate'
+import { formatRate, LSF_LOAN_DEFAULT } from './utils'
+import { formatAmount } from '../utils'
+import { BrokerLoansTable } from './BrokerLoansTable'
+import WarningIcon from '../../shared/images/warning.svg'
+
+interface LoanBrokerData {
```
pdp2121, line 15: Is this already in xrpl.js?

src/containers/Vault/VaultLoans/BrokerLoansTable.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const paginatedLoans = filteredLoans.slice(startIndex, endIndex)
+
+  const handlePageChange = (page: number) => {
+    if (page >= 1 && page <= totalPages) {
+      setCurrentPage(page)
+    }
+  }
+
+  // Generate page numbers to display
+  const getPageNumbers = (): (number | string)[] => {
```
pdp2121: We already have Pagination to be reused in src/containers/shared/components/Pagination/Pagination.tsx

src/containers/Vault/CurrencyToggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface Props {
+  nativeCurrency: string // e.g., "XRP" or "RLUSD"
+  selected: string // The currently selected currency (nativeCurrency or "USD")
+  onToggle: (currency: string) => void
+  usdDisabled?: boolean // Disable USD option when no price is available
+  usdLoading?: boolean // Show loading state while fetching price
+}
+
+export const CurrencyToggle = ({
```
pdp2121, line 17: Can this be a reusable component?

src/containers/Vault/utils.ts
```diff
@@ -0,0 +1,122 @@
+import { formatLargeNumber } from '../shared/utils'
+
+/**
+ * Decode the Data field from hex to UTF-8 if needed
```
pdp2121: There's a reusable `convertHexToString` already

src/containers/Vault/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  data: string | undefined,
+): string | undefined => {
+  const parsed = parseVaultDataJson(data)
+  if (parsed && typeof parsed.w === 'string' && parsed.w.trim()) {
+    return parsed.w.trim()
+  }
+  return undefined
+}
+
+/**
```
pdp2121: You can use `parseAmount` function in NumberFormattingUtils.ts which would do the same thing

src/containers/Vault/VaultDepositors/index.tsx
ckeshava: ok. I have refactored the code to use `HoldersTable` component in this commit: 4f556df6

src/containers/Vault/VaultLoans/index.tsx
kuan121: Not really — the default three retries should be sufficient. By the way, does it retry immediately, or does it use some form of exponential backoff?

src/containers/Vault/index.tsx
ckeshava: thanks. I have updated the code to use these two suggestions in ff6f1931

src/containers/Vault/VaultLoans/index.tsx
ckeshava: image

By default, it uses the Exponential Backoff mechanism. This page has more information about the defaults: https://tanstack.com/query/v3/docs/framework/react/guides/important-defaults

Most of the Explorer codebase uses the default configuration for the `useQuery` method.

src/containers/shared/components/TransactionTable/TransactionTable.tsx
ckeshava: The Amount column has been removed in 556e2ba0

src/containers/Vault/VaultDepositors/index.tsx
ckeshava: Update: This suggestion has been incorporated in 556e2ba0. I revamped the design to fetch all the VaultDepositors because that is required for the `HoldersTable` react component.

src/containers/shared/hooks/useTokenToUSDRate.ts
ckeshava: As per this conversation, I am transforming all the hex strings into human-readable currency codes: https://github.com/ripple/explorer/pull/1281#discussion_r2766292458

src/containers/Vault/VaultHeader/index.tsx
ckeshava: ok, I have used the NumberFormatting utility methods here: 288e3892

src/containers/Vault/utils.ts
ckeshava: ok. I have removed this new method in this commit: 288e3892

src/containers/Vault/utils.ts
ckeshava: ok. updated this in 7956d5ac

src/containers/Vault/VaultLoans/BrokerLoansTable.tsx
ckeshava: ok. I have refactored this in 43670726

src/containers/Vault/CurrencyToggle.tsx
ckeshava: I'm not aware of other projects that can make use of this component. How can I make this more generalized? Are you suggesting different input properties for the `CurrencyToggle` component? For this project, `USD` and the native-currency of the Vault are the two relevant assets.

src/containers/Vault/VaultLoans/BrokerDetails.tsx
ckeshava: The `LoanBroker` ledger object model does not have the latest updates to the Lending Protocol. For instance, I could not find `ManagementFeeRate` field in the xrpl.js model file: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/ledger/LoanBroker.ts

cc @Patel-Raj11  we will need to update the models before the next release.

src/containers/Vault/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+      if (typeof ticker === 'string') {
+        return ticker
+      }
+      return `MPT (${asset.mpt_issuance_id.substring(0, 6)}...)`
+    }
+    return ''
+  }
+
+  return (
+    <Page vaultId={vaultId}>
```
kuan121, line 150: The left and right margins/paddings of this page are much smaller than those of other Explorer object pages ([MPT Object](https://livenet.xrpl.org/mpt/05EECEBE97A7D635DE2393068691A015FED5A89AD203F5AA), [IOU](https://livenet.xrpl.org/token/524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De) and [Account](https://livenet.xrpl.org/accounts/rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De)), please check what's missing.

public/locales/en-US/translations.json
```diff
@@ -802,5 +805,50 @@
...
+  "loan_status_paid_off": "Paid Off",
+  "next_due_date": "Next Due Date",
+  "origination_date": "Origination Date",
+  "frequency": "Frequency",
+  "installments": "Installments",
+  "prepayment_fee": "Prepayment Fee",
+  "no_loans_message": "No loans found for this broker.",
+  "loan_default_detected": "Loan default detected in this broker.",
+  "no_loan_brokers_message": "No loan brokers have been set up for this vault.",
+  "depositors_fetch_error": "Unable to fetch depositors information from Clio node at this time.",
```
kuan121: As discussed, we don’t need to show this level of technical detail to users, since they don’t care where the data comes from. I suggest we hide the Depositors section, as it provides no real value and takes up space.

src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+      <div className="vault-details-container">
+        <div className="details-column">
+          <table className="token-table">
+            <tbody>
+              <TokenTableRow
+                label={t('vault_id')}
+                value={
+                  <CopyableText
+                    text={vaultId}
+                    displayText={`${vaultId.substring(0, 8)}...${vaultId.substring(vaultId.length - 6)}`}
```
kuan121: Nit: You can create a `shortenVaultId()` in containers/shared/utils so that we can somehow ensure the same short format for Vault ID across Explorer pages.

src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+                    showCopyIcon
+                  />
+                }
+              />
+              {owner && (
+                <TokenTableRow
+                  label={t('owner')}
+                  value={<Account account={owner} displayText={`${owner.slice(0, 8)}...${owner.slice(-4)}`}/>}
```
kuan121: You can use `shortenAccount`

src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+                      }
+                      target="_blank"
+                      rel="noopener noreferrer"
+                    >
+                      {vaultWebsite}
+                    </a>
+                  }
+                />
+              )}
+              <TokenTableRow label={t('data')} value={decodedData || '-'} />
```
kuan121: Is `decodedData` a JSON object? Do we plan to show a JSON string as the value of the DATA field?

src/containers/Vault/VaultLoans/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  // Build maps of broker ID to loans and loan counts
+  const brokerLoansMap: Record<string, any[]> = {}
+  const loanCountMap: Record<string, number> = {}
+  brokerLoansQueries.forEach((query) => {
+    if (query.data) {
+      brokerLoansMap[query.data.brokerId] = query.data.loans
+      loanCountMap[query.data.brokerId] = query.data.loans.length
+    }
+  })
```
kuan121, line 138: nit: As discussed, we could sort the loan brokers by the number of loans, total debt across all their loans so that the order is deterministic and we show the most "interesting"/"relevant" loan broker first.

src/containers/Vault/index.tsx
ckeshava: ok. updated in 8bed0c19

public/locales/en-US/translations.json
ckeshava: Yes. I am not displaying this `VaultDepositors` component in the webpage. This message will not be displayed to the users under any circumstance.

ckeshava: I have removed the display of VaultDepositors component from the Vault/index.tsx page, so this message will never be displayed. 

Nevertheless, I have updated the message to be less verbose. 0bfab577

src/containers/Vault/VaultHeader/index.tsx
ckeshava: [aea08dd](https://github.com/ripple/explorer/pull/1281/commits/aea08dddf89c6d81c42ee401e8e1310d13f50212)

src/containers/Vault/VaultHeader/index.tsx
ckeshava: [aea08dd](https://github.com/ripple/explorer/pull/1281/commits/aea08dddf89c6d81c42ee401e8e1310d13f50212)

src/containers/Vault/VaultLoans/index.tsx
ckeshava: [bd3c57f](https://github.com/ripple/explorer/pull/1281/commits/bd3c57f1c36e2eae3e273ff1293f71e6d1fa6720)

src/containers/Vault/VaultHeader/index.tsx
ckeshava: Yes, `decodedData` is decoded from the hexadecimal encoded data obtained from XRPL on-chain.
1. If the Vault-metadata adheres to the proposed standards, yes, it will be a JSON string with 'n' and 'w' keys.
2. Otherwise, it could be any data that has been provided by the Vault-creator.

src/containers/Vault/VaultLoans/BrokerDetails.tsx
ckeshava: Are you referring to the `LoanBrokersData interface`? xrpl.js needs to be updated. It does not have the latest fields of the Lending Protocol, like `ManagementFeeRate` in its models.

pdp2121:
Other than the above, the UI looks good to me

src/containers/shared/utils.js
```diff
@@ -584,3 +595,11 @@ export const convertToHttpURL = (url) => {
   // Otherwise, assume it's a plain domain and add https://
   return `https://${url}`
 }
+
+/**
+ * Truncates the vaultID to ensure better readability
+ * @param {string} vaultID - The complete VaultID obtained from XRPL on-chain data
+ * @returns {string} The truncated VaultID
+ */
+export const shortenVaultID = (vaultID) =>
```
pdp2121, line 609: I think we already have a shorten function somewhere right? cc: @kuan121

src/containers/shared/utils.js
```diff
@@ -558,6 +561,14 @@ export const shortenMPTID = (
 export const shortenTxHash = (txHash = '') =>
   txHash.length > 12 ? `${txHash.slice(0, 6)}...${txHash.slice(-6)}` : txHash
 
+export const shortenLoanBrokerID = (loanBrokerID = '') =>
```
pdp2121, line 569: ditto

src/containers/Vault/VaultLoans/BrokerDetails.tsx
pdp2121: Can you add a TODO to remind us to update once it's available on js?

src/containers/Vault/CurrencyToggle.tsx
pdp2121: We can create a Toggle component that accept values an callbacks. In Explorer development, we tries to make components reusable as much as we can to reduce the amount of code loaded to browser

src/containers/shared/utils.js
kuan121: @pdp2121 

We currently have several ID-shortening utilities in this file.

Personally, I’d prefer defining a dedicated function for each ID type—for example, shortenAccountId, shortenMPTID, shortenNFTTokenID, etc.—since we may want to apply different abbreviation rules (such as varying prefix and suffix lengths) depending on the ID type.

The goal is to ensure that whenever we shorten a specific type of ID in Explorer, we consistently use the same function. That way, we avoid ending up with inconsistent formats (e.g., different versions of a VaultID) across different Explorer pages.

pdp2121: Makes sense!

ckeshava:
> The icon looks off and the tooltip is above the icon too much. Please use the hover icon in `src/containers/shared/images/hover.svg` you can also search for keyword renderTextTooltip for some example. It should look like this: Also, this toggle is off screen on mobile view. I would say let's make the toggle below VaultID in smaller screens

I have incorporated this suggestion. The `VaultHeader` now looks like this on mobile devices with 375 pixel width: 
image


This is the new location of the CurrencyToggle tooltip: 
image

@pdp2121  thanks for your suggestion, I re-used the HoverIcon. I was scratching my head to understand why my work was different in the hover icon style :)

public/locales/en-US/translations.json
```diff
@@ -17,7 +17,7 @@
   "explorer": "Explorer",
   "xrpl_org": "XRPL.org",
   "github": "GitHub",
-  "header.search.placeholder": "Search by Token, Address, Ledger or Txn",
+  "header.search.placeholder": "Search by Token, Address, Ledger, Txn or VaultID",
```
pdp2121, line 20: Could you add a note to other translations to update this as well? or create a new issue

ckeshava: https://github.com/ripple/explorer/issues/1291 👍

PR #1292, opened 2026-03-11
Feature: Vault Rankings Page

achowdhry-ripple:

---

Summary

  - Vault Rankings Page (/vaults): New listing page with server-side pagination,
  sorting (TVL, outstanding loans, utilization ratio, interest rate), asset type
  filtering (All/XRP/Stablecoins), and text search across vault names, accounts,
  assets, and websites
  - Vault Detail Page (/vault/:id): Individual vault view with header metadata
  (owner, asset, TVL, max supply, withdrawal policy, credentials, website),
  transaction history, loan broker tabs with per-broker loan tables, and
  depositor/shareholder breakdown with share value calculations
  - Aggregate Metrics: Header stats showing Total Value Locked, Outstanding Loans,
  Loans Originated, Avg. Interest Rate, # of Vaults, and Utilization Ratio — all
  converted from XRP to USD using the live oracle rate
  - Asset Price Conversion: Server-side caching of xrplmeta token prices for RWA
  assets (refreshed every 5 minutes), enabling accurate USD conversion for all vault
   assets — not just XRP-denominated ones
  - Vault Transaction Components: Simple, Description, and TableDetail views for
  VaultCreate, VaultDeposit, VaultWithdraw, VaultSet, VaultDelete, and VaultClawback
   transaction types
  - Test Coverage: Full test suites for the rankings page (container rendering,
  metrics, filters, table headers/rows, search, empty state, disclaimer), API layer
  (field mapping, sort/filter/search param handling, asset prices endpoint), vault
  detail page, vault header, loans, depositors, currency toggle, transactions, and
  utilities

  Test plan

  - Verify /vaults loads with correct aggregate metrics in the header
  - Confirm TVL and outstanding loan values display in USD for both XRP and non-XRP
  vaults
  - Test sorting by each column (TVL, outstanding loans, utilization, interest rate)
   in both directions
  - Test asset type filters (All Assets, XRP, Stablecoins) update the table
  - Test search filters results by vault name/account/asset/website
  - Click through to a vault detail page (/vault/:id) and verify header,
  transactions, loans, and depositors tabs
  - Verify navigation link appears in sidebar on mainnet only
  - Run npx jest src/containers/Vaults/test/ src/containers/Vault/test/ — all 88
  tests pass

Screenshot 2026-03-20 at 3 26 34 PM
Screenshot 2026-03-20 at 3 26 29 PM

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization


server/routes/v1/vaults.js
```diff
@@ -0,0 +1,128 @@
+const axios = require('axios')
+const log = require('../../lib/logger')({ name: 'vaults' })
+
+const XRPLMETA_BASE = 'https://s1.xrplmeta.org'
```
pdp2121: Can this be an `env` variable?

server/routes/v1/vaults.js
```diff
@@ -0,0 +1,128 @@
...
+    // Filter to RWA asset class (stablecoins used by vaults)
+    const rwaTokens = tokens.filter((t) => t.asset_class === 'rwa')
+    log.info(`Found ${rwaTokens.length} RWA tokens for vault price lookup`)
+
+    const prices = {}
+
+    await Promise.all(
+      rwaTokens.map(async (token) => {
+        try {
+          const metaUrl = `${XRPLMETA_BASE}/token/${token.currency}:${token.issuer_account}`
```
pdp2121: Why are we fetching from xrplmeta instead of LOS? Is this because it's more up-to-date?

src/containers/Vaults/api.ts
```diff
@@ -0,0 +1,100 @@
+import axios from 'axios'
+import type { VaultData } from './types'
+
+export interface VaultsMetrics {
```
pdp2121: Can these interfaces moved to types.ts file?

src/containers/Vaults/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+      .catch((error) => {
+        Log.error(error)
+        trackException(`vaults stats fetch --- ${JSON.stringify(error)}`)
+        metricsLoadedRef.current = true
+        markInitialLoaded()
+      })
+  }, [trackException])
+
+  // Fetch and periodically refresh asset prices from xrplmeta
+  useEffect(() => {
```
pdp2121: Should use `useQuery` here instead. Similar to other pages

src/containers/Vaults/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+  }, [trackScreenLoaded])
+
+  const markInitialLoaded = () => {
+    if (metricsLoadedRef.current && tableLoadedRef.current) {
+      setInitialLoaded(true)
+    }
+  }
+
+  // Fetch aggregate stats once on mount
+  useEffect(() => {
```
pdp2121: Same here. Should use `useQuery`

src/containers/Vaults/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+    return () => {
+      if (searchTimerRef.current) {
+        clearTimeout(searchTimerRef.current)
+      }
+    }
+  }, [searchQuery])
+
+  // Fetch vaults list whenever params change
+  const fetchVaults = useCallback(() => {
```
pdp2121: This can also be handled with useQuery

src/containers/Vaults/VaultsTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+    if (sortField === field) {
+      setSortOrder(sortOrder === 'asc' ? 'desc' : 'asc')
+    } else {
+      setSortField(field)
+      setSortOrder('desc')
+    }
+    setPage(1)
+  }
+
+  const renderSortHeader = (field: string, label: string) => {
```
pdp2121: I think we already have the reusable `SortTableColumn` component

src/containers/Vaults/api.ts
achowdhry-ripple: done

server/routes/v1/vaults.js
achowdhry-ripple: no particular reason -- switched to los with a fallback to xrplmeta

server/routes/v1/vaults.js
achowdhry-ripple: done

src/containers/Vaults/index.tsx
achowdhry-ripple: done

src/containers/Vaults/index.tsx
achowdhry-ripple: done

src/containers/Vaults/index.tsx
achowdhry-ripple: what would be the advantage for this one, since this handles a user input and debouncing rather than the query itself

src/containers/Vaults/VaultsTable.tsx
achowdhry-ripple: swapped to reusable component

server/routes/v1/vaults.js
kuan121: My two cents: I don’t think introducing a fallback here is necessary, as it would add additional complexity to the codebase. If LOS becomes unavailable for any reason, the Vault Ranking page would likely not function as intended anyway.

More broadly, if we decide that a fallback mechanism is needed, it would probably make more sense to design and implement it consistently across major features rather than introducing it in an ad hoc manner in only a few places.

achowdhry-ripple: I don't feel strongly about the fallback, removed it

src/containers/Vaults/index.tsx
pdp2121: `useQuery` is better here because it automatically handles the complex dependency management. When any filter changes (page, sort, search), `useQuery` automatically refetches and caches the results. With `useCallback`, you're manually managing loading states, error handling, and re-fetching logic that `useQuery` does automatically. The debouncing happens before the query key changes, so useQuery still gets the benefit of caching different search/filter combinations.

achowdhry-ripple: Ah the lines offset from the other changes, I thought this referred to something else. made this change, so it uses useQuery, thanks!

server/routes/v1/index.js
```diff
@@ -12,6 +17,9 @@ if (process.env.VITE_ENVIRONMENT !== 'custom') {
   api.use('/metrics', getCurrentMetrics)
   api.use('/tokens/search/:query', getTokensSearch)
   api.use('/tokens', getAllTokens)
+  api.get('/vaults/aggregate-statistics', getVaultsAggregateStats)
+  api.get('/vaults/asset-prices', getVaultAssetPrices)
+  api.get('/vaults', getVaults)
```
kuan121, line 22: I was wondering why do we need this additional layer of indirection? Can `src/containers/Vaults/index.tsx` call `getVaultsAggregateStats` directly, which in turn fetches data from LOS?

Is there a convention or architectural guideline that frontend components should go through the Explorer server, which then interacts with LOS, XRPLMeta, or other external services?

src/containers/shared/components/SortColumn.tsx
```diff
@@ -16,6 +16,7 @@ interface SortTableProps {
   setSortOrder: (order: SortOrder) => void
```
kuan121, line 16: This is a great example of reusable components. Thanks @pdp2121 !

src/containers/App/navigation.ts
```diff
@@ -31,6 +35,12 @@ export const navigationConfig: NavigationMenuAnyRoute[] = [
     title: 'tokens',
     current: (path: string) => isTokens(path),
   },
+  FEATURE_VAULTS_PAGE &&
+    process.env.VITE_ENVIRONMENT === 'mainnet' && {
```
kuan121, line 39: Do you mean to set `FEATURE_VAULTS_PAGE` to `false`? If it's `true`, we will show the Vault Ranking page for mainnet after we deploy this change to prod.

achowdhry-ripple: It is set to that during development right now so I can see it. I'll change it to false once I make any code changes based on your review, so that I can confirm it works before feature flagging it

server/routes/v1/index.js
achowdhry-ripple: I think CORS errors is one main reason, in addition to separation of concerns, caching when necessary, and more organized error handling. This is the pattern we have for tokens and dex trades in explorer as well, @pdp2121 feel free to add other reasons if I missed any

src/containers/App/navigation.ts
achowdhry-ripple: toggled it, should be false now. will keep true on local to test

server/routes/v1/index.js
kuan121: > I think CORS errors is one main reason, 

For the Account Page, we make all the calls to LOS from the frontend component without issues.

> in addition to separation of concerns, caching when necessary, and more organized error handling. 

We can do all these things in the `api.ts` file that reside in the frontend component folder.

kuan121: One potential drawback of this approach is that it routes all browser requests (from multiple users/IP addresses) through the Explorer server, effectively consolidating them into API calls from a single IP address, which may be subject to rate limiting enforced by Clio, LOS, XRPLMeta or any external services.

src/containers/App/navigation.ts
achowdhry-ripple: Thank you for raising this -- The fundamental truth of vault information comes from Clio/rippled ledger data, and the data this UI sources comes from LOS (https://los.prod.ripplex.io/). the data and any discrepancies will be auditable from there. I hope this clarifies your question!

server/routes/v1/index.js
pdp2121: If the results are cached then we should use Explorer backend. If everytime the request sent to the backend fires a call to external service, then it should happen on the frontend instead

kuan121: @pdp2121 Thanks for the clear explanation. This makes sense.

kuan121: @achowdhry-ripple Resolving this comment

kuan121 (approved):

PR #1316, opened 2026-04-17
Extract info card style into shared file

kuan121:

Extract the shared card styles used by the AMM Pool page (Basic Info, Market Data, Auction) and the AMM Rankings page (General Info) into a single [`src/containers/shared/css/info-card.scss`](src/containers/shared/css/info-card.scss).

Changes:
- New shared `info-card.scss` with documented usage and CSS-based row dividers (replaces explicit `<div className="info-card-separator" />` elements)
- `AMMPool/styles.scss` and `AMMRankings/ammRankings.scss` import the shared styles instead of defining their own card rules
- `BasicInfoCard`, `MarketDataCard`, `AuctionCard` wrapper class renamed `amm-pool-info-card` → `info-card`; separator `<div>`s removed
- `GeneralInfoCard` refactored to use the shared `info-card` class structure (was using bespoke `general-info-card` / `stat-row` / `stat-label` etc.)
- Increased gap between the General Info card and the TVL/Volume chart on the AMM Rankings page from 16px → 32px

### Context of Change

The AMM Pool page and AMM Rankings page both display info cards with the same visual pattern — an icon title followed by label/value rows separated by horizontal dividers. They were implemented independently with different class names (`amm-pool-info-card` vs `general-info-card`) and different structures, leading to visual inconsistencies and duplicate CSS.

Consolidating to a single shared stylesheet:
- Guarantees visual consistency across both pages
- Removes ~95 lines of duplicated SCSS
- Lets future pages adopt the same pattern with a single `@use` import
- CSS-based row dividers eliminate the need for explicit separator elements in JSX (cleaner markup)

### Type of Change

### Codebase Modernization

## Before / After

**Before**: Two sets of info card styles — `amm-pool-info-card` in AMMPool/styles.scss and `general-info-card` in ammRankings.scss — with different spacing, structure, and class names.
Screenshot 2026-04-17 at 3 31 43 PM

**After**: Single shared `info-card.scss` used by both pages. All four cards (Basic Info, Market Data, Auction, General Info) now use identical class names (`info-card`, `info-card-title`, `info-card-icon`, `info-card-rows`, `info-card-row`, `info-card-label`, `info-card-value`). Visual appearance is unchanged for the AMM Pool page; the General Info card on the Rankings page now matches the AMM Pool cards exactly.
Screenshot 2026-04-17 at 3 31 32 PM

## Test Plan

1. All existing unit tests pass
2. Lint and style check pass
3. Manually verified visual appearance on both `/amm/:id` and `/amm` pages matches before the refactor

### DIFF
--- a/src/containers/AMMPool/InfoCards/AuctionCard.tsx
+++ b/src/containers/AMMPool/InfoCards/AuctionCard.tsx
@@ -123,7 +123,7 @@ export const AuctionCard: FC<AuctionCardProps> = ({
     replacementCostRaw != null ? getLPTokenUSD(replacementCostRaw) : null
 
   return (
-    <div className="amm-pool-info-card">
+    <div className="info-card">
       <h3 className="info-card-title">
         <AuctionIcon className="info-card-icon" />
         {t('auction')}
@@ -142,7 +142,6 @@ export const AuctionCard: FC<AuctionCardProps> = ({
             )}
           </span>
         </div>
-        <div className="info-card-separator" />
         <div className="info-card-row">
           <span className="info-card-label">{t('expiration')}</span>
           <span className="info-card-value">
@@ -155,14 +154,12 @@ export const AuctionCard: FC<AuctionCardProps> = ({
               : '--'}
           </span>
         </div>
-        <div className="info-card-separator" />
         <div className="info-card-row">
           <span className="info-card-label">{t('discounted_fee')}</span>
           <span className="info-card-value info-card-value-orange">
             {discountedFee}
           </span>
         </div>
-        <div className="info-card-separator" />
         <div className="info-card-row">
           <span className="info-card-label">{t('price_paid')}</span>
           <span className="info-card-value">
@@ -182,7 +179,6 @@ export const AuctionCard: FC<AuctionCardProps> = ({
             )}
           </span>
         </div>
-        <div className="info-card-separator" />
         <div className="info-card-row">
           <span className="info-card-label">{t('replacement_cost')}</span>
           <span className="info-card-value">
--- a/src/containers/AMMPool/styles.scss
+++ b/src/containers/AMMPool/styles.scss
@@ -1,6 +1,7 @@
 @use '../shared/css/variables' as *;
 @use '../shared/css/table';
 @use '../shared/css/data-tables-notice';
+@use '../shared/css/info-card';
 
 .amm-pool-page {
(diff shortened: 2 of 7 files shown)


src/containers/shared/css/info-card.scss
```diff
@@ -0,0 +1,131 @@
...
+ *   info-card-value-orange  — warning/attention state
+ *   info-card-value-link    — clickable link appearance
+ *
+ * Use info-card-subtitle for secondary text within a value cell.
+ */
+
+.info-card {
+  padding: 24px;
+  border: 1px solid $black-60;
+  border-radius: 8px;
```
cybele-ripple, line 38: Does this differ from the existing `$border-radius`?

kuan121: Good question. 
1. ammRankings.scss uses `border-radius: $border-radius;` where `$border-radius` is 8px.
2. amm pool styles.scss uses `border-radius: 8px;`

The border-radius before and after are the same.

cybele-ripple (approved):

src/containers/AMMPool/InfoCards/AuctionCard.tsx
```diff
@@ -142,7 +142,6 @@ export const AuctionCard: FC<AuctionCardProps> = ({
             )}
           </span>
         </div>
-        <div className="info-card-separator" />
```
kuan121, line 145: I remove these lines since we no longer explicitly need a div element to create the divider line. That will be achieved using CSS.

src/containers/AMMRankings/ammRankings.scss
```diff
@@ -28,7 +29,7 @@
     flex-direction: column-reverse;
     align-items: stretch;
     margin-bottom: 64px;
-    gap: 16px;
+    gap: 32px;
```
kuan121, line 32: The current gap between the TVL/Volume chart and the General Info card is too smaller, I increase it to a reasonable value.

src/containers/AMMPool/styles.scss
```diff
@@ -139,102 +140,6 @@
   }
 }
 
-.amm-pool-info-card {
```
kuan121, line 142: The follow block is removed and the similar CSS is put in info-card.scss to avoid duplicate CSS

pdp2121 (approved):

PR #1323, opened 2026-05-13
Add maintenanece window banner

kuan121:

Adds a reusable `MaintenanceBanner` component that announces scheduled downtime to Explorer users.

- Mounted in `App.tsx` so the banner appears on **every page** (home, ledgers, transactions, accounts, validators, etc.) across all environments (dev / staging / prod).
- Window is configured by two ISO 8601 UTC timestamps at the top of `MaintenanceBanner.tsx` — `MAINTENANCE_START` and `MAINTENANCE_END`.
- Live countdown ticks every second until the window opens; once inside the window the countdown text disappears (the rest of the site is unavailable anyway); once `now >= MAINTENANCE_END` the banner unmounts entirely with no further re-renders.

Current scheduled window: **Friday, 15 May 2026, 01:00–01:30 UTC** (~30 min).

### Context of Change

Per a request from Platform team. Production maintenance will take Explorer offline for up to 30 minutes. Following best practices for operator-facing tools, we want to surface this to users via an in-product banner ahead of time.

### Type of Change

- [ ] Documentation Updates
- [x] Translation Updates
- [ ] Release

### Codebase Modernization

- [] Updated files to React Hooks (new component is a function component using `useState` / `useEffect`)
- [] Updated files to TypeScript (`.tsx` / `.scss` only)

## Before / After

**Before:** No mechanism to communicate scheduled downtime to users — they hit the outage with no warning.

**After:** A persistent banner appears at the top of every Explorer page during the configured window:
Desktop View
Desktop

Mobile View
Mobile

## Test Plan

**Automated** (`src/containers/shared/components/MaintenanceBanner/test/MaintenanceBanner.test.tsx`):

Run `npm test -- MaintenanceBanner` — 6 cases, all passing:

- [x] Renders banner with countdown when current time is before the window
- [x] Countdown updates every second
- [x] Hides countdown line once the window has started
- [x] Does not render once the window has ended
- [x] Does not render when `start`/`end` is `null`
- [x] Does not render when `start`/`end` is not a valid ISO timestamp

Tests are time-controlled via `jest.useFakeTimers()` + `setSystemTime()` and date-controlled via optional `start`/`end` props, so they're independent of real wall-clock and of the production maintenance window.

**Manual verification:**

- [x] `npm start`, set `MAINTENANCE_START` to a few minutes in the future locally → banner shows with live countdown
- [x] Wait until `now >= start` → countdown line disappears, banner stays
- [x] Wait until `now >= end` → banner unmounts on next tick
- [x] Set `MAINTENANCE_START = null` → banner does not render
- [x] Banner renders on `/`, `/transactions`, `/validators`, `/ledgers/<n>`, `/accounts/<addr>`
- [x] Banner renders correctly on mobile viewport widths (column layout, smaller margins)


pdp2121 (approved):

src/containers/shared/components/MaintenanceBanner/test/MaintenanceBanner.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { render, screen, cleanup } from '@testing-library/react'
+import { act } from 'react'
+import { I18nextProvider } from 'react-i18next'
+import i18n from '../../../../../i18n/testConfigEnglish'
+import { MaintenanceBanner } from '../MaintenanceBanner'
+
+const START = '2030-01-01T10:00:00Z'
+const END = '2030-01-01T10:30:00Z'
```
cybele-ripple, line 8: Should we consider using environmental variables so we don't need to do a PR to implement a new banner? In this case, we need to open a PR each time we display the banner

kuan121: Good question. Even with the environment variable approach, the current Explorer env variable setup still requires us to open a PR in [ripplex-dge-gitops](https://github.com/xpring-eng/ripplex-dge-gitops/blob/main/aws/prod/us-east-1/explorer/values.yaml) for each new maintenance window.

cybele-ripple (approved):

PR #1328, opened 2026-06-11
fix(vault): scale XRP drops & MPT amounts on Vault detail page

Patel-Raj11:

- Apply the same `parseAmount(convertScaledPrice(BigInt(raw), scale))` pattern used by the MPT page to all amount fields on the Vault detail page (header + loans), so raw ledger values are converted to display units before formatting.
- **Along the way**: fix a related display bug where `Max Total Supply`, `Available to Borrow`, and `Unrealized Loss` rows showed a `$X USD` prefix/suffix when the currency toggle was set to USD but never multiplied by the conversion rate — i.e., a 100 XRP vault rendered `$100 USD` instead of `~$200 USD`. Those three rows now route through `convertToDisplayCurrency` just like Total Value Locked does.

### Type of Change

- [x] Tests (updated fixtures that were locking in the buggy behavior, added MPT scaling + USD toggle coverage)

### Codebase Modernization

- [ ] Updated files to React Hooks (already on hooks)
- [ ] Updated files to TypeScript (already TS)

## Before / After

**XRP vault** (`devnet` example, `AssetsTotal` = `2500000000` drops):

Before

XRP
image

MPT
image

After
image

MPT
image

## Test Plan

Added/Fixed tests.

### DIFF
--- a/src/containers/Vault/VaultHeader/index.tsx
+++ b/src/containers/Vault/VaultHeader/index.tsx
@@ -15,6 +15,7 @@ import {
   shortenAccount,
   getCurrencySymbol,
   isCurrencyExoticSymbol,
+  convertScaledPrice,
 } from '../../shared/utils'
 import './styles.scss'
 import { useAnalytics } from '../../shared/analytics'
@@ -47,6 +48,7 @@ interface Props {
   data: VaultData
   vaultId: string
   displayCurrency: string
+  assetScale?: number
 }
 
 // Vault flags from XLS-65d spec
@@ -59,7 +61,12 @@ const WITHDRAWAL_POLICIES: { [key: number]: string } = {
   1: 'first_come_first_served',
 }
 
-export const VaultHeader = ({ data, vaultId, displayCurrency }: Props) => {
+export const VaultHeader = ({
+  data,
+  vaultId,
+  displayCurrency,
+  assetScale,
+}: Props) => {
   const { t } = useTranslation()
   const { trackException } = useAnalytics()
   const rippledSocket = useContext(SocketContext)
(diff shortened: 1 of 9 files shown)


src/containers/Vault/VaultHeader/index.tsx
```diff
@@ -353,7 +377,15 @@ export const VaultHeader = ({ data, vaultId, displayCurrency }: Props) => {
...
                 label={t('available_to_borrow')}
                 value={(() => {
-                  const parsedAmt = parseAmount(assetsAvailable ?? '0', 2)
+                  const convertedAmount =
+                    convertToDisplayCurrency(assetsAvailable)
+                  if (
+                    convertedAmount === undefined &&
+                    displayCurrency === 'USD'
+                  ) {
+                    return '--'
```
pdp2121: nit: I think there's a default empty value variable somewhere, instead of hardcoding repetitively

src/containers/Vault/VaultLoans/LoanRow.tsx
```diff
@@ -112,7 +114,14 @@ export const LoanRow = ({
   )
 
   const formatAmount = (amount: string | number): string => {
-    const num = typeof amount === 'string' ? Number(amount) : amount
+    const raw = String(amount)
```
pdp2121, line 117: This seems repetitive, can we use/create a shared function?

src/containers/Vault/VaultHeader/index.tsx
Patel-Raj11: I think it's not exported constant, each component has a local constant as shown below (for instance, AMM page, Amendments Table etc). I can create a constant here following the same convention, if we need an repo wide export and want to consume it, we can refactor entire codebase in another task.

`const DEFAULT_EMPTY_VALUE = '--'`

src/containers/Vault/VaultLoans/LoanRow.tsx
Patel-Raj11: Actually, I noticed this but `formatAmount`, `formatFee` seems to be repetitive entirely before these additions and there are some repetitions in other components introduced by Vault page as far as currency conversion and display goes. I think it's better to revisit these pages and tackle them as opposed to trying to clean the code in this PR. I can extract out the XRP/MPT switch part but that doesn't add any value until we clean the entire components.

pdp2121 (approved):

PR #1339, opened 2026-08-27
MPT Object page improvement

kuan121:

Fixes four reported display issues on the MPToken object page, and replaces the placeholder circulating-supply figure with a real calculation shared with the IOU page.

- **Supply / Circ Supply** — `Supply` is now the on-chain `OutstandingAmount`; `Circ Supply` subtracts holders owning ≥ 20% (skipped for RWA tokens). Fixes the reported "circulating supply > supply".
- **Issuer name** is no longer truncated with an address-style truncator.
- **Metadata URIs** are no longer clipped at 65 characters.
- **Website chips** (MPT + IOU issuer) show the registrable domain (`franklintempleton.com`) instead of a middle-truncated string that destroyed the TLD; all domain helpers are consolidated in `domainUtils`.
- **Shared logic + cleanup** — circulating-supply rules extracted to a shared, unit-tested util used by both token pages; two dead modules removed.

### Context of Change

Reported against [livenet.xrpl.org/mpt/064E27366D15D1F5614B4D4E3183F835BCD62E7F6FD9FFD8](https://livenet.xrpl.org/mpt/064E27366D15D1F5614B4D4E3183F835BCD62E7F6FD9FFD8) (Franklin `sgBENJI`). Each root cause was confirmed against the live `ledger_entry mpt_issuance` response rather than inferred.

**1. Circulating supply > supply.** This issuance sets no `MaximumAmount`, so `maxAmt` was `undefined` and `MarketData` rendered `BigInt(maxAmt || '0')` → **Supply = 0**, while Circ Supply showed the real ~28.14. Per spec an unset maximum means the cap is `2^63-1`, which is meaningless to display.

After product review the fix is: **Supply = `OutstandingAmount`**, and **Circ Supply = Supply minus holders ≥ 20%** — the same rule the IOU page already applied. Large holders are issuer/treasury/whale wallets that are not meaningfully in circulation. The exclusion is **skipped for RWA tokens** (`asset_class === 'rwa'`), whose large holders are custodians; stablecoins are an RWA subclass and are covered by the same check.

**2. Truncated issuer name.** `shortenAccount()` — a 7…5 _address_ truncator — was applied to the human-readable issuer name, rendering "Franklin Templeton Investments" as "Frankli…ments".

**3. Truncated metadata.** The shared `JsonView` hard-coded `collapseStringsAfterLength={65}`, clipping the 122-character URI in the metadata panel.

**4. Website chip destroyed the TLD.** `shortenDomain(url, 12, 7)` middle-truncates, producing `www.franklin…hnology`. The TLD is the part users rely on to judge whether a link is safe, so website chips now show the registrable domain parsed with a real public-suffix list (`tldts`) — correctly keeping `thing.co.uk` rather than collapsing it to `co.uk`. `tldts` was already in the tree as a transitive dependency and is promoted to a direct one. The same treatment is applied to the **IOU issuer chip**. The **NFT URL column keeps `shortenDomain`** — its per-NFT resource URIs need the path to stay visible, so domain-only display would collapse distinct rows. All three helpers (`getRegistrableDomain`, `shortenDomainFromLeft`, `shortenDomain`) now live together in `shared/domainUtils.ts`; non-HTTP schemes (`ipfs://`, `ar://`) fall back to the host/CID token.

**Shared logic.** The ≥20% and RWA rules now live in `Token/shared/utils/circulatingSupply.ts` and are used by both token pages. IOU's calculation was promoted out of its component into that module so it can be unit-tested directly. `isRwaAssetClass` accepts `unknown` and strict-compares against `'rwa'`, so untrusted, non-string issuer metadata compares unequal rather than throwing.

**Dead code.** `IOU/hooks/useMarketCalculations.ts` and `IOU/utils/tokenCalculations.ts` (plus their tests) were duplicates reachable only from their own tests — the shipping IOU page had its own copy.

### Type of Change

- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### Codebase Modernization

All touched and new files are function components / TypeScript. New modules (`shared/domainUtils.ts`, `Token/shared/utils/circulatingSupply.ts`, `Token/MPT/utils/circulatingSupply.ts`) are TypeScript.

## Before / After

**Market Data box**

| Field       | Before                      | After                                  |
| ----------- | --------------------------- | -------------------------------------- |
| Supply      | `0.00` (no `MaximumAmount`) | `28.14` (on-chain `OutstandingAmount`) |
| Circ Supply | `28.14` (> Supply)          | `≤ Supply`, excluding ≥20% holders     |

Circ Supply shows a spinner while holders load, and `--` if that request fails — it never silently shows the unadjusted supply as if it were a real circulating figure.

**Header**

| Element      | Before                                 | After                                      |
| ------------ | -------------------------------------- | ------------------------------------------ |
| Issuer name  | `Frankli…ments` (truncated, in title)  | Full name, clickable, in General Overview  |
| Website chip | `www.franklin…hnology` (TLD destroyed) | `franklintempleton.com`, full URL on hover |
| Metadata URI | clipped at 65 chars                    | rendered in full                           |

**Rule summary**

| Page | Skip ≥20% exclusion when… | `asset_class` source |
| ---- | ------------------------- | -------------------- |
| IOU  | `asset_class === 'rwa'`   | LOS                  |
| MPT  | `asset_class === 'rwa'`   | on-chain metadata    |

Stablecoins are an RWA subclass, so this single `asset_class === 'rwa'` check covers them on both pages.

## Test Plan

**Automated** — all green:

- `npm run build-ts` — no type errors
- `npm run lint:ci` — ESLint, Stylelint, Prettier (`--max-warnings 0`)
- `npm run test:ci` — **1713 tests pass**, coverage above the 70/80 thresholds
- `pre-commit run --all-files`
- `npm run build` — production bundle builds with the new dependency

**New/updated tests:**

- `Token/shared/test/utils/circulatingSupply.test.ts` — threshold boundary (exactly 20%), large-holder subtraction, `isRwaAssetClass` (strict lowercase match + non-string/`unknown` safety), and `calculateIouCirculatingSupply` (reported supply, supply fallback, RWA skip)
- `Token/MPT/test/utils/circulatingSupply.test.ts` — asset-scale handling, RWA skip, negative floor, and two exact-integer regressions: float residue → `0`, and amounts beyond double precision
- `shared/test/domainUtils.test.ts` — registrable domain (incl. `thing.co.uk`), non-URL fallback, left-truncation preserving the TLD, and `shortenDomain` middle-truncation
- `MarketData` / `Header` / `Metadata` / `GeneralOverview` / `DomainLink` — Supply vs Circ Supply, spinner and `--` states, untruncated issuer name and metadata, `title` tooltip

**Manual:** `npm start`, open `/mpt/064E27366D15D1F5614B4D4E3183F835BCD62E7F6FD9FFD8` (see screenshot below) — Supply is the full outstanding amount, Circ Supply is ≤ Supply (cross-check excluded accounts against Holders table rows with % ≥ 20), the chip reads `franklintempleton.com` with the full URL on hover, the issuer name renders in full, and the metadata URI is not clipped. IOU token pages show unchanged supply and market-cap values after the refactor.
MPT Improvement


src/containers/Token/IOU/hooks/useMarketCalculations.ts
```diff
@@ -1,48 +0,0 @@
-import { useMemo } from 'react'
```
kuan121, line 1: Removed as it's dead code that isn't used anywhere.

src/containers/Token/IOU/test/hooks/useMarketCalculations.test.ts
```diff
@@ -1,273 +0,0 @@
-import { renderHook } from '@testing-library/react'
```
kuan121, line 1: Removed as it's dead code that isn't used anywhere.

src/containers/Token/IOU/test/utils/tokenCalculations.test.ts
```diff
@@ -1,217 +0,0 @@
-import {
```
kuan121, line 1: Removed as it's dead code that isn't used anywhere.

src/containers/Token/IOU/utils/tokenCalculations.ts
```diff
@@ -1,38 +0,0 @@
-import { parseIntegerAmount } from '../../../shared/NumberFormattingUtils'
```
kuan121, line 1: Removed as it's dead code that isn't used anywhere.

src/containers/Token/MPT/index.tsx
```diff
@@ -100,6 +113,35 @@ export const MPT = () => {
...
 
+  // Circulating supply is only meaningful once holders resolve. While the fetch
+  // is pending (or hasn't started yet) show a spinner; if it failed, leave it
+  // undefined so the header renders "--" rather than the unadjusted supply.
+  const circulatingSupplyLoading =
+    holdersLoading || !(holdersData || holdersError)
+
+  // Circulating supply = outstanding amount minus large (>= 20%) holders, except
+  // for RWA tokens where those holders are custodians/treasuries (no exclusion).
+  const isRwa = isRwaAssetClass(
```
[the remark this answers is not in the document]
kuan121: Is this still an issue in the latest commit?

src/containers/Token/MPT/index.tsx
```diff
@@ -100,6 +113,35 @@ export const MPT = () => {
...
+    mptokenIssuance?.parsedMPTMetadata?.asset_class as string | undefined,
```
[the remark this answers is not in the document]
kuan121: Is this still an issue in the latest commit?

src/containers/Token/MPT/index.tsx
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/ripple/explorer/pull/1339/changes/ea9e59f4787ca9c7bc768d78ee28c0b9ae6eab8f

src/containers/Token/MPT/index.tsx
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/ripple/explorer/pull/1339/changes/ea9e59f4787ca9c7bc768d78ee28c0b9ae6eab8f

src/containers/Token/shared/utils/circulatingSupply.ts
```diff
@@ -0,0 +1,98 @@
...
+  tokenData: IouTokenSupplyData,
+  holdersData: IouHoldersData | undefined,
+): number => {
+  if (tokenData.circ_supply) {
+    return Number(tokenData.circ_supply)
+  }
+
+  const supply = Number(tokenData.supply) || holdersData?.totalSupply || 0
+
+  if (isRwaAssetClass(tokenData.asset_class)) {
```
cybele-ripple, line 93: This checks if asset_class is `rwa`, but [the function it replaces](https://github.com/ripple/explorer/pull/1339/changes#diff-dd7d86f10be7dbf328b961e0de43796e14a7ecca104d4c794061c47260ca5259L43) checked if `asset_subclass` was `stablecoin`. Is this intentional?

src/containers/shared/utils.js
```diff
@@ -560,15 +560,6 @@ export const shortenEncryptionKey = (key = '') =>
 
 export const stripHttpProtocol = (url = '') => url.replace(/^https?:\/\//, '')
```
cybele-ripple, line 561: Is `stripHttpProtocol` still needed? Looks like `getRegistrableDomain` now accomplishes what `stripHttpProtocol` did

src/containers/Token/shared/utils/circulatingSupply.ts
kuan121: Intentional, yes — and it's worth stating the behavior change explicitly, so thanks for flagging.

The product decision was to key the exclusion on **RWA** rather than on stablecoins specifically: for RWA tokens the large holders are custodians/treasuries whose balances are legitimately part of circulating supply, so they shouldn't be subtracted. Stablecoins are an RWA *subclass*, so `asset_class === 'rwa'` subsumes the old `asset_subclass === 'stablecoin'` check.

I checked that empirically rather than assuming it. Across the top 1000 tokens by holders on XRPL Meta, every token with `asset_subclass: 'stablecoin'` also has `asset_class: 'rwa'` (6 of them, e.g. USDC), and there are **zero** stablecoins with a non-`rwa` asset class — so no stablecoin loses its skip.

The new check is also deliberately broader: other RWA subclasses (e.g. `rwa`/`commodity`) now skip the exclusion too, which is the intent.

It additionally unifies IOU and MPT on a single rule — MPT reads `asset_class` from on-chain XLS-89 metadata and doesn't fetch `asset_subclass`, so a subclass-based check wasn't available on that page.

src/containers/shared/utils.js
kuan121: Good catch — you're right, it's dead now. Its last caller was the IOU issuer chip, which this PR moved to `getRegistrableDomain`; after that it was referenced only by its own unit test.

Removed in 1c687133, along with its test block.

`convertToHttpURL` stays, though — it's still used for the MPT logo URL and does a different job (rewriting `ipfs://` to a gateway URL rather than just stripping a scheme).

src/containers/Token/shared/utils/circulatingSupply.ts
kuan121: It's intentional. Previously, we only perform the check against stablecoin, which is a subset of RWA. Now, we have decided (confirmed with our PM) to extend the check to all RAW assets

src/containers/shared/utils.js
kuan121: Good catch! I've removed it from the latest commit https://github.com/ripple/explorer/pull/1339/changes/059582ab8ec90be82adf733dbbd429ebf32ee4cc

src/containers/Token/MPT/index.tsx
```diff
@@ -100,6 +113,33 @@ export const MPT = () => {
     },
   )
 
+  // Circulating supply is only meaningful once holders resolve. While the fetch
+  // is pending (or hasn't started yet) show a spinner; if it failed, leave it
+  // undefined so the header renders "--" rather than the unadjusted supply.
+  const circulatingSupplyLoading =
```
cybele-ripple: Is there a chance that `mptokenIssuance.issuer` could ever be missing? If it is ever missing, then the  holders query never runs so `holdersLoading`, `holdersData`, and `holdersError` are false, undefined, and false, respectively. This would mean that `circulatingSupplyLoading` is true and [the loading spinner in MarketData.tsx](https://github.com/ripple/explorer/blob/c5e976c80c9826fea3b88755dd4fbb8b83a0f672/src/containers/Token/MPT/Header/MarketData.tsx#L56) never resolves

src/containers/Token/MPT/index.tsx
```diff
@@ -100,6 +113,33 @@ export const MPT = () => {
...
+    holdersLoading || !(holdersData || holdersError)
+
+  // Circulating supply = outstanding amount minus large (>= 20%) holders, except
+  // for RWA tokens where those holders are custodians/treasuries (no exclusion).
+  const isRwa = isRwaAssetClass(mptokenIssuance?.parsedMPTMetadata?.asset_class)
+  const circulatingSupply = useMemo(
+    () =>
+      holdersData
```
cybele-ripple: Does it make sense to have `circulatingSupply` dependent on `holdersData` even in the case where `isRwa` is true? The RWA branch of `calculateMptCirculatingSupply` never uses holder data. The spinner will appear until the holder fetch resolves even when the holder data isn't needed

src/containers/Token/MPT/index.tsx
kuan121: Issuer is a required field on MPTokenIssuance, so any issuance we successfully fetch has it.

src/containers/Token/MPT/index.tsx
kuan121: Good point. I updated the code so that `circulatingSupply` is computed immediately when `isRwa` is true without relying on the holder data (see https://github.com/ripple/explorer/pull/1339/commits/5065260275f64ab2434d9ea624b001fc99c9c861)

src/containers/shared/components/JsonView/JsonView.tsx
```diff
@@ -46,7 +56,7 @@ export const JsonView = ({
       <ReactJson
         src={data}
         collapsed={isExpanded ? false : 5}
-        collapseStringsAfterLength={65}
+        collapseStringsAfterLength={collapseStringsAfterLength}
```
pdp2121, line 59: Should we set a default for this since it's an optional field?

pdp2121 (approved):

src/containers/shared/components/JsonView/JsonView.tsx
kuan121: It's 65 on L28

PR #1320, opened 2026-04-28
  fix: Reject non-Vault ledger entries in getVault

ckeshava:
## High Level Overview of Change
The existing implementation of several `get<LedgerObject>` methods in the Explorer do not validate the type of ledger-objects. This has led to cases where user specifies the hash of a `PermissonedDomain` object however they are redirected to a `Vault` page incorrectly.

This bug was discovered with the input `EF98FDBA404CBEB4F746DA1026B859E260BBB459D111268F6A26BBC7C4811A04` on the XRPL Devnet. This value points to a `PermissionedDomain`, however users are incorrectly directly to the `Vault` Page.

Note: other rippled-related methods have not been updated in this PR because they have not exhibited a strong need for a strong response type-check. This also enables a nimble bug-fix PR.                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                 
Solution proposed in this PR:                                                                                                                                                
                                                                                                                                                                                                 
  - Add a `LedgerEntryType === 'Vault'` guard in `getVault`. Any other                                                                                                                           
    ledger entry type now rejects with `Error('Not a Vault', 404)`.                                                                                                                              
  - Add `src/rippled/lib/test/rippled.test.ts` with unit-test coverage                                                                                                                           
    for `getVault`'s success path and every error branch.     
    
    
This PR is deployed in the dev environment of the Explorer: https://devnet.dev.ripplex.io/. Typing the above hash in the `SearchBar` should show a message akin to "No txn/nft/vault found for this hash"/                                                                                                                                   

<!--
Please include a summary/list of the changes.
If too broad, please consider splitting into multiple PRs.
-->



<!--
Please include the context of a change.
If a bug fix, when was the bug introduced? What was the behavior?
If a new feature, why was this architecture chosen? What were the alternatives?
If a refactor, how is this better than the previous implementation?

If there is a design document for this feature, please link it here.
-->

### Type of Change

<!--
Please check relevant options, delete irrelevant ones.
-->

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release







<!--
If just refactoring / back-end changes, this can be just an in-English description of the change at a technical level.
[...]

Issue #1291, opened 2026-03-11
Update translation value for `header.search.placeholder` key


ckeshava:
The translation for `header.search.placeholder` key will be updated in the PR: https://github.com/ripple/explorer/pull/1281. Please update this translation value in non-English languages. The new value includes `VaultID` text i.e. ` "Search by Token, Address, Ledger, Txn or VaultID"`

PR #1146, opened 2025-03-31
feat: Add basic ledger entry page

mvadari:
## High Level Overview of Change

This PR adds a new type of page for ledger entries/objects. For now it just has a default page that shows all the fields of an object, but over time we can add special ones.

It also:
* Adds links to these pages in the `Detailed` tab metadata breakdown
* Redirects to the `Account` page if the object is an account
* Handles deleted objects (if connected to a Clio node)

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Translation Updates
- [ ] Release

### TypeScript/Hooks Update

- [ ] Updated files to React Hooks
- [ ] Updated files to TypeScript

## Before / After

image

PR #1342, opened 2026-09-12
feat(search): resolve loan broker and loan ids to their vault

gamween:

Pasting a `LoanBroker` or `Loan` ledger index into the search returns `hash_not_found`, although the explorer already renders both objects on the vault page and both resolve over `ledger_entry`. This routes them to the vault that owns them.

The vault lookup in `determineHashType` becomes a single `getLedgerEntry` call switched on `LedgerEntryType`:

| entry | routes to |
|---|---|
| `Vault` | the vault page, unchanged |
| `LoanBroker` | its `VaultID` |
| `Loan` | its broker, then that broker's vault |
| anything else | falls through to `hash_not_found`, unchanged |

The request count is unchanged: one `ledger_entry` call served vaults before and serves all three now. A `Loan` costs one extra call to read its broker. The next ledger object type is a `case` rather than another round trip.

### Context of Change

**The behaviour today.** Type `055FDE074EC86E62F94F32D296898DA8A9D2586262AD1B912CA6066377D0BE9F` into the search box on devnet.xrpl.org and it navigates to `/search/055FDE07…`, the not-found page. Type the vault id `6B79B084F9EA00FB3B8B368D5710F50A871D488DE43095500B38C9FAD6D2C13C` and it navigates to `/vault/…`, which renders that same broker under Loans, with its first-loss capital, debt ceiling and loan table. The loan `A7920C443CA7AD71803F7F8E230FC7E1B53FC560EF10DD5EA59DE9D621F54165` behaves the same way.

**Relationship to #1320.** That PR deliberately made any non-`Vault` ledger entry fail into not-found, after a `PermissionedDomain` id was landing on a vault page. This change preserves that exactly: the `default` case of the switch throws, so anything other than the three types falls through as before. There is a test for it, using a `PermissionedDomain` node.

**Relationship to #1146.** `feat: Add basic ledger entry page` has been an open draft against `staging` since March 2025 and takes the general road: a `/entry/:id` route for any ledger object. It predates the Vault page from #1281. The position here is narrower on purpose: a broker or a loan is better served by the vault page that already renders its cover, its debt ceiling and its loan table than by a generic field dump, and this change is a two-line switch that #1146 can absorb or replace. Happy to close this if you would rather land #1146 first.

**Known limitation.** A vault may own more than one `LoanBroker`. The vault page selects the first broker tab from local state, so a searched broker is not yet the selected tab. Anchoring it needs a tab or query parameter on `VAULT_ROUTE`; glad to add it here if you prefer.

**Translations.** Only `en-US` is touched. The other six locales keep their current strings, and #1291 already tracks re-translating `header.search.placeholder` after #1281 changed it; this moves that key again and adds `hash_not_found` to the same list.

### Type of Change

- [ ] Documentation Updates
- [x] Translation Updates
- [ ] Release

### Codebase Modernization

Both touched source files were already hooks and TypeScript; no conversion was needed.

### Test Plan

`src/containers/Header/test/Search.test.js`: the file's 4 tests pass. The scenarios below are assertions added to its existing `search values` test, not new test cases:

- a loan broker resolves to its vault
- a loan resolves through its broker to the same vault, asserting `getLoanBroker` is called with the id named on the loan
- a `PermissionedDomain` entry still falls through to `/search/:id`, the #1320 behaviour
- a `LoanBroker` with no `VaultID` also falls through rather than building a broken route
- the existing cases now mock `getLedgerEntry` in place of `getVault`

`npx eslint --max-warnings 0`, `npx prettier --check` and `npx tsc --build` are clean. The branch was also run against live Devnet with the objects above.

Found while building a lending application on XLS-65 and XLS-66 at the XRPL Lending Protocol hackathon in Paris.


gamween:
Both strings now name loan brokers explicitly, `8c361b5`:

- `header.search.placeholder`: Search by Token, Address, Ledger, Txn, Vault, Loan Broker, Loan or LedgerHash
- `hash_not_found`: Could not find any Transactions, Vaults, Loan Brokers, Loans, Ledgers or NFTs that match the specified ID

Only `en-US` is touched; the other six locales keep the current string until translated.

brianstohl87-cpu (commented):
Better now?

